PR #12, opened 2022-12-13
Add support for fetch/pull

adityasaky:


adityasaky:
Current status: the workflow is getting there but there are a number of cases to check for still. Also, I need to rethink the helpers or write new ones because as of [533fe32](https://github.com/adityasaky/gittuf/pull/12/commits/533fe32bae428c63b203d5bceb4847e9018779e0), it's still rather inefficient.

adityasaky:
Merging this though a significant refactor may come about as a result of #13.

Issue #15, opened 2023-01-12
Distribute public keys for signed commits / tags


adityasaky:
Currently, we distribute the keys used to sign states (this applies whether it's states as a whole or RSL described in #13). When specifying policies, we can currently say that some state of a ref was signed off on by the holder of a key recorded within gittuf. We want to also be able to say that a state was reached because of the holder of key K1 AND the commits introduced in this state change were signed by the holder of some key K2 (K1 and K2 may well be the same key).

There are a number of benefits to doing this.
1. By supporting this, gittuf can bootstrap the keys used for Git signing. This is valuable for users whose threat model may not require finer access control policies or a reference state log (though they still get the latter for free).
2. We largely tackle the observer problem. By not being able to define policies for the commit signers and instead relying on the signer of the state log entry, we misattribute authorship of commits that may be merged by another maintainer. With the feature described here, we can allow User A to merge changes into a protected branch (which they are authorized to do) where the changes are made by User B who is the sole authorized developer to change some file in the repository. The policy validation can now confirm the protected file was modified by the authorized user instead of inferring the modification was performed by User A who merely merged it into a protected branch.

cc @znewman01

adityasaky:
This will be part of #52.

adityasaky:
Implementation plan:

Provide `gittuf verify-commit` and `gittuf verify-tag` functionality that shadows the equivalent commands in Git, except that the gittuf versions use public keys in the policy metadata.

In the case of `verify-commit`, all public keys are used to check the signature as commits aren't inherently associated with a ref. For `verify-tag` though, we could look at the policy for authorized tag creators and warn if a tag is signed by someone not authorized.

Also, these commands must be disconnected from the latest policy as again, commits may be created and signed by someone who has seen been rotated out. We have to therefore check all policy states to find public keys.

Issue #13, opened 2022-12-23
Decouple branch states from rule states


adityasaky:
Currently, new valid states for branches (latest trusted commit, for example) is recorded in targets metadata. This issue discusses decoupling the records of branch states from the states that describe the access control policies for a repository. This has a number of advantages:
1. It greatly reduces the footprint of TUF metadata
2. It is simpler to maintain state linearity of the TUF metadata if the state is only tracking changes to access control policies--these are far rarer than commits in practice
3. A separate record of branch states means that it can record states of branches not explicitly protected by the rules (ex: short lived feature branches) but are still vulnerable to branch state attacks
4. (side effect) The CLI would improve as the user would no longer have to specify the role they're updating

adityasaky:
A decoupled state log would very closely resemble what's described in https://www.usenix.org/system/files/conference/usenixsecurity16/sec16_paper_torres-arias.pdf. For every reference, we'd have an entry recording current state and the previous state. Each entry would be signed by the user updating the state log (and this user can be separate from that creating a commit at some reference, merely an observer).

adityasaky:
Decoupling as described in this thread requires a redesign of metadata management. Specifically, the internal only gitstore package will now have:
* `policy-state`, very similar to the current state log but from now only handling policies described as TUF metadata
* `reference-state`, brand new namespace within `refs/gittuf/`, each tracked ref in a copy of the repo has its own entry
* `last-trusted`, very similar to the current version except instead of state IDs pointing to `policy-state`, they point to `reference-state` entries

## reference-state

Stored within `refs/gittuf/reference-state/`, each entry corresponds to a tracked ref. For example, the `main` branch is tracked at `refs/gittuf/reference-state/main`. This file points to a blob artifact corresponding to the reference state log entry. The entry has the following fields:
* `last-entry`: ID of the preceding blob
* `tip`: ID of the tip of the reference
* `signature`: a signature object over the `last-entry` and `tip` fields

Question: should an entry contain a record of the policy state it was created against? This might help detect and reconcile drifts in states.

adityasaky:
Side note: `gitstore` is no longer a lightweight interface to handle interactions to store TUF metadata in the gittuf namespace. We should rename it and clean up the implementation. Perhaps `internal/gittuf-metadata` and `package metadata`?

adityasaky:
[...]

PR #16, opened 2023-01-19
Refactor specification to incorporate gittuf design

adityasaky:
I'm updating the spec document to incorporate the design. It includes the changes discussed in #13 and #15 and should guide the development of the implementation as we move away from the current prototype.


docs/specification.md
```diff
@@ -1,49 +1,190 @@
...
+
+Last Modified: January 19, 2023
+
+Version: 0.1.0
+
+## Introduction
+
+This document describes gittuf, a security layer for Git repositories. gittuf
+applies several key properties part of the
+[The Update Framework (TUF)](https://theupdateframework.io/) such as delegations
```
JustinCappos, line 11:
```suggestion
[The Update Framework (TUF)](https://theupdateframework.io/) such as delegations,
secure key distribution, key revocation, trust rotation, read / write access control, 
```

docs/specification.md
```diff
@@ -1,49 +1,190 @@
...
+and namespaces to Git repositories. This enables owners of the repositories to
+distribute (and revoke) contributor signing keys and define policies about which
+contributors can make changes to some namespaces within the repository. gittuf
+also implements the Reference State Log which was originally described in an
```
JustinCappos, line 15:
```suggestion
also protects against branch state attacks by extending the Reference State Log design
which was originally described in an
```

docs/specification.md
```diff
@@ -1,49 +1,190 @@
...
+the repository.
+
+Policies for gittuf access are defined using a subset of TUF roles. The owners
+of the repository hold the keys used to sign the Root role that delegates trust
+to the other roles. The top level Targets role and any Targets roles it
+delegates to contain restrictions on protected namespaces. The specifics of the
+delegation structure vary from repository to repository as each will have its
+own constraints.
+
+A typical TUF delegation connects two TUF Targets role. In the delegations
```
JustinCappos, line 97:
```suggestion
A typical TUF delegation connects two TUF Targets roles. In the delegations
```

docs/specification.md
```diff
@@ -1,49 +1,190 @@
...
+graph where each node is a Targets role, delegations are the edges connecting
+them. The delegations graph is traversed when verifying some target until the
+leaf node is encountered, after which the target's entry is compared against the
```
JustinCappos, line 100: This is a little confusing.  There really isn't a meaningful notion of leaf nodes as you may have cycles in delegations and may also have intermediate nodes that directly list targets in addition to delegating.

adityasaky: I'll take another pass. gittuf has a very opinionated approach to TUF delegations. Users are not expected to hand-craft metadata and we always expect the entries to be the commits themselves or the RSL entry so we shouldn't have the intermediate metadata recording hashes issue.

adityasaky:
I've moved the SHA-256 bits to a standalone document. I see this going forward as:
1. primary specification that handles the TUF access control policies and corresponding workflows
2. other accompanying documents that describe extensions to this specification for hash algorithm agility, attestations, anything else we want to add

I believe the specification document is ready for reviews / discussions. We should tackle the SHA-256 bits separately.

docs/specification.md
JustinCappos: What do you mean that entries are the commits themselves?  I thought they were either a path, branch, or tag?

adityasaky: I've updated this part of the document to try and make it clearer.

JustinCappos: I would tweak the new text to more state that TUF's namespacing is general and can be used for various git namespaces like tags, files, branches, etc.

adityasaky: Mentioned namespaces generally and said gittuf applies it to git but modifies the overall workflow to incorporate RSLs etc.

PR #23, opened 2023-02-24
Update specification.md

JustinCappos:
I'm trying to clarify a few things and these likely should be separate PRs.  I think the second change is likely controversial and we should discuss...

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -97,8 +97,9 @@ compromised, new metadata is issued to revoke its trust.
 
 Second, TUF allows for defining _namespaces_ for the repository. TUF's notion of
 namespaces aligns with Git's, and TUF namespaces can be used to reason about
-both Git refs and files tracked within the repository. Namespaces are combined
-with TUF's _delegations_ to define sets of actors who are authorized to make
+both Git refs (tags and branches) and files tracked within the repository. 
+Namespaces are combined with TUF's _delegations_ to define the set of actors 
+who are authorized to view a namespace as well as those who are authorized to make
 changes to some namespace. As such, the owner of the repository can use gittuf
 to define actors representing other contributors to the repository, and delegate
 to them only the necessary authority to make changes to different namespaces of
@@ -121,13 +122,12 @@ target. gittuf applies this namespaced delegations graph traversal to Git and
 also incorporate RSLs and Git's implicit change tracking mechanisms.
 
 In gittuf, the delegations graph is similarly traversed, except that it
-explicitly does not expect any Targets metadata to contain an entry. Instead,
+explicitly does not expect any Targets metadata to contain a target entry. Instead,
 the delegation mechanism is used to identify the set of keys authorized to sign
 the target such as an RSL entry or commit being verified. Therefore, the
-delegation graph is searched until a delegation is encountered such that no
-metadata exists in the repository for the delegatee role. At this point, the
-search is terminated and the keys listed in that delegation entry are used as
-the set of authorized keys.
+delegation graph is used to decide which keys git actions should trust, but no
+targets entries are used.  Any key which delegated trust up to this part of the 
+namespace (including the last delegation), is trusted to sign the git actions.
 
 This mechanism is employed when verifying both RSL entries for Git ref updates
 _and_ when verifying the commits introduced between two ref updates. The latter
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -97,9 +97,8 @@ compromised, new metadata is issued to revoke its trust.
 
 Second, TUF allows for defining _namespaces_ for the repository. TUF's notion of
 namespaces aligns with Git's, and TUF namespaces can be used to reason about
-both Git refs (tags and branches) and files tracked within the repository. 
-Namespaces are combined with TUF's _delegations_ to define the set of actors 
-who are authorized to view a namespace as well as those who are authorized to make
+both Git refs and files tracked within the repository. Namespaces are combined
+with TUF's _delegations_ to define sets of actors who are authorized to make
 changes to some namespace. As such, the owner of the repository can use gittuf
 to define actors representing other contributors to the repository, and delegate
 to them only the necessary authority to make changes to different namespaces of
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -97,8 +97,9 @@ compromised, new metadata is issued to revoke its trust.
 
 Second, TUF allows for defining _namespaces_ for the repository. TUF's notion of
 namespaces aligns with Git's, and TUF namespaces can be used to reason about
-both Git refs and files tracked within the repository. Namespaces are combined
-with TUF's _delegations_ to define sets of actors who are authorized to make
+both Git refs (tags and branches) and files tracked within the repository.
```
adityasaky, line 100: I've defined git refs at the start of the document. The reason I don't want to specify this here is because people may have different interpretations of "branches". Some may consider only `refs/heads/*` as branches while others may consider anything under `refs/*` that is NOT `refs/tags/*` as branches. We use the RSL to track `refs/gittuf/policy`, for example. The definition describes that arbitrary refs may exist.

docs/specification.md
```diff
@@ -97,8 +97,9 @@ compromised, new metadata is issued to revoke its trust.
 
 Second, TUF allows for defining _namespaces_ for the repository. TUF's notion of
 namespaces aligns with Git's, and TUF namespaces can be used to reason about
-both Git refs and files tracked within the repository. Namespaces are combined
-with TUF's _delegations_ to define sets of actors who are authorized to make
+both Git refs (tags and branches) and files tracked within the repository. 
+Namespaces are combined with TUF's _delegations_ to define the set of actors 
+who are authorized to view a namespace as well as those who are authorized to make
```
adityasaky, line 102: I think we have some work to do on defining how read policies work first? For writes, we know exactly what we're verifying--commits/tags/other refs/RSL entries/files etc. For reads, it's less clear. Do we implement a `gittuf reveal <secret file>`? Does TUF metadata include the encrypted symmetric key that can be used to decrypt the secret in `custom`? Should `custom` always indicate the type of protection? There's also the matter of `custom` belongs in targets entries and not in delegation entries so this is a little complex. WDYT?

docs/specification.md
```diff
@@ -121,13 +122,12 @@ target. gittuf applies this namespaced delegations graph traversal to Git and
...
-explicitly does not expect any Targets metadata to contain an entry. Instead,
+explicitly does not expect any Targets metadata to contain a target entry. Instead,
 the delegation mechanism is used to identify the set of keys authorized to sign
 the target such as an RSL entry or commit being verified. Therefore, the
-delegation graph is searched until a delegation is encountered such that no
-metadata exists in the repository for the delegatee role. At this point, the
-search is terminated and the keys listed in that delegation entry are used as
-the set of authorized keys.
+delegation graph is used to decide which keys git actions should trust, but no
+targets entries are used.  Any key which delegated trust up to this part of the
```
adityasaky, line 129: I think I'm okay with this, it likely simplifies the implementation of delegation tree traversal.

docs/specification.md
JustinCappos: > I think we have some work to do on defining how read policies work first? For writes, we know exactly what we're verifying--commits/tags/other refs/RSL entries/files etc. For reads, it's less clear. Do we implement a `gittuf reveal <secret file>`? 

I don't think we need a reveal command.  I think that when you commit, etc. files you can decrypt things locally.  


> Does TUF metadata include the encrypted symmetric key that can be used to decrypt the secret in `custom`? 

Yes, I think so.  Although storing it other places may also be acceptable.

> Should `custom` always indicate the type of protection? There's also the matter of `custom` belongs in targets entries and not in delegation entries so this is a little complex. WDYT?

Custom should likely indicate the type of protection.  I think it needs to indicate the type of namespace, so this seems sensible.

The fact that delegation entries cannot include a custom field seems like an oversight in the TUF specification to me.  We might see what others think...

docs/specification.md
JustinCappos: okay, great.  It helps to get rid of the weird case where Alice wants to keep control while delegating to Bob and whomever Bob wants, but Bob hasn't listed Alice.  Alice could remove the delegation, but this isn't ideal.

docs/specification.md
adityasaky: > I don't think we need a reveal command. I think that when you commit, etc. files you can decrypt things locally.

I'm a little wary here. It's too easy to `git commit` a decrypted file.

> The fact that delegation entries cannot include a custom field seems like an oversight in the TUF specification to me. We might see what others think...

I'm experimenting with this here: https://github.com/adityasaky/gittuf/pull/27/files#diff-6f6158fdacef6b49dbe116bd6b52a72ab2a8cdd9cd0fe6cd7f531e86b163d185R87.

JustinCappos: > > I don't think we need a reveal command. I think that when you commit, etc. files you can decrypt things locally.
> 
> I'm a little wary here. It's too easy to `git commit` a decrypted file.

It should be transparent to you.  In other words, if Alice is committing foo and Bob can't read foo, Alice's commit encrypts the data first before adding it.  

Obviously, this only works if Alice is using git-tuf and not git.

docs/specification.md
adityasaky: @JustinCappos can I reverse this specific change?

docs/specification.md
adityasaky: Correct, but if Alice can read it and her gittuf client decrypts it, by default git assumes there's either a modification of the checked in file or that a new file has been created (if decrypted to a separate location). At that point, if Alice does a `git add .`, she may accidentally check in a secret. We may be able to do something like decrypting to a file in an ignored `.gittuf-secret` file but the UX is still problematic for humans. It may be simpler when configuring a bot to use a secret.

docs/specification.md
JustinCappos: Yes, this is fine.

docs/specification.md
JustinCappos: Okay, you bring up some interesting points.  

So the options include:

1. Masking files where read access isn't universal from git-tuf by .gitignore.  This would also hide it from git...
2. Copying read-only files to a new location, like you have in your example
3. Living with the problem and calling misuse a 'user error'.  This may not be as bad as it seems, because checking in a non-encrypted file to a git repo where there are access control rules, could trigger a repo running git-tuf scripts to reject the change.
4. ???

I'd also like to mention that this all becomes more complex if Alice and Bob are sharing a repo (no read-only access) and later they want to add Charlie, but restrict what Charlie can read.  I'm not sure how to think about this case, but the way I've been thinking about doing this with encryption doesn't seem to work to stop Charlie from reading the versions of files that existed before Charlie joined.

adityasaky: Let's move this to an issue to define how gittuf handles secrets. I'd like to merge the PR with the change below so I'm reversing this as well.

adityasaky (approved):

docs/specification.md
adityasaky: https://github.com/adityasaky/gittuf/issues/29

Issue #19, opened 2023-02-17
Use in-memory git repositories for tests


adityasaky:
Where possible, we should use in-memory git repositories for tests instead of creating fs based repositories in `/tmp`. We currently do not do this for most of our tests because our git config parser shells out to the git binary. See: https://github.com/adityasaky/gittuf/pull/18#discussion_r1108961567 for full discussion.

PR #24, opened 2023-02-25
Add support for RSL `Entry` and `Annotation` types

adityasaky:


rsl/rsl.go
```diff
@@ -39,37 +43,132 @@ func InitializeNamespace() error {
 	return nil
 }
 
-// AddEntry adds an RSL entry to the log for the ref and hash passed in.
-func AddEntry(refName string, commitID plumbing.Hash, sign bool) error {
+type RSLEntry struct {
+	RefName  string
+	CommitID plumbing.Hash
+	Revoke   []plumbing.Hash
```
wlynch, line 49: My current understanding of the RSL (please correct me if I'm wrong) is we're attaching a list of revoked revisions to a RSL object on the tuf ref that references a {ref, commit} state in a "main branch" (we may want to add comments for each of these fields).

To figure out a complete RSL state, do we need to walk the entire tuf ref, or is looking at the latest HEAD good enough? (i think based on the command this only appends new revocations?)

I'm worried about operations needing to walk the entire RSL tree for each operation any time we need to make a query - maybe we can precompute this and reference it as it's own blob? 🤔

adityasaky: I think there are a few different questions there, let me try and answer them.

1. The RSL is stored at `refs/gittuf/reference-state-log`. It points to the tip commit, the latest RSL entry available.
2. A standard RSL entry (not a revocation) contains `RefName` and `CommitID`. `RefName` can be any ref in the repository, using the full `refs/<>/` format. The `CommitID` is the object that ref is supposed to be at at the time of signing the entry.
3. A revocation entry includes the identifiers of one or more prior RSL entries. The revoking entry and the entries revoked must all have identical `RefName`s.
4. The tip of the RSL namespace points to the latest entry. So looking at the tip isn't enough to figure out the expected states of all tracked refs, it is necessary to walk back. I'm certainly open to optimize this, I had, for example, in the prototype a local-only record of the last trusted entry for each red tracked on that copy of the repository. I think we have to threat model that though. I suspect that `GetEntry` results can at least be locally cached when performing an operation that has a lot of those calls.

wlynch: 👍 cool, that tracks with what I had in my head. Thanks for confirming!

I'm still a bit concerned about needing to walk the entire graph (https://github.com/torvalds/linux would be a good repo to simulate on 🤣 ), but I'm okay with deferring on solving this for now.

My thinking was have ref to a blob that contains the entire computed state:

image

(not sure if this gives us the same threat model properties though)

rsl/rsl.go
```diff
@@ -39,37 +43,132 @@ func InitializeNamespace() error {
...
+func NewNormalEntry(refName string, commitID plumbing.Hash) *RSLEntry {
+	return &RSLEntry{RefName: refName, CommitID: commitID, Revoke: []plumbing.Hash{}}
+}
+
+// NewRevocationEntry returns an RSLEntry object that revokes a prior, specified
+// RSL entry.
+// TODO: fix docstring
+func NewRevocationEntry(refName string, commitID plumbing.Hash, revokedEntryIDs []plumbing.Hash) *RSLEntry {
+	return &RSLEntry{RefName: refName, CommitID: commitID, Revoke: revokedEntryIDs}
+}
```
wlynch, line 63: up to you: If you want you could use a variadic param here:

```suggestion
func NewRevocationEntry(refName string, commitID plumbing.Hash, revokedEntryIDs ...plumbing.Hash) *RSLEntry {
	return &RSLEntry{RefName: refName, CommitID: commitID, Revoke: revokedEntryIDs}
}
```

rsl/rsl.go
```diff
@@ -39,37 +43,132 @@ func InitializeNamespace() error {
...
+		if iteratorHash == entryID {
+			return parseRSLEntryMessage(commitObj.Message), nil
+		}
+
+		if len(commitObj.ParentHashes) == 0 {
+			return &RSLEntry{}, ErrRSLEntryNotFound
+		}
+
+		if len(commitObj.ParentHashes) > 1 {
+			return &RSLEntry{}, ErrRSLBranchDetected
```
wlynch, line 150: It'd be interesting to see how we can handle this 🤔 

Merge commits are not uncommon given it's the default for GitHub PRs.

(fine with ignoring for now though)

rsl/rsl.go
```diff
@@ -39,37 +43,132 @@ func InitializeNamespace() error {
...
+func parseRSLEntryMessage(message string) *RSLEntry {
+	entry := &RSLEntry{
+		Revoke: []plumbing.Hash{},
+	}
+
+	for _, l := range strings.Split(message, "\n") {
+		ls := strings.Split(l, ":")
+		switch strings.TrimSpace(ls[0]) {
+		case "ref":
+			entry.RefName = strings.TrimSpace(ls[1])
```
wlynch, line 166: You'll want to check the len - this will panic you don't get an actual split.

I also don't know if `:` is valid for ref names - you may need `ls[1:]` and `TrimPrefix`

wlynch (approved):

rsl/rsl.go
adityasaky: My thinking is we can control this as all RSL entries are created by gittuf. We only care about merge commits in `refs/gittuf/reference-state-log` to avoid any forking in the log.

rsl/rsl.go
adityasaky: Good call, this is neater.

rsl/rsl.go
adityasaky: > I'm still a bit concerned about needing to walk the entire graph (https://github.com/torvalds/linux would be a good repo to simulate on :rofl: ), but I'm okay with deferring on solving this for now.

I was quite worried about that as well but @JustinCappos rightly pointed out that those repositories are outliers. Also, I suspect clients will rarely have to walk the full history or anything close to it. Most walks will be to update local copies from what's available upstream.

> My thinking was have ref to a blob that contains the entire computed state:

Originally, I was planning to include a blob for each entry but I dismissed it as all the information for any one entry can easily be stored in the commit itself. Pre computed states for all refs would be nice but I think validating any entry present in the latest precomputed blob would still require walking back the tree to when the change was made.

rsl/rsl.go
adityasaky: `fatal: 'new:branch' is not a valid branch name`

Luckily not as `:` is also used for `local:remote` mapping.

> You'll want to check the len - this will panic you don't get an actual split.

Good call, I'll add that.

adityasaky: [3fbfb93](https://github.com/adityasaky/gittuf/pull/24/commits/3fbfb93fd65b7ccd8983df4e5c8ceffe41d7a021)

rsl/rsl.go
adityasaky: [3fbfb93](https://github.com/adityasaky/gittuf/pull/24/commits/3fbfb93fd65b7ccd8983df4e5c8ceffe41d7a021)

adityasaky:
@wlynch This one's been reworked significantly. @JustinCappos and I were discussing how to keep the different types of entries simpler and we settled on having one `Entry` type and one `Annotation` type. `Entry` is very simple, just includes the ref and commit information. `Annotation` comes later and can add information about any prior `Entry` or `Annotation`. It has a `Skip` boolean which is equivalent to what the earlier revocation types were, but with that set to false, an annotation can simply add more information to a prior entry. So, it can be used to highlight potential violations not already caught in an entry and so on.

I haven't yet updated the specification document but I think this is ready for a code review. I'm not very happy with  `EntryType` and the type casting it results in and I briefly played around with using generics, but it didn't quite make sense for parsing an entry without knowing ahead of time if it was an `Entry` type or `Annotation` type.

adityasaky:
@JustinCappos I've updated the specification with `Annotation` support. I think there's more to hash out with the workflow but I think that'll happen as we implement the workflow and run it against test scenarios.

adityasaky:
I'm going to submit one more change in the morning.

adityasaky:
I'll clean up commit history post review on this one. This is good to go now for another code review!

internal/rsl/rsl.go
```diff
@@ -0,0 +1,273 @@
...
+		}
+
+		iteratorHash = commitObj.ParentHashes[0]
+	}
+}
+
+func parseRSLEntryText(text string) (EntryType, error) {
+	lines := strings.Split(text, "\n")
+	header := strings.TrimSpace(lines[0])
+	lines = lines[2:]
```
wlynch, line 202: You should do a len check here to make sure you're getting 3+ lines.

internal/rsl/rsl_test.go
wlynch, line 1: Same general comments here about whether you want to test via filesystems directly or just use an in-memory repo. I'll defer to you.

internal/rsl/rsl.go
```diff
@@ -0,0 +1,273 @@
...
+			continue
+		}
+
+		ls := strings.Split(l, ":")
+		if len(ls) < 2 {
+			return nil, ErrInvalidRSLEntry
+		}
+
+		switch strings.TrimSpace(ls[0]) {
+		case "entryID":
```
wlynch, line 239: Consider pulling these into a const, since these need to match up.

internal/rsl/rsl.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	commitObj, err := repo.CommitObject(ref.Hash())
+	if err != nil {
+		return nil, err
+	}
+
+	return parseRSLEntryText(commitObj.Message)
```
wlynch, line 160: You may want to consider refactoring this so this just calls GetEntry. This way any changes in entry parsing can be consolidated there. This is just a special case where we want to get the first value.

wlynch (changes requested):
Some minor comments, but mostly lgtm!

internal/rsl/rsl_test.go
adityasaky: I took a pass at the tests and the ones that use a repo currently assume it's in the filesystem for the `refs/gittuf` namespace. I think I'd rather revisit them via #19.

internal/rsl/rsl.go
adityasaky: Agreed but we'd be doing 80% of the work to get the entry ID to pass to GetEntry.

adityasaky:
Addressed most of the comments bar `GetEntry`/`GetLatestEntry` and cleaned up the commit history. Thanks for the review, @wlynch!

internal/rsl/rsl.go
adityasaky: And GetEntry would then redo getting the reference...

internal/rsl/rsl.go
```diff
@@ -0,0 +1,296 @@
...
+	if strings.HasPrefix(text, AnnotationHeader) {
+		return parseAnnotationText(text)
+	}
+	return parseEntryText(text)
+}
+
+func parseEntryText(text string) (*Entry, error) {
+	lines := strings.Split(text, "\n")
+	lines = lines[2:]
+	if len(lines) < 2 {
```
wlynch, line 232: This needs to happen before the access above, otherwise the lines[2] might be OOB

wlynch (changes requested):
Just need to clean up some bounds checking, otherwise lgtm!

internal/rsl/rsl.go
adityasaky: Good catch! I moved it one line up and changed the comparison to < 4 to account for the header. I also added two tests apiece for entries and annotations, one with missing header, the other with missing information but including the header to assert `ErrInvalidRSLEntry`

internal/rsl/rsl.go
```diff
@@ -0,0 +1,296 @@
...
+		RSLEntryIDs: []plumbing.Hash{},
+	}
+
+	messageBlock, _ := pem.Decode([]byte(text)) // rest doesn't seem to work when the PEM block is at the end of text, see: https://go.dev/play/p/oZysAfemA-v
+	if messageBlock != nil {
+		annotation.Message = string(messageBlock.Bytes)
+	}
+
+	lines := strings.Split(text, "\n")
+	lines = lines[2:]
```
wlynch, line 267: here as well!

adityasaky: Same change made, < 4. I guess the test I added before wasn't actually triggered because of the missing header hehe.

wlynch (approved):

adityasaky:
Thanks, @wlynch!

PR #27, opened 2023-03-06
Add support for policies

adityasaky:
This PR is currently a mess. The code's a mess, the commit history is a mess. I'm using this branch as a way to back up my work as I work things out. I'll clean everything up once it all comes together. :smile:

PR #32, opened 2023-03-16
Add plumbing for policy

adityasaky:
This PR introduces parts of #27 to make it easier to review. I'll keep working on that PR to add initial policy support but I'll probably keep breaking parts of it away for reviewability.

Additionally, some of the things here with the signers and so on can probably be upstreamed to go-securesystemslib. If they pass muster here, I can submit them separately, but I think it may be faster to define them here for now. I'm curious, also, about what we want to support out of the box for signing metadata. We can take inspiration from python-securesystemslib perhaps?

The tuf package includes models of TUF metadata with some minor modifications. Largely, they are based on what's defined in go-tuf.

The signerverifier package provides some plumbing for use with DSSE to sign the TUF metadata. Within it, the ed25519 package creates a DSSE compatibler signer and verifier. The dsse package provides opinionated APIs to create and verify signed metadata.


internal/signerverifier/ed25519/ed25519.go
```diff
@@ -0,0 +1,120 @@
...
+	var k interface{}
+	if err := json.Unmarshal(keyContents, &k); err != nil {
+		return &Ed25519SignerVerifier{}, err
+	}
+
+	t := k.(map[string]interface{})
+
+	var keyID string
+	if _, ok := t["keyid"]; ok {
+		keyID = t["keyid"].(string)
```
wlynch, line 35: Instead of doing this type casting, I'd just define a struct and unmarshal there, then do the transformation into the SV.

internal/signerverifier/ed25519/ed25519.go
```diff
@@ -0,0 +1,120 @@
...
+		return &Ed25519SignerVerifier{}, err
+	}
+
+	if _, ok := keyval["private"]; ok {
+		privateS := keyval["private"].(string)
+		private, err := hex.DecodeString(privateS)
+		if err != nil {
+			return &Ed25519SignerVerifier{}, err
+		}
+		if len(private) == ed25519.PrivateKeySize/2 {
```
wlynch, line 60: As someone who's less familiar with ed25519, why do we need to do this?

internal/signerverifier/ed25519/ed25519_test.go
```diff
@@ -0,0 +1,148 @@
...
+	assert.Equal(t, expectedPrivateKey, sv.private)
+	assert.Equal(t, expectedPublicKey, sv.public)
+
+	publicKeyPath := path.Join("test-data", "test-key.pub")
+	publicKeyBytes, err := os.ReadFile(publicKeyPath)
+	if err != nil {
+		t.Error(err)
+	}
+
+	sv, err = NewSignerVerifierFromSecureSystemsLibFormat(publicKeyBytes)
```
wlynch, line 53: I'm assuming these are 2 logically different tests because we're reading in new data.

It might be helpful to break this up using t.Run() to make this distinction clearer. 2 ways you could do this:

1. just wrap each case with t.Run individually. e.g.
```go
t.Run("foo", func(t *testing.T) {
...
})
t.Run("bar", func(t *testing.T) {
...
})
```
2. Set up a table driven test.

No strong preferences here. Up to you!

wlynch (changes requested):
Overall looks good!

Just some small tweaks.

internal/signerverifier/ed25519/ed25519.go
adityasaky: I've taken a pass at this. I must note that it's not clean because I didn't want to include private key support in the internal/tuf package.

internal/signerverifier/ed25519/ed25519.go
adityasaky: Yep, I really should add a comment there!

adityasaky: Done!

adityasaky:
@wlynch Thanks for the review, I think this should be good to go minus any thoughts you have on handling private keys. :smile:

PR #33, opened 2023-03-20
More plumbing for policy

adityasaky:
Like #32, this moves some more parts of #27 into a standalone PR.

After this is good to go, we'll be focusing on the pure creation of TUF metadata bits. Side note: I have something based off sigstore/root-signing for generating the root metadata with multiple parties but we should discuss the UX.

### DIFF
--- a/internal/common/utils.go
+++ b/internal/common/utils.go
@@ -4,10 +4,44 @@ import (
 	"bytes"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/go-git/go-git/v5"
 )
+
+const GittufNamespace = "refs/gittuf"
+
+// InitializeGittufNamespace creates the refs/gittuf directory if it does not
+// already exist. It does not initialize the refs for policy and RSL namespaces.
+func InitializeGittufNamespace(targetDir string) error {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	if err := os.Chdir(targetDir); err != nil {
+		return err
+	}
+	defer os.Chdir(cwd) //nolint:errcheck
+
+	repoRootDir, err := GetRepositoryRootDirectory()
+	if err != nil {
+		return err
+	}
+
+	refPath := filepath.Join(repoRootDir, GetGitDir(), GittufNamespace)
+	if _, err := os.Stat(refPath); err != nil {
+		if os.IsNotExist(err) {
+			if err := os.Mkdir(refPath, 0755); err != nil {
+				return err
+			}
+		} else {
+			return err
+		}
+	}
+
+	return nil
+}
 
 func GetRepositoryRootDirectory() (string, error) {
 	cmd := exec.Command("git", "rev-parse", "--show-toplevel")
@@ -33,6 +67,16 @@ func CreateTestRepository() (string, error) {
 	if err != nil {
 		return testDir, err
 	}
+
 	cmd := exec.Command("git", "init", testDir)
-	return testDir, cmd.Run()
+	if err := cmd.Run(); err != nil {
+		return testDir, err
+	}
+
+	return testDir, InitializeGittufNamespace(testDir)
+}
+
+func GetGitDir() string {
+	// TODO: handle detached gitdir
+	return ".git"
 }
--- a/internal/gitinterface/blob.go
+++ b/internal/gitinterface/blob.go
@@ -1,0 +1,49 @@
+package gitinterface
+
+import (
+	"errors"
+	"io"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+)
+
+var ErrWrittenBlobLengthMismatch = errors.New("length of blob written does not match length of contents")
+
+// ReadBlob returns the contents of a the blob referenced by blobID.
+func ReadBlob(repo *git.Repository, blobID plumbing.Hash) ([]byte, error) {
+	blob, err := repo.BlobObject(blobID)
+	if err != nil {
+		return nil, err
+	}
+
+	reader, err := blob.Reader()
+	if err != nil {
+		return nil, err
+	}
+
+	return io.ReadAll(reader)
+}
+
+// WriteBlob creates a blob object with the specified contents and returns the
+// ID of the resultant blob.
+func WriteBlob(repo *git.Repository, contents []byte) (plumbing.Hash, error) {
+	obj := repo.Storer.NewEncodedObject()
+	obj.SetType(plumbing.BlobObject)
+
+	writer, err := obj.Writer()
+	if err != nil {
+		return plumbing.ZeroHash, err
+	}
+
+	length, err := writer.Write(contents)
+	if err != nil {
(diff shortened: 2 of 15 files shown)


internal/gitinterface/blob_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"path"
+	"testing"
+
+	"github.com/adityasaky/gittuf/internal/common"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestReadBlobAndWriteBlob(t *testing.T) {
+	testFileReadSrc, err := os.Open(path.Join("test-data", "test-file-read"))
```
adityasaky, line 16: @wlynch any tips on how to handle the setup here and in tree_test.go? Also, I think I should update the chdir across the board to clean up after the test...

adityasaky: Never mind, I think I worked it out. :smile:

adityasaky:
Ahh I just noticed I forgot to update return statements to use nil based on the comment on the previous PR.

adityasaky:
Done!

internal/common/utils.go
```diff
@@ -33,6 +67,16 @@ func CreateTestRepository() (string, error) {
...
 	cmd := exec.Command("git", "init", testDir)
-	return testDir, cmd.Run()
+	if err := cmd.Run(); err != nil {
+		return testDir, err
+	}
+
+	return testDir, InitializeGittufNamespace(testDir)
+}
+
+func GetGitDir() string {
```
wlynch, line 79: Any reason why this should be a func over a const?

(my guess based on the TODO comment is you have future plans for this)

internal/common/utils.go
```diff
@@ -4,11 +4,45 @@ import (
...
+// already exist. It does not initialize the refs for policy and RSL namespaces.
+func InitializeGittufNamespace(targetDir string) error {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	if err := os.Chdir(targetDir); err != nil {
+		return err
+	}
+	defer os.Chdir(cwd) //nolint:errcheck
```
wlynch, line 25: This is fine as is, but if you want in the future [`--work-dir` flag](https://git-scm.com/docs/git#Documentation/git.txt---work-treeltpathgt) will let you run commands from another folder.

internal/gitinterface/blob.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+)
+
+func ReadBlob(repo *git.Repository, blobID plumbing.Hash) (int, []byte, error) {
+	blob, err := repo.BlobObject(blobID)
+	if err != nil {
+		return -1, nil, err
+	}
+	contents := make([]byte, blob.Size)
```
wlynch, line 13: nit:

```suggestion
	contents := make([]byte, 0, blob.Size)
```

What you have is probably fine assuming that the blob.Size always == the length read, but this is probably safer.

internal/gitinterface/blob.go
```diff
@@ -0,0 +1,38 @@
...
+	reader, err := blob.Reader()
+	if err != nil {
+		return -1, nil, err
+	}
+	length, err := reader.Read(contents)
+	if err != nil {
+		return -1, nil, err
+	}
+	return length, contents, nil
```
wlynch, line 22: I'm not sure returning the length is very useful.

This should usually be == blob.Size / len(contents).
If by chance it's not, we're not returning the reader so callers won't be able to easily pick up where the last read left off.

If you want to guarantee everything is read, use `io.ReadAll`.

internal/gitinterface/blob_test.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	if err := os.Chdir(testDir); err != nil {
+		t.Fatal(err)
+	}
+
+	repo, err := common.GetRepositoryHandler()
+	if err != nil {
+		t.Fatal(err)
+	}
```
wlynch, line 77:
```suggestion
	repo, err := git.Init(memory.NewStorage(), memfs.New())
	if err != nil {
		t.Fatal(err)
	}
```

See https://github.com/sigstore/gitsign/blob/2509c5c5f931b5de2cc9ca2449d7e309a65a2ec9/internal/config/config_test.go#L32-L35

internal/gitinterface/blob_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	cmd := exec.Command("git", "hash-object", "-w", "test-file-read")
+	if err := cmd.Run(); err != nil {
+		t.Fatal(err)
+	}
+
+	repo, err := common.GetRepositoryHandler()
+	if err != nil {
+		t.Fatal(err)
+	}
```
wlynch, line 37: Instead of doing all of this, you could do the in-memory setup + WriteBlob, then verify you get the same content on read.

internal/gitinterface/tree_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"testing"
+
+	"github.com/adityasaky/gittuf/internal/common"
+	"github.com/go-git/go-git/v5/plumbing/filemode"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWriteTree(t *testing.T) {
+	testDir, err := common.CreateTestRepository()
```
wlynch, line 15: You can do the same in-memory repo trick here as well.

wlynch (changes requested):

internal/common/utils.go
adityasaky: Git repos can be initialized with a different `$GIT_DIR` or `--separate-git-dir` (https://git-scm.com/docs/git-init#Documentation/git-init.txt---separate-git-dirltgit-dirgt). In that case, rather than a directory, we have a `.git` file:

```
~/tmp/gitobj $ cat .git
gitdir: /home/saky/tmp/gitdir
```

Eventually, this func should support parsing `.git` if it's a file to identify the actual git dir. Note, IIRC gitsign doesn't support this yet and I might submit a patch when I get the time to implement it here as well.

internal/gitinterface/blob.go
adityasaky: Okay, switched to this and dismissed `length` altogether. This should also address https://github.com/adityasaky/gittuf/pull/33#discussion_r1143501940 but I'm still curious if `len` is used by the reader rather than capacity.

wlynch: I was wrong with the last comment - it uses len not cap. Sorry about that!

> Reader is the interface that wraps the basic Read method.
> 
> Read reads up to len(p) bytes into p.

adityasaky: No worries, it's good to file this away for future :smile:

adityasaky:
Sneaking in a fix for `path.Join` invocations which should actually be `filepath.Join`...

wlynch (approved):

adityasaky:
Thanks @wlynch!

PR #39, opened 2023-04-25
Add repo handler

adityasaky:
The first commit cleans up a lot of the repo handling (and moves several tests to in-memory storage). The second commit adds a lightweight `Repository` handler which (as @wlynch suggested) can provide interfaces that users would use via the CLI.

### DIFF
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -1,10 +1,7 @@
 package policy
 
 import (
-	"os"
-	"path/filepath"
-
-	"github.com/adityasaky/gittuf/internal/common"
+	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
 )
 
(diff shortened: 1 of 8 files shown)


internal/policy/policy.go
```diff
@@ -15,29 +12,10 @@ const (
 
 // InitializeNamespace creates a git ref for the policy. Initially, the entry
 // has a zero hash.
-// Note: policy.InitializeNamespace assumes the gittuf namespace has been
-// created already.
-func InitializeNamespace() error {
-	repoRootDir, err := common.GetRepositoryRootDirectory()
-	if err != nil {
+func InitializeNamespace(repo *git.Repository) error {
+	if err := repo.Storer.SetReference(plumbing.NewHashReference(plumbing.ReferenceName(PolicyRef), plumbing.ZeroHash)); err != nil {
```
adityasaky, line 16: Should we add a check here to ensure we don't re-init to zero?

adityasaky: Or does it belong in `Repository.InitializeNamespaces`?

wlynch: Sounds like a good idea to me! I'd lean towards checking here, since this is exported callers could call this directly, and checking here would also cover InitializeNamespaces.

adityasaky: Done! Added a test to RSL and policy that checks for this.

internal/policy/policy.go
```diff
@@ -13,31 +12,30 @@ const (
...
-			}
+	if _, err := repo.Reference(plumbing.ReferenceName(PolicyStagingRef), true); err != nil {
+		if !errors.Is(err, plumbing.ErrReferenceNotFound) {
+			return err
 		}
+	} else {
+		return ErrPolicyExists
+	}
+
+	if err := repo.Storer.SetReference(plumbing.NewHashReference(plumbing.ReferenceName(PolicyRef), plumbing.ZeroHash)); err != nil {
```
wlynch, line 36: You can wrap this in a loop to clean this up a tad:

```go
for _, name := []string{PolicyRef, PolicyStagingRef} {
  if _, err := repo.Reference(plumbing.ReferenceName(PolicyRef), true); err != nil {
			return err
		}
	} else {
		return ErrPolicyExists
	}
}
```

wlynch (approved):
One last small thing, but feel free to merge after

internal/policy/policy.go
adityasaky: Done.

PR #41, opened 2023-05-07
Add policy functions

adityasaky:
This PR adds `gittuf policy` to manage top level and delegated targets metadata. It has interfaces to add and remove "rules".


cmd/policy.go
```diff
@@ -0,0 +1,177 @@
...
+
+var (
+	targetsFileName    string
+	ruleName           string
+	authorizedKeyFiles []string
+	rulePatterns       []string
+)
+
+// init for gittuf policy init
+func init() {
```
wlynch, line 43: I would generally limit the usage of init. While this works, Go init ordering is somewhat janky in that it relies on filename ordering, which can be a source of brittleness.

I'd recommend following gitsign/cosign here and having `New()` funcs that get built at main - https://github.com/sigstore/gitsign/blob/main/internal/commands/root/root.go

Having separate files that match the command structure also makes it easier to navigate to the source of each command.

internal/policy/policy.go
```diff
@@ -193,6 +198,32 @@ func LoadStateForEntry(ctx context.Context, repo *git.Repository, e rsl.EntryTyp
 	return state, nil
 }
 
+func (s *State) FindAuthorizedSigningKeyIDs(ctx context.Context, roleName string) ([]string, error) {
```
wlynch, line 201: nit: godoc

internal/policy/policy.go
```diff
@@ -193,6 +198,32 @@ func LoadStateForEntry(ctx context.Context, repo *git.Repository, e rsl.EntryTyp
...
+	if err != nil {
+		return []string{}, err
+	}
+
+	if roleName == RootRoleName {
+		return rootMetadata.Roles[RootRoleName].KeyIDs, nil
+	}
+
+	if roleName == TargetsRoleName {
+		return rootMetadata.Roles[TargetsRoleName].KeyIDs, nil
```
wlynch, line 216: If we want to be safe, we should double check this key returned non-nil before we try to access the KeyIDs field.

internal/policy/policy.go
```diff
@@ -449,3 +489,81 @@ func (s *State) GetRootMetadata() (*tuf.RootMetadata, error) {
...
+		if env, ok := s.DelegationEnvelopes[roleName]; ok {
+			payloadBytes, err = env.DecodeB64Payload()
+		} else {
+			return nil, ErrMetadataNotFound
+		}
+	}
+
+	if err != nil {
+		return nil, err
+	}
```
wlynch, line 511: A good best practice is to simplify conditions and avoid nesting by trying to fail fast around exit cases. i.e. this could be re-written as:

```suggestion
	e := s.TargetsEnvelope
	if roleName != TargetsRoleName {
		env, ok := s.DelegationEnvelopes[roleName]
		if !ok {
			return nil, ErrMetadataNotFound
		}
		e = env
	}
	payloadBytes, err := e.DecodeB64Payload()
	if err != nil {
		return nil, err
	}
```

internal/policy/targets.go
```diff
@@ -0,0 +1,108 @@
...
+func RemoveDelegation(targetsMetadata *tuf.TargetsMetadata, ruleName string) (*tuf.TargetsMetadata, error) {
+	if ruleName == AllowRuleName {
+		return nil, ErrCannotManipulateAllowRule
+	}
+
+	allDelegations := targetsMetadata.Delegations.Roles
+
+	for i, delegation := range allDelegations {
+		if delegation.Name == ruleName {
+			allDelegations = append(allDelegations[:i], allDelegations[i+1:]...)
```
wlynch, line 88: If we expect this to happen multiple times, it might be easier to just scan through the list once and append everything that doesn't match the name to a pre-provisioned list. IIRC the second half of append copies all of the elements one by one because of the `...`.

wlynch (changes requested):

internal/policy/targets.go
adityasaky:
```
	allDelegations := targetsMetadata.Delegations.Roles
	updatedDelegations := []tuf.Delegation{}

	for _, delegation := range allDelegations {
		if delegation.Name != ruleName {
			updatedDelegations = append(updatedDelegations, delegation)
		}
	}
	targetsMetadata.Delegations.Roles = updatedDelegations
```

Do you mean like so?

adityasaky: Maybe my math is off but because it was previously matched on `== ruleName`, wasn't the append happening on the same number of elements total?

cmd/policy.go
adityasaky: I've refactored `cmd` as a whole in the very last commit. Happy to submit it in a separate PR as it also modifies the `trust` subcommand and drops the `dev` one etc.

wlynch (approved):

PR #50, opened 2023-05-20
Try to clarify more of the recovery text and goals.

JustinCappos:


adityasaky:
@JustinCappos I added a commit with some clarifications.

docs/specification.md
```diff
@@ -372,28 +385,58 @@ repeated.
...
+affected by the compromise can skip those changes altogether.
 
-### Recovering from a developer compromise
+### Handling the removal of information
 
-If a developer's keys are compromised and used to make changes to the
-repository, maintainers must immediately sign updated policies revoking their
-keys. Further, maintainers may reset the states of the affected refs and sign
-new RSL entries with corrected states.
+When files are deleted, for example ...
```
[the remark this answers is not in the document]
JustinCappos: What I mean is if some or all of the gittuf metadata is deleted by an unauthorized user.  

The first case you mention I think is fairly straightforward to handle if we always have gittuf metadata as part of resolving what to do if someone not using gittuf breaks the access control restrictions.  The second case is likely impractical to handle given how things work with git today.

docs/specification.md
```diff
@@ -372,28 +385,58 @@ repeated.
...
+gittuf clients would generate.
+
 ### Recovering from accidental changes and pushes
 
-There are several scenarios here. If a user makes changes locally and tries to
-push them to the blessed copy, it should be quite easy to detect and reject the
-changes. A pre-receive hook on the server can be employed to ensure the client
-is also sending valid metadata for the set of changes. If not, the operation
-must be terminated.
+A trusted user makes changes to a branch protected by gittuf policies. The user
```
JustinCappos, line 400: Needs set up to be clear about what is happening here.  Also, what a "trusted user" is, isn't clear.

docs/specification.md
```diff
@@ -370,30 +373,112 @@ when verification was in progress locally. If this is the case, the provisional
...
+with signed pushes?) and include some evidence about why B thinks the change
+came from A (TODO: signed push can be evidence option?).
+
+TODO: decide if B needs to be authorized for the affected branch to sign for A.
+
+### Recovery Scenarios
+
+These scenarios are some examples where recovery is necessary.
+
+#### A change is made without an RSL entry
```
JustinCappos, line 429: What is the order of operations when it comes to an RSL entry and a commit?  Which comes first?  If it is RSL first, then we may also need to discuss handling the case where only the RSL entry appears.  This could occur with a legitimate client when it or the repo crashes / loses network connectivity in the middle of an action.

docs/specification.md
```diff
@@ -370,30 +373,112 @@ when verification was in progress locally. If this is the case, the provisional
...
+
+Bob does not use gittuf and pushes to a branch. Alice notices this as her gittuf
+client detects a push to the branch without an accompanying RSL entry. She
+validates that the change came from Bob and creates an RSL entry on his behalf,
+identifying him and including information about how she verified it was him.
+Alice applies M2.
+
+#### An incorrect RSL entry is added
+
```
JustinCappos, line 438:
```suggestion

There are several ways in which an RSL entry can be considered "incorrect".  
```

docs/specification.md
```diff
@@ -370,30 +373,112 @@ when verification was in progress locally. If this is the case, the provisional
...
+If an entry is malformed (structurally), Git should catch it if it's not a valid
+commit. In addition, gittuf must catch any valid commits with malformed RSL
+data. For example, a necessary field may be missing, or may contain invalid
+data. gittuf must reject such entries and warn the user when they are
+encountered. As this scenario requires the author's client to be buggy or
+malicious, it cannot be trusted to catch the invalid entries when the actor
+pushes to the main copy of the repository. However, correct gittuf clients
+verify that RSL entries are not invalid when they pull from the main copy of
```
JustinCappos, line 446: This sentence is important.  I would restructure this paragraph to say this up front and be clear about it.  

One idea (which retains your notion that git will reject some malformed actions), is to say up front that there are two classes of problems, those which git will reject and those which are valid according to git, but not gittuf.  You can then talk about how to handle each case.

docs/specification.md
```diff
@@ -370,30 +373,112 @@ when verification was in progress locally. If this is the case, the provisional
...
+the repository. If the main copy is enabled with gittuf, such incidents can be
+caught before other users receive the invalid changes.
```
JustinCappos, line 448: Is there any type of invalid RSL entry which a gittuf client will not be able to detect upon pull?

docs/specification.md
```diff
@@ -370,30 +373,112 @@ when verification was in progress locally. If this is the case, the provisional
...
+
+If the main copy is not enabled with gittuf, when a user detects such an
```
JustinCappos, line 450:
```suggestion
If the repository is not enabled with gittuf, when a user detects such an
```

Also, I'd suggest removing some of the negatives here.

docs/specification.md
```diff
@@ -370,30 +373,112 @@ when verification was in progress locally. If this is the case, the provisional
...
+incident, they can use the mechanisms described in M1 to invalidate the
+incorrect entry.
+
+Bob has a buggy gittuf client and pushes an invalid entry to the main copy of
+the repository. Alice pulls and receives a warning from her gittuf client. Alice
+reverses Bob's changes, creating an RSL entry for the affected branch if needed,
+and includes an RSL annotation skipping Bob's RSL entry.
```
JustinCappos, line 457: Can Bob arbitrarily pollute the RSL with garbage if he is malicious / buggy?  Do we care?  Why or why not?

docs/specification.md
```diff
@@ -370,30 +373,112 @@ when verification was in progress locally. If this is the case, the provisional
...
+
+#### A gittuf access control policy is violated
+
+Bob creates an RSL entry for a branch he's not authorized for by gittuf policy.
+He pushes a change to that branch.  Alice notices this and (TODO: decide if
+alice needs to be authorized) she's authorized by policy for the affected
+branch. Alice reverses Bob's change, creating a new RSL entry reflecting that.
```
JustinCappos, line 464:
```suggestion
alice needs to be authorized). 
Alice reverses Bob's change, creating a new RSL entry reflecting that.
```

docs/specification.md
```diff
@@ -370,30 +373,112 @@ when verification was in progress locally. If this is the case, the provisional
...
+Alice also creates an RSL annotation marking Bob's entry as one to be skipped.
+Alice, therefore, uses M1. Finally, if the policy violation was because of a key
+compromise, the policy is updated to revoke and rotate the affected key.
```
JustinCappos, line 467: How does Alice know if this is due to a key compromise?

docs/specification.md
```diff
@@ -370,30 +373,112 @@ when verification was in progress locally. If this is the case, the provisional
...
+#### Attacker modifies or deletes historical RSL entry
+
+Overwriting or deleting an historical RSL entry is a complicated proposition.
+Git's content addressable properties mean that a SHA-1 collision is necessary to
+overwrite an existing RSL entry in the Git object store. Further, the attacker
+also needs more than push access to the repository as Git will not accept an
+object it already has in its store (TODO: double check). Similarly, deleting an
+entry from the object store preserves the RSL structure cosmetically but
+verification workflows that require the entry will fail.
+
```
JustinCappos, line 478:
```suggestion

#### Dealing with fork* attacks

```

docs/specification.md
```diff
@@ -370,30 +373,112 @@ when verification was in progress locally. If this is the case, the provisional
...
+Alternatively, an attacker may attempt a forking attack where different
+developers receive different RSL states. This will not succeed, however, against
+any developers that already have the modified or deleted RSL entries in their
+local copy of the repository. We also observe that at least one developer must
+have the affected RSL entry--the author of that entry. (TODO: snapshot /
+timestamp?)
```
JustinCappos, line 484: This trails off.  It needs a bit more explanation.  Nothing too deep needs to be said.

docs/specification.md
```diff
@@ -370,30 +373,112 @@ when verification was in progress locally. If this is the case, the provisional
...
+
+#### Attacker modifies or deletes historical RSL entry
+
+Overwriting or deleting an historical RSL entry is a complicated proposition.
+Git's content addressable properties mean that a SHA-1 collision is necessary to
+overwrite an existing RSL entry in the Git object store. Further, the attacker
+also needs more than push access to the repository as Git will not accept an
+object it already has in its store (TODO: double check). Similarly, deleting an
+entry from the object store preserves the RSL structure cosmetically but
+verification workflows that require the entry will fail.
```
JustinCappos, line 477: Okay, but what is the impact here?  You've talked about why it is hard to pull off, but what happens if it is done?

docs/specification.md
adityasaky: The `Actor Workflows` section needs to be rewritten to describe these things: how RSL entries are created, how they are synced between repo copies and so on. For now, I've updated the "incorrect RSL entries" subsection to handle situations where the RSL entry appears without the corresponding change.

docs/specification.md
adityasaky: I can't think of any at the moment. RSL entries don't contain a ton of information so we should be able to verify the validity of whatever they contain...

JustinCappos: Okay, be sure to ensure that things like "the actual commit pointed to is sensible" in addition to "the hash is a hash and points to a commit"..

adityasaky: Enumerated some invalid entry examples.

docs/specification.md
adityasaky: Does it read better now?

docs/specification.md
adityasaky: Rewrote actor workflows above.

docs/specification.md
```diff
@@ -311,89 +432,153 @@ available state entry is `D`:
...
-### Recovering from a developer compromise
-
-If a developer's keys are compromised and used to make changes to the
-repository, maintainers must immediately sign updated policies revoking their
-keys. Further, maintainers may reset the states of the affected refs and sign
-new RSL entries with corrected states.
-
-TODO: evaluate if consecutive state verification fails on clients behind the
-times. Should recovery rewrite non-valid RSL entries? Defeats the purpose?
+gittuf presents mechanisms that can be used in many different scenarios.
```
JustinCappos, line 449: gittuf uses two basic mechanisms for recovery.  We describe these core building blocks for recovery before we discuss the exact scenarios when they are applied and why they provide the desired security properties.

JustinCappos, line 449:
```suggestion
gittuf uses two basic mechanisms for recovery.  We describe these core building blocks for recovery before we discuss the exact scenarios when they are applied and why they provide the desired security properties.
```

Notice how this explains what we are telling them, why, and what things will come later.

docs/specification.md
```diff
@@ -311,89 +432,153 @@ available state entry is `D`:
...
+
+#### M1: Reset to known good state
```
JustinCappos, line 451:
```suggestion
#### M1: Removing RSL content to reset to known good state
```

docs/specification.md
```diff
@@ -311,89 +432,153 @@ available state entry is `D`:
...
+to recover. The client that originally created the invalid entries must first
+switch to a correct implementation of gittuf and then sync the local
+repository's state with the main repository.
+
+If the main repository is also gittuf enabled, such incidents can be caught
+before other users receive the incorrect RSL entries. The repository, though,
+must not behave like a typical gittuf client. Instead, gittuf's repository
+behavior is slightly different as RSL entries are submitted before the changes
+they represent. The repository must wait to receive the full changes rather than
+immediately rejecting the RSL entry. TODO: the repository-specific behavior
```
JustinCappos, line 530: why?  I believe I know, but you don't explain anything here...

docs/specification.md
JustinCappos: I understand, but don't feel like you are convincing about the actual impact this has.  So, if verification workfloes that require the entry fail, what do you do?  Is this okay?

docs/specification.md
```diff
@@ -311,89 +432,153 @@ available state entry is `D`:
...
+
+Also note that while Git uses SHA-1 for its object store, cryptographic 
+signatures are generated and verified using stronger hash algorithms. Therefore,
+a successful SHA-1 collision for an RSL entry will not go undetected as all
+entries are signed.
+
+#### Dealing with fork* attacks
+
+An attacker may attempt a forking attack where different developers receive
+different RSL states. This will not succeed, however, against any developers
```
JustinCappos, line 565: What does it mean to succeed in this context?

docs/specification.md
```diff
@@ -311,89 +432,153 @@ available state entry is `D`:
...
+that already have the modified or deleted RSL entries in their local copy of the
+repository. We also observe that at least one developer must have the affected
+RSL entry--the author of that entry. So, the attacker must maintain at least two
+RSLs, one for the author of the affected entry and one for clients without the
+entry. Also, if multiple users receive the affected entry before the attack, the
```
JustinCappos, line 570: Why does it matter if the attacker needs to maintain at least two RSLs?

docs/specification.md
```diff
@@ -311,89 +432,153 @@ available state entry is `D`:
...
+attacker may not be able to trivially identify which RSL must be sent to each
+user, improving the chances of detection of the attack. Finally, while gittuf
+primarily uses TUF's root of trust and delegations, it is possible that TUF's
+snapshot and timestamp roles can be leveraged to further mitigate forking
+attacks.
+
+#### An authorized key is compromised
+
+When a key authorized by gittuf policies is compromised, it must be revoked and
```
JustinCappos, line 579: why?

docs/specification.md
```diff
@@ -311,89 +432,153 @@ available state entry is `D`:
...
+rotated. gittuf policies that grant permissions to the key must be updated to
+revoke the key, possibly adding the actor's new key in the process. Further, if
+a security analysis shows that the key was used to make malicious changes, those
+changes must be reverted and the corresponding RSL entries signed with the
+compromised key must be skipped.
```
JustinCappos, line 584: why?

docs/specification.md
```diff
@@ -311,89 +432,153 @@ available state entry is `D`:
...
+Invalid entries that are not rejected by Git must be caught by gittuf. Some
+examples of such invalid entries are:
+
+* RSL entry is for a non-existing Git reference
+* Commit recorded in RSL entry does not exist
+* Commit recorded in RSL entry does not match the tip of the corresponding Git
+  reference
+* RSL annotation contains references to RSL entries that do not exist or are not
+  RSL entries (i.e. the annotation points to other commits in the repository)
+
```
JustinCappos, line 512: What about appending random data that doesn't match RSL entries?

docs/specification.md
```diff
@@ -311,89 +432,153 @@ available state entry is `D`:
...
+clients, these entries cannot be detected prior to them being pushed to the main
+repository.
+
+As correctly implemented gittuf clients verify the validity of RSL entries when
+they pull from the main repository, the user is warned if invalid entries are
+encountered. Then, the user can then use M1 to invalidate the incorrect entry.
+Other clients with the invalid entry only need to fetch the latest RSL entries
+to recover. The client that originally created the invalid entries must first
+switch to a correct implementation of gittuf and then sync the local
+repository's state with the main repository.
```
JustinCappos, line 523: Why does the client that made the invalid entries need to do anything different?

docs/specification.md
```diff
@@ -311,89 +432,153 @@ available state entry is `D`:
...
+
+If the main repository is also gittuf enabled, such incidents can be caught
+before other users receive the incorrect RSL entries. The repository, though,
+must not behave like a typical gittuf client. Instead, gittuf's repository
+behavior is slightly different as RSL entries are submitted before the changes
+they represent. The repository must wait to receive the full changes rather than
```
JustinCappos, line 529: Isn't it the case that clients will also have this problem to some extent?  I think overall, we need to consider this problem more holistically.

docs/specification.md
```diff
@@ -311,89 +432,153 @@ available state entry is `D`:
...
+
+Invalid entries that are not rejected by Git must be caught by gittuf. Some
+examples of such invalid entries are:
+
+* RSL entry is for a non-existing Git reference
+* Commit recorded in RSL entry does not exist
+* Commit recorded in RSL entry does not match the tip of the corresponding Git
+  reference
+* RSL annotation contains references to RSL entries that do not exist or are not
+  RSL entries (i.e. the annotation points to other commits in the repository)
```
JustinCappos, line 511: How does the next person that pulls always correctly know the last entry (head) of the RSL?

docs/specification.md
adityasaky: Yes. I think we need to write this up separately in the document and I wanted to talk to you about it. Also because it changes what we discussed before about the order of submitting RSL entries and repo changes.

docs/specification.md
adityasaky: Can you elaborate on "correctly"?

docs/specification.md
adityasaky: Updated to only call out that the broken impl must be fixed / updated.

docs/specification.md
adityasaky: Clients fetch changes to refs during the verification workflow based on what the RSL says. While they do have to wait, they're not in exactly the same situation as the repo.

I want to sketch this out for the repo some more (and in part this is based on whether the repo can be gittuf enabled or not). I think that using [atomic pushes)[https://git-scm.com/docs/git-push#Documentation/git-push.txt---no-atomic) mostly solves this problem for us.

docs/specification.md
JustinCappos: I mean, how do I find the head of the RSL?  Can I be wrong / confused in some way?  How?

adityasaky: Currently, the head of the RSL is just requested from `refs/gittuf/reference-state-log`. This opens the possibility that an attacker who controls that copy of the repo can freeze or set HEAD to an older RSL commit (rollback). This is in part discussed in the forking attack section as at least one developer knows the RSL's true state. In addition, we have to talk about whether we're introducing snapshot and timestamp roles.

If we go that direction, I think the snapshot role only needs to record the HEAD of the RSL, as the targets roles can be derived from the latest RSL entry for the policy namespace. In this case, the snapshot and timestamp would not be stored with the rest of the TUF metadata, but could still be in the repo, say `refs/gittuf/snapshot`.

adityasaky: Actually, only timestamp is enough if it directly records the RSL head. The RSL has snapshot properties already.

JustinCappos: Okay, this is interesting.  I do think that we should consider using timestamp in this way.  It's also worth thinking about who actually holds that key and signs in this case though.  Why do we trust them and how is this done if the repo isn't gittuf enabled?

adityasaky: TODO: Note how RSL replaces the snapshot role and how timestamp can be used to record the tip of the RSL.

adityasaky: From IRL meeting: including timestamp has very mild benefits, we should leave it out. AI: discuss snapshot <-> RSL properties.

adityasaky:
@JustinCappos this PR has evolved a bit. I'm going to move the non recovery changes to a separate branch and submit tomorrow. I'd like to merge the parts we're happy with such as the recovery and so on.

docs/specification.md
adityasaky: #66

docs/specification.md
adityasaky: Added some text to the RSL subsec above re its snapshot properties.

docs/specification.md
adityasaky: #66 for the repository-specific behavior.

docs/specification.md
adityasaky: Updated.

docs/specification.md
adityasaky: Updated.

docs/specification.md
adityasaky: Indicated that detection is followed by restoring the affected repository from other, local copies of the repository.

docs/specification.md
```diff
@@ -83,7 +86,18 @@ forking attack.
 The RSL is tracked at `refs/gittuf/reference-state-log`, and is implemented as a
 distinct commit graph. Each commit corresponds to one entry in the RSL, and
 standard Git signing mechanisms are employed for the actor's signature on the
-RSL entry.
+RSL entry. The latest entry is identified using the tip of the RSL Git ref.
+
+Finally, as the RSL records changes to other Git refs in the repository, it
```
JustinCappos, line 91:
```suggestion
Note that the RSL and liveness of the repository in Git remove the need for some traditional TUF roles.
As the RSL records changes to other Git refs in the repository, it
```

JustinCappos (commented):
It looks fine, but I'd recommend a change like the one I suggested.  Feel free to fix the grammar in that sentence or elsewhere, as it is a bit awkward.

adityasaky (approved):

PR #52, opened 2023-05-28
Add verification functions

adityasaky:
~Blocked by #41~

Addresses #15


internal/policy/policy.go
```diff
@@ -170,10 +172,38 @@ func LoadStateForEntry(ctx context.Context, repo *git.Repository, e rsl.EntryTyp
 
 		if entry.Name == fmt.Sprintf("%s.json", RootRoleName) {
 			state.RootEnvelope = env
+			rootMetadata, err := state.GetRootMetadata()
+			if err != nil {
+				return nil, err
+			}
+			for _, key := range rootMetadata.Keys {
+				key := key // avoid iterator pointer issues
+				key.ID()   //nolint:errcheck
```
adityasaky, line 181: @wlynch any thoughts on how we can ensure keyID is set when we're pulling tuf.Key out of the metadata outside of including KeyID in the serialized form?

internal/policy/verify_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+	entry := rsl.NewEntry("refs/heads/main", plumbing.ZeroHash)
+	entryID := createTestRSLEntryCommit(t, repo, entry)
+	entry.ID = entryID
+
+	err := verifyEntry(context.Background(), repo, state, entry)
+	assert.Nil(t, err)
+}
+
+func createTestRSLEntryCommit(t *testing.T, repo *git.Repository, entry *rsl.Entry) plumbing.Hash {
```
adityasaky, line 37: this entire test file is a mess with a lot of repetition from other methods that are private.

internal/policy/verify.go
```diff
@@ -0,0 +1,140 @@
...
+	// entryStack has a list of RSL entries in reverse order
+	entryQueue := []*rsl.Entry{}
+	for j := len(entryStack) - 1; j >= 0; j-- {
+		// TODO: instead of reversing it, we can likely just protest this in
+		// reverse order but the policy states are the issue
+		entryQueue = append(entryQueue, entryStack[j])
+	}
+
+	for _, entry := range entryQueue {
+		// FIXME: we're not verifying policy RSL entry signatures because
```
adityasaky, line 81: This probably deserves a separate issue. Fundamentally, the question is if RSL entries for the policy namespace should be verified for signatures. If we do, we must allow _any_ public key in the delegations tree to make changes to the namespace. In that case, we run into issues such as what key signs an RSL entry for the policy namespace if the changes were to root.json?

On the other hand, we always run `state.Verify` which checks that the policy state is internally consistent, including all reachable metadata signatures starting from root.json. We then have the problem of where we get the root keys from. Currently, it looks into the keys embedded in the repository but these can be easily manipulated if the RSL entry for the policy namespace is not verified. Thoughts? If we use a special case to verify signatures on diffs to root keys / root.json, we again have the question of whether keys used for root.json can also sign git commits for the RSL entry

@JustinCappos @wlynch @znewman01

adityasaky: As git supports ssh keys, the last point may not be a problem 🤔

adityasaky:
@wlynch This is missing some tests and cleaning up but I'd appreciate an initial look from you!

internal/policy/policy.go
```diff
@@ -170,10 +172,38 @@ func LoadStateForEntry(ctx context.Context, repo *git.Repository, e rsl.EntryTyp
...
+			state.DelegationEnvelopes[roleName] = env
+			targetsMetadata, err := state.GetTargetsMetadata(roleName)
+			if err != nil {
+				return nil, err
+			}
+			for _, key := range targetsMetadata.Delegations.Keys {
+				key := key // avoid iterator pointer issues
+				key.ID()   //nolint:errcheck
+				state.AllPublicKeys = append(state.AllPublicKeys, key)
+			}
```
wlynch, line 206: I'd add comments explaining the differences between these 2 (they look very similar).

internal/policy/policy.go
```diff
@@ -230,6 +260,50 @@ func (s *State) FindAuthorizedSigningKeyIDs(ctx context.Context, roleName string
...
+
+			if s.HasTargetsRole(delegation.Name) {
+				delegatedMetadata, err := s.GetTargetsMetadata(delegation.Name)
+				if err != nil {
+					return nil, err
+				}
+				for keyID, key := range delegatedMetadata.Delegations.Keys {
+					allPublicKeys[keyID] = key
+				}
+				delegationsQueue = append(delegationsQueue[:len(delegationsQueue)-1], delegatedMetadata.Delegations.Roles...)
```
wlynch, line 301: This is taking all but the last element of the queue and then adding the new target roles.

It's not immediately obvious to me why this is correct (e.g. why are we ignoring the last element) - I'd add comments explaining this in more detail.

internal/policy/policy.go
```diff
@@ -230,6 +260,50 @@ func (s *State) FindAuthorizedSigningKeyIDs(ctx context.Context, roleName string
...
+
+			// TODO: clarify logic for delegation.Terminating
+
+			if s.HasTargetsRole(delegation.Name) {
+				delegatedMetadata, err := s.GetTargetsMetadata(delegation.Name)
+				if err != nil {
+					return nil, err
+				}
+				for keyID, key := range delegatedMetadata.Delegations.Keys {
+					allPublicKeys[keyID] = key
```
wlynch, line 299: Might be worth adding more comments explaining why we're doing this (it looks like this is some sort of override?)

Is it a problem if this key id has already been pulled out into trustedkeys from a previous iteration of the delegation queue? (is this even possible?)

internal/policy/verify.go
```diff
@@ -0,0 +1,140 @@
...
+	for {
+		if iteratorEntry.GetID() == firstEntry.ID {
+			break
+		}
+
+		parentEntryTmp, err := rsl.GetParentForEntry(repo, iteratorEntry)
+		if err != nil {
+			return err
+		}
+		parentEntry := parentEntryTmp.(*rsl.Entry) // TODO: handle annotations
```
wlynch, line 63: Type switching is usually a signal that the interface may not be what you want - e.g. why have `GetParentForEntry` take in a `EntryType` instead of a `*Entry`? 🤔

internal/policy/verify.go
```diff
@@ -0,0 +1,140 @@
...
+
+		if parentEntry.RefName == target || parentEntry.RefName == PolicyRef {
+			entryStack = append(entryStack, parentEntry)
+		}
+
+		iteratorEntry = parentEntry
+	}
+
+	// entryStack has a list of RSL entries in reverse order
+	entryQueue := []*rsl.Entry{}
```
wlynch, line 73:
```suggestion
	entryQueue := make([]*rsl.Entry, 0, len(entryStack))
```

internal/policy/verify.go
```diff
@@ -0,0 +1,140 @@
...
+	for j := len(entryStack) - 1; j >= 0; j-- {
+		// TODO: instead of reversing it, we can likely just protest this in
+		// reverse order but the policy states are the issue
```
wlynch, line 76: I don't fully understand this

internal/policy/verify.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	for _, key := range trustedKeys {
+		if err := gitinterface.VerifyCommitSignature(commitObj, key); err != nil {
+			if !errors.Is(err, gitinterface.ErrIncorrectVerificationKey) {
+				return err
+			}
+		} else {
+			return nil
+		}
```
wlynch, line 136: This is probably a case where you may want to reverse the check to make it clearer under what conditions the loop will continue.

```suggestion
		err := gitinterface.VerifyCommitSignature(commitObj, key)
		if err == nil {
			// Key is valid
			return nil
		}
		if !errors.Is(err, gitinterface.ErrIncorrectVerificationKey) {
			return err
		}
		// Haven't found a valid key - continue
```

internal/repository/verify.go
```diff
@@ -0,0 +1,54 @@
...
+	// Check if branch
+	refName := plumbing.NewBranchReferenceName(target)
+	_, err := repo.Reference(refName, false)
+	if err != nil {
+		if !errors.Is(err, plumbing.ErrReferenceNotFound) {
+			return "", err
+		}
+	} else {
+		return string(refName), nil
+	}
```
wlynch, line 40: Similar comment here - prefer exiting early rather than nesting the continue condition:

```suggestion
	if err == nil {
		return string(refName), nil
	}
	if !errors.Is(err, plumbing.ErrReferenceNotFound) {
		return "", err
	}
```

internal/rsl/rsl.go
```diff
@@ -152,6 +161,38 @@ func (a Annotation) createCommitMessage() (string, error) {
...
+		}
+
+		if len(commitObj.ParentHashes) == 0 {
+			// Found first entry
+			tmpEntry, err := parseRSLEntryText(commitObj.Hash, commitObj.Message)
+			if err != nil {
+				return nil, err
+			}
+
+			entry := tmpEntry.(*Entry)
```
wlynch, line 184: Will this ever not be a `*Entry`?

If yes, consider changing `parseRSLEntryText` to return a `*Entry`.
If no, add a type check to handle the case if a non-`*Entry` is returned.

wlynch (commented):
Overall, looks good! Few minor tweaks.

There seems to be ~3 separate changes in here:

1. Adding the verification.
2. Adding RSL commands.
3. Adding support for GPG keys.

Consider breaking these up into separate PRs?

internal/rsl/rsl.go
adityasaky: Good point, the first entry is expected to be `*Entry` and not `*Annotation` because it's always for the policy. I'll change the return type and pull out the RSL specific changes as a separate PR

internal/policy/policy.go
adityasaky: The issue was with how keys were being serialized in metadata. It was `map[keyID]: key` except `key` didn't have `.KeyID` set. So when reading in a key from metadata we'd either have to recalculate it or do some processing when looping through the map to use the key there. Either way, it was messy. In #61, I've just included the serializing of key IDs in the map. This actually aligns with in-toto as it happens but perhaps not some TUF implementations. Also, #61 offloads this to go-securesystemslib so we don't have to keep tweaking things here.

internal/policy/policy.go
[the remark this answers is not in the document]
adityasaky: (Addressing outstanding comments for #62.)

To confirm, you're talking about when keyID is already in the map and we're writing to it again? I _think_ that's fine as the keyIDs are uniquely calculated. So while map[keyID] may get overwritten, the values should be the same?

adityasaky: (Addressing outstanding comments for https://github.com/adityasaky/gittuf/pull/62.)

I'm tracking all the public keys encountered so that when we're on a delegation matched path, we can look up the keys from the global map to populate trusted keys. This is because gittuf differs from regular TUF where all keys along a matched delegation path are trusted rather than just the keys listed in the leaf delegation. The overwriting shouldn't impact the _value_ of the key which matters. trustedKeys may end up with duplicates but that doesn't affect threshold verification luckily. That said, I can swap that out to also be a map[key]bool to ape set behavior.

internal/policy/verify.go
adityasaky: I've swapped it to be *Entry but we'll still have to handle annotations. Right now we don't create any annotations but in future, we'd also be tracking them here instead of always assuming it's *Entry. Then, we'd check the annotations to see if other entries are to be skipped.

internal/policy/verify.go
adityasaky:
```
// We reverse the entries so that they are chronologically sorted. If we
// process them in the reverse order, we have to go past each entry
// anyway to find the last policy entry to use. It also makes the entry
// processing easier to reason about.
```

Updated the comment.

internal/policy/verify.go
adityasaky: Much cleaner, thanks!

adityasaky:
Closed in favour of #62.

PR #61, opened 2023-06-20
Cleans up key management and adds GPG key support

adityasaky:
This PR pulls some changes out of #52 and includes some other refactoring. Specifically, the locally defined ed25519 support is removed in favour of go-securesystemslib.


adityasaky:
I think this once and for all addresses the key ID issue as well. Earlier, the key ID was not part of the serialized key in the metadata which led to redundant calculations. Now, it's included. It's a bit repetitive when looked at in the metadata but the code is probably cleaner this way.

PR #65, opened 2023-07-12
Update actor workflows

adityasaky:
Originally introduced in #50.

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -1,6 +1,6 @@
 # gittuf Specification
 
-Last Modified: June 27, 2023
+Last Modified: July 11, 2023
 
 Version: 0.1.0-draft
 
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -235,6 +235,133 @@ repository. Clara and Dave are only authorized to work on files within `bar/*`
...
+gittuf clients uses the `origin` Git remote to identify the main repository. As
+the RSL must be linear with no branches, gittuf employs a variation of the
+`Secure_Fetch` and `Secure_Push` workflows described in the RSL academic paper.
+
+#### RSLFetch: Receiving remote RSL changes
+
+Before local RSL changes can be made or pushed, it is necessary to verify that
+they are compatible with the remote RSL state. If the remote RSL has entries
+that are unavailable locally, entries made locally will be rejected by the
+remote.
```
JustinCappos, line 317: I believe I understand, but a simple example may help.

docs/specification.md
```diff
@@ -235,6 +235,133 @@ repository. Clara and Dave are only authorized to work on files within `bar/*`
...
+   successfully.
+1. Perform the verification workflow for the new entries in the remote RSL,
+   incorporating remote changes to the local policy namespace. The verification
+   workflow is performed for each Git reference in the new entries, relative to
+   the local state of each reference. If verification fails, abort and warn
+   user. Note that the verification workflow must fetch each Git reference to
+   its corresponding remote tracker, `refs/remotes/origin/<ref>`. TODO: discuss
+   if verification is skipped for entries that work with namespaces not present
+   locally.
+1. For each modified Git reference, update the local state.
```
JustinCappos, line 333: Perhaps expand this a little?

docs/specification.md
```diff
@@ -235,6 +235,133 @@ repository. Clara and Dave are only authorized to work on files within `bar/*`
...
+1. Set local RSL to the remote RSL's tip.
+
+#### RSLPush: Submitting local RSL changes
+
+1. Execute `RSLFetch` repeatedly until there are no new RSL entries in the
+   remote RSL. Every time there is a remote update, the user must be prompted to
```
JustinCappos, line 339: Why is it necessary to prompt the user?  Could this be automated?

docs/specification.md
```diff
@@ -235,6 +235,133 @@ repository. Clara and Dave are only authorized to work on files within `bar/*`
...
+   fetch and re-apply their changes to the RSL. In addition, the user may also
+   have to re-apply their changes to other Git references that may have changed
```
JustinCappos, line 341: When and why?

docs/specification.md
```diff
@@ -235,6 +235,133 @@ repository. Clara and Dave are only authorized to work on files within `bar/*`
...
+   in the remote. Changes to the gittuf policy must be fetched and applied
+   locally.
+1. Verify the validity of the RSL entries being submitted using locally
+   available gittuf policies. If verification fails, abort and warn user.
```
JustinCappos, line 345: Is this for cases like Bob didn't make an RSL entry because his client isn't gittuf enabled?

docs/specification.md
adityasaky: updated. It could be automated but conflicts make it hard to guarantee always. I suspect this will change a bit once we start implementing, I can note these are drafts until we support them.

docs/specification.md
adityasaky: No, this is for gittuf clients only. This is more for cases where Bob accidentally tries to push changes to something he's not authorized to by policy.

adityasaky: updated

JustinCappos (approved):
I think this is in the right spirit.  The interface is not as smooth as we can perhaps make it, but otherwise it's a good addition.  I support merging and then iterating on the interface later.

PR #68, opened 2023-07-14
Document detached authorization ideas

adityasaky:
WIP

### DIFF
--- a/docs/extensions/detached-authorization.md
+++ b/docs/extensions/detached-authorization.md
@@ -1,0 +1,143 @@
+# Supporting Detached Authorizations for Actor Actions
+
+Last Modified: July 13, 2023
+
+Version: early draft, not ready for 0.1.0
+
+Status: eventually-merge-into-specification
+
+Certain actions in repositories need to be approved by multiple actors. In a
+more general sense, TUF supports these notions using a "threshold" of
+signatures, wherein a minimum number of signatures from an authorized set of
+keys may be required for some role. Supporting this functionality in Git
+repositories is complicated by the following factors:
+* Git objects that support signatures can only embed a single signature.
+* Git is typically (anecdotally?) used more asynchronously than a typical TUF
+  deployment.
+
+Note: it is possible that the final design of a detached "authorization" is an
+application of a more general detached "signature". For now, the two are
+designed in sync.
+
+## Example Scenario
+
+Consider a policy that spans the Git ref namespace and the files namespace. For
+example, a repository may only allow Alice and Bob to merge to the main branch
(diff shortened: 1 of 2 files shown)


reza-curtmola:
This is well written and sounds like a reasonable way to move forward with handling policies that combine the ref and file namespaces. I wonder how does a detached authorization factor into the scenario where, after merging changes to 'foo' directory, someone branches from main and the authorization given by Carol was only valid for the main branch. How will verification go for this other branch about the past change to 'foo'?

Regarding revocation: I didn't carefully consider all the scenarios you raised in questions, but this is a difficult problem in general. If we want to support revocation for detached authorizations, I favor limiting the possible options and choosing simplicity.

adityasaky:
> How will verification go for this other branch about the past change to 'foo'?

There is indeed a question of how much usability overhead this adds to creating new branches. I'm currently wary of allowing policies on files across all Git refs as Carol would continually be expected to provide authorizations for users generating new feature branches and so on. The reason I haven't listed it here is that it's technically detached from how authorizations themselves work and applies more to how policy rules are written for file + git ref namespaces.

adityasaky:
Added some more thoughts: https://github.com/adityasaky/gittuf/pull/68/files#diff-845fa320653e13c6663fa01f7240194636a8c8a6e25a3cbab45b8e55f225a61eR88-R113

adityasaky:
More updates! I've added an "Alternative Option" that discusses using multiple RSL entries instead of a separate authorization mechanism. Let me know what you think, @JustinCappos and @reza-curtmola, based on what we decide, I'll start implementing it in #77 and its follow ups.

adityasaky:
Moved multiple RSL entries option to a separate document with more details and some example scenarios that discuss how clients must behave. I have some reservations but I think this may end up being less complicated in the long run.

JustinCappos:
I had a thought.  Instead of adding a new field to check, would it make sense to only check RSL entries when the commit, etc. is actually created?  Then if Alice and Bob both need to have their sign off checked, Alice could just make the RSL entry and Bob could do his + the commit?

JustinCappos:
I'm going to ask a dumb question.  Why do you need to sign the commit ID?
Is it because you need to know where it fits into the Merkle tree for the
RSL?

adityasaky:
Yes. A standard RSL entry basically says "<BRANCH> is at <COMMIT ID>" and this is then signed by the person who sets the branch to that commit ID. The proposal here overloads the entry to say "I agree with <BRANCH> being set at <COMMIT ID>".

adityasaky:
Closing in favour of #195.

PR #72, opened 2023-08-01
Early gitsign support

adityasaky:


adityasaky:
@wlynch I think this is ready for an early pass while I also work out how to test gitsign sigs without calling out to rekor.

internal/gitinterface/commit.go
```diff
@@ -79,7 +86,60 @@ func VerifyCommitSignature(commit *object.Commit, key *tuf.Key) error {
 
 		return nil
 	case signerverifier.FulcioKeyType:
-		return ErrUnknownSigningMethod // TODO: implement
+		root, err := fulcioroots.Get()
+		if err != nil {
+			return errors.Join(ErrVerifyingSigstoreSignature, err)
```
wlynch, line 91: If you don't want to bump to Go 1.20 you could do something like this instead https://go.dev/play/p/L5P9FiIZS07

wlynch (approved):
Sigstore unit testing aside, this lgtm!

internal/gitinterface/commit.go
adityasaky: Yeah, I think I'll keep 1.19 around for a bit longer.

PR #77, opened 2023-08-08
Supporting dual namespaces in delegations

adityasaky:
This is a work in progress to support delegations combining Git namespaces and file namespaces.


adityasaky:
Closing this PR in favor of #82. We will revisit dual namespace policies with attestations support.

PR #80, opened 2023-08-18
Support git diff

adityasaky:
This breaks away some changes from #77.

Specifically, the changes here define a mechanism to get the list of changed file paths between two commits and the ability to derive this from an RSL entry for some namespace.

Note: I've included some test clean up as I noticed I was defining the same config, clock etc a bunch of times. I can submit that separately if needed. That's all in the last commit.

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -20,6 +20,34 @@ import (
 	"github.com/jonboulle/clockwork"
 )
 
+const (
+	testName  = "Jane Doe"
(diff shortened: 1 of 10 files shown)


wlynch (approved):
I'm assuming some of this is in an in-between state (e.g. because we're not really using this meaningfully in verify yet), but otherwise lgtm!

adityasaky:
yep, the next step is to use it meaningfully for file-namespace policies.

PR #84, opened 2023-08-29
Update readme

adityasaky:
Updates the readme to acknowledge we're closer to an alpha release. Also removes mention of the original prototype as that's no longer relevant.


JustinCappos (approved):
LGTM

PR #90, opened 2023-08-31
Add initial changelog

adityasaky:
Also adds SECURITY.md, some edits to contributor workflow, and a badge from badgeapp.


JustinCappos (approved):
Looks good.  I'm hopeful we will get our badging done sooner rather than later!

adityasaky:
We've met the passing criteria with this change! We're also well on our way to the silver tier.

PR #99, opened 2023-09-04
(WIP) Syncing

adityasaky:

PR #100, opened 2023-09-04
Plumbing for syncing functions

adityasaky:
This PR is a subset of the changes in #99. It adds some of the necessary plumbing for syncing gittuf namespaces to different namespaces.


internal/gitinterface/sync.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	return FetchRefSpec(ctx, repo, remoteName, refSpecs)
+}
+
+// CloneAndFetch clones a repository using the specified URL and additionally
+// fetches the specified refs. If trackRemote is set, the refs are fetched to a
+// local remote tracker.
+func CloneAndFetch(ctx context.Context, remoteURL, dir, initialBranch string, refs []string, trackRemote bool) (*git.Repository, error) {
+	repo, err := git.PlainCloneContext(ctx, dir, false, &git.CloneOptions{URL: remoteURL, ReferenceName: plumbing.ReferenceName(initialBranch)})
```
wlynch, line 115: I'd recommend avoiding the initial branch name if you can, or use HEAD. It'll save headache down the road when you deal with `main` vs `master` branches.

adityasaky: I'll test this some more. My thought was `initialBranch` can be set to "" defaulting to whatever is the main branch on the remote.

adityasaky: Done! Added a test to both Clone methods for when initialBranch is an empty string. When creating clone options, I'm explicitly checking length before setting it, I can confirm go-git defaults to HEAD when the initial branch is not set.

adityasaky: This was only to simulate `git clone -b` so I think we're fine now.

adityasaky:
Thanks, @wlynch!

PR #101, opened 2023-09-05
Add verify-commit

### LINKED ISSUE #15: Distribute public keys for signed commits / tags

adityasaky:
Repurposing this PR to introduce verify-commit functionality. I have some draft code for verify-tag but it's a little more nuanced because a tag is both an object and a ref (so has RSL implications). For now, I think verify-commit is a go?

---

Original message:

WIP, needs more tests and clones some code also introduced in #82 and #100. Must only be merged after those.

TODO:
1. should we have a `gittuf policy add-key` or so that only adds a key somewhere without defining more granular policies? That'll allow for using gittuf only for simple signature checks.
2. should there be an optional --branch parameter for verify-commit so that it uses a specific policy state instead of all of them? Right now we have to use keys from all states, including revoked keys as we don't know when a commit was introduced wrt the RSL.
3. I think we may have a more broad `verify-merge` use case, on that note, for a commit <-> branch. In that case, we'd look ahead, find the RSL entry that introduced the commit to the branch, and verify that entry specifically.

Fixes #15

### DIFF
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -223,6 +223,55 @@ func GetStateForCommit(ctx context.Context, repo *git.Repository, commit *object
 	return LoadStateForEntry(ctx, repo, commitPolicyEntry)
 }
 
+// PublicKeys returns all the public keys associated with a state.
+func (s *State) PublicKeys() (map[string]*tuf.Key, error) {
+	allKeys := map[string]*tuf.Key{}
+
+	// Add root keys
+	for _, key := range s.RootPublicKeys {
+		key := key
+		allKeys[key.KeyID] = key
+	}
+
+	// Add keys from the root metadata
+	rootMetadata, err := s.GetRootMetadata()
+	if err != nil {
+		return nil, err
+	}
+	for keyID, key := range rootMetadata.Keys {
+		key := key
+		allKeys[keyID] = key
+	}
+
+	// Add keys from top level targets metadata
+	if s.TargetsEnvelope == nil {
+		// Early states where this hasn't been initialized yet
+		return nil, err
+	}
+	targetsMetadata, err := s.GetTargetsMetadata(TargetsRoleName)
+	if err != nil {
+		return nil, err
+	}
+	for keyID, key := range targetsMetadata.Delegations.Keys {
+		key := key
+		allKeys[keyID] = key
+	}
+
+	// Add keys from delegated targets metadata
+	for roleName := range s.DelegationEnvelopes {
+		delegatedMetadata, err := s.GetTargetsMetadata(roleName)
+		if err != nil {
+			return nil, err
+		}
+		for keyID, key := range delegatedMetadata.Delegations.Keys {
+			key := key
+			allKeys[keyID] = key
+		}
+	}
+
+	return allKeys, nil
+}
+
 // FindAuthorizedSigningKeyIDs traverses the policy metadata to identify the
 // keys trusted to sign for the specified role.
 func (s *State) FindAuthorizedSigningKeyIDs(ctx context.Context, roleName string) ([]string, error) {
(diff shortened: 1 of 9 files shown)


internal/policy/policy.go
```diff
@@ -223,6 +223,55 @@ func GetStateForCommit(ctx context.Context, repo *git.Repository, commit *object
 	return LoadStateForEntry(ctx, repo, commitPolicyEntry)
 }
 
+// Keys returns all the public keys associated with a state.
+func (s *State) Keys() (map[string]*tuf.Key, error) {
```
wlynch, line 227: nit: do we want to call this `PublicKeys` so there's no ambiguity for what kind of keys this is returning?

internal/policy/verify.go
```diff
@@ -129,6 +140,93 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
 	return nil
 }
 
+// VerifyCommit verifies the signature on the specified commits (identified by
+// their hash or via a reference that is resolved). For each commit, the policy
+// applicable when the commit was first recorded (directly or indirectly) in the
+// RSL is used.
+func VerifyCommit(ctx context.Context, repo *git.Repository, ids []string) map[string]string {
```
wlynch, line 147:
```suggestion
func VerifyCommit(ctx context.Context, repo *git.Repository, ids ...string) map[string]string {
```

Might be a good place for a variadic so you don't always have to pass in an array if you only want to verify a single commit.

wlynch, line 147: Document what the return type is and how to interpret it.

e.g. it would be invalid if a commit isn't captured in the status. There also isn't a good way to separate out valid statuses from invalid statuses without inspecting the string (we may want to split these into valid/invalid statuses).

internal/policy/verify_test.go
```diff
@@ -81,6 +86,64 @@ func TestVerifyRelativeForRef(t *testing.T) {
...
+	assert.Equal(t, expectedStatus, status)
+
+	// Try a tag
+	tagHash, err := gitinterface.Tag(repo, commitIDs[len(commitIDs)-1], "v1", "Test tag", false)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expectedStatus = map[string]string{tagHash.String(): nonCommitMessage}
+	status = VerifyCommit(context.Background(), repo, []string{tagHash.String()})
```
wlynch, line 136: nit: define the ctx at the top of the test and reuse (this makes it easier to add context info later if needed)

internal/policy/verify.go
adityasaky: I've updated the docstring to indicate how the status is to be parsed. Right now it's meant only for human consumption when they explicitly just use `verify-commit`. In the rest of the gittuf workflow, we don't currently want / need this. I do agree, though, that it may be worth breaking this up to return program-consumable statuses, with messages created at the CLI or repo package when it's for human consumption. Shall I track that as a ticket?

wlynch (approved):

adityasaky:
Thanks for the review!

PR #105, opened 2023-09-07
(WIP) Object store hash algorithm agility

adityasaky:
This is a proof of concept implementation for maintaining SHA-256 hashes for the default SHA-1 hashes. The hashes are calculated recursively for all objects, replacing the original SHA-1 hashes where needed in trees, commits, and tags.

### DIFF
--- a/internal/gitinterface/agility.go
+++ b/internal/gitinterface/agility.go
@@ -1,0 +1,224 @@
+package gitinterface
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/go-git/go-git/v5"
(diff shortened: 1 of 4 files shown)


adityasaky:
1) Added some scaffolding to support different algorithms. This isn't used yet though because we need to settle the data model for the hash mapping.
2) This should be compatible with existing gittuf repos, it depends on the workflows where hashes are re-calculated and verified.
3) This goes back to the data model, certainly things we need to discuss!

I've added tests using Git's experimental implementation of SHA-256 repositories and have confirmed the hashes align with this prototype. :D

adityasaky:
Closing as Git has early support for SHA-1 <-> SHA-256 compatibility. We should reevaluate with support for other hash algorithms later.

PR #107, opened 2023-09-12
Handle panics, direct users to issue tracker

adityasaky:
We have no panics in the codebase, any panic is either a bug or something we ought to handle explicitly (say missing bounds checks, != nil, etc). This PR adds a panic handler for the root CLI command directing users to the issue tracker to report how they encountered this bug.


main.go
```diff
@@ -1,12 +1,22 @@
...
 	"os"
 
 	"github.com/gittuf/gittuf/internal/cmd/root"
 )
 
 func main() {
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Printf("unexpected error: %s\n\n", fmt.Sprint(r))
+			fmt.Printf("Please consider opening an issue on https:/github.com/gittuf/gittuf/issues tagged 'bug' with the steps to reproduce this state. Thanks!\n")
```
wlynch, line 14:
```suggestion
			fmt.Fprintln(os.Stderr, "unexpected error: ", r)
			fmt.Fprintln(os.Stderr, string(debug.Stack()))
			fmt.Fprintln(os.Stderr, "Please consider opening an issue on https:/github.com/gittuf/gittuf/issues tagged 'bug' with the steps to reproduce this state. Thanks!")
```

Prints the stack trace, and sends everything to stderr instead of stdout.

https://go.dev/play/p/nYM8QG_0SJb

wlynch, line 14: If you want to make this even easier, you can create a bug template to auto-label this: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

adityasaky: This is neat! `debug.PrintStack()` prints to stderr inherently so I used that instead. Thanks!

adityasaky: Added a bug report template.

wlynch (approved):

PR #110, opened 2023-09-12
Evaluation mode

adityasaky:
This adds an evaluation mode to gittuf and the first eval-only function, the ability to record RSL entries for non-latest commits. This is required to evaluate gittuf on existing repositories to simulate its behaviour over time.

### DIFF
--- a/internal/cmd/common/common.go
+++ b/internal/cmd/common/common.go
@@ -14,7 +14,10 @@ import (
 const (
 	GPGKeyPrefix = "gpg:"
 	FulcioPrefix = "fulcio:"
+	EvalModeKey  = "GITTUF_EVAL"
 )
+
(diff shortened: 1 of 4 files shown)


internal/cmd/common/common.go
```diff
@@ -77,3 +80,11 @@ func ReadKeyBytes(key string) ([]byte, error) {
...
 	return kb, nil
 }
+
+func EvalMode() bool {
+	return os.Getenv(EvalModeKey) == "1"
+}
+
+func NotInEvalModeError() error {
+	return ErrNotInEvalMode
+}
```
wlynch, line 90: Here we don't need a func wrapper for the error, because users can't actually modify any of the error properties. You can drop this in favor of using `ErrNotInEvalMode` directly./

wlynch (approved):
One small change, otherwise lgtm!

internal/cmd/common/common.go
adityasaky: Good point, I'm not sure what I was doing there hehe.

PR #112, opened 2023-09-13
Clean up GPG key load from bytes

adityasaky:
This PR cleans up how GPG keys are loaded across the codebase. This helps out some inconsistencies with the fingerprint being loaded into the key.KeyID.


internal/signerverifier/gpg/gpg.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/ProtonMail/go-crypto/openpgp"
+	"github.com/gittuf/gittuf/internal/signerverifier"
+	"github.com/gittuf/gittuf/internal/tuf"
+	sslibsv "github.com/secure-systems-lab/go-securesystemslib/signerverifier"
+)
+
+// LoadGPGKeyFromBytes returns a tuf.Key for a GPG / PGP key passed in as
+// armored bytes. The returned tuf.Key uses the primary key's fingerprint as the
+// key ID.
+func LoadGPGKeyFromBytes(contents []byte) (*tuf.Key, error) {
```
wlynch, line 17: If you want to get fancier / more flexible, you could take in a reader and pipe stdout directly into this.

Nice to have though and we're not dealing with large amounts of data here, so up to you if you want to do this. 🤷

wlynch (approved):

internal/signerverifier/gpg/gpg.go
adityasaky: I took a quick pass at this and I think we don't gain much. This is because we need a reader for openpgp to load a keyring but we also need the armored bytes ourselves to set the public field in keyval. So if we pass in a reader, we can pass that directly to openpgp but then we have to re-serialize using a writer for the keyval.

PR #113, opened 2023-09-13
Add support for tags

adityasaky:
* Add general support for tags
* Generalizes signing / verifying operations in gitinterface, reusing code from commit signatures


adityasaky:
The CI pointed me to an issue with how `testGitConfig` was being set and used. I imagine a constructor or so would help but I'm now just setting `config.User.{...}` directly. I've applied this fix for both the tag tests introduced in this PR and the existing commit ones.

wlynch (approved):

adityasaky:
Thanks @wlynch!

PR #114, opened 2023-09-13
Make commit policy finder reusable

adityasaky:
This commit introduces LoadStateForCommit, functionality that was previously embedded in verifyEntry. This enables reuse in the gittuf verify-commit workflow (#101).

### DIFF
--- a/internal/policy/helpers_test.go
+++ b/internal/policy/helpers_test.go
@@ -144,8 +144,9 @@ func createTestStateWithPolicy(t *testing.T) *State {
 	}
 
 	return &State{
-		RootEnvelope:    rootEnv,
-		TargetsEnvelope: targetsEnv,
-		RootPublicKeys:  []*tuf.Key{key},
+		RootEnvelope:        rootEnv,
+		TargetsEnvelope:     targetsEnv,
+		DelegationEnvelopes: map[string]*sslibdsse.Envelope{}, // FIXME: this isn't the best fix. Instead, LoadState* methods should set this to nil when no delegated roles exist.
+		RootPublicKeys:      []*tuf.Key{key},
 	}
 }
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -200,6 +200,29 @@ func LoadStateForEntry(ctx context.Context, repo *git.Repository, e rsl.EntryTyp
 	return state, nil
 }
 
+// GetStateForCommit scans the RSL to identify the first time a commit was seen
+// in the repository. The policy preceding that RSL entry is returned as the
+// State to be used for verifying the commit's signature. If the commit hasn't
+// been seen in the repository previously, no policy state is returned. Also, no
+// error is returned. Identifying the policy in this case is left to the calling
+// workflow.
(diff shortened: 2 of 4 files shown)


internal/policy/helpers_test.go
```diff
@@ -148,8 +148,9 @@ func createTestStateWithPolicy(t *testing.T) *State {
 	}
 
 	return &State{
-		RootEnvelope:    rootEnv,
-		TargetsEnvelope: targetsEnv,
-		RootPublicKeys:  []*tuf.Key{key},
+		RootEnvelope:        rootEnv,
+		TargetsEnvelope:     targetsEnv,
+		DelegationEnvelopes: map[string]*sslibdsse.Envelope{}, // FIXME: this isn't the best fix. Instead, LoadState* methods should set this to nil when no delegated roles exist.
```
wlynch, line 153: Is this something we want to track with an issue, or nah?

internal/policy/policy.go
```diff
@@ -198,6 +198,26 @@ func LoadStateForEntry(ctx context.Context, repo *git.Repository, e rsl.EntryTyp
...
 }
 
+// GetStateForCommit scans the RSL to identify the first time a commit was seen
+// in the repository. The policy preceding that RSL entry is returned as the
+// State to be used for verifying the commit's signature.
+func GetStateForCommit(ctx context.Context, repo *git.Repository, commit *object.Commit) (*State, error) {
+	firstSeenEntry, err := rsl.GetFirstEntryForCommit(repo, commit)
+	if err != nil {
+		if errors.Is(err, rsl.ErrNoRecordOfCommit) {
+			return nil, nil // should err be set here to indicate it's never been seen?
```
wlynch, line 208: Either seems reasonable, but make sure to document this in the godoc.

internal/policy/verify.go
```diff
@@ -201,21 +201,13 @@ func verifyEntry(ctx context.Context, repo *git.Repository, policy *State, entry
...
-				return err
-			}
-			commitPolicy, err = LoadStateForEntry(ctx, repo, commitPolicyEntry)
-			if err != nil {
-				return err
-			}
+			return err
+		}
+		if commitPolicy == nil {
+			commitPolicy = policy
```
wlynch, line 210: This looks like it isn't a change from existing behavior, but this might warrant some more documentation, since it looks like the policy passed in may not be the one that's actually used under certain conditions?

wlynch (approved):
Mostly LGTM! Just some documentation we should clean up.

internal/policy/helpers_test.go
adityasaky: I plan to submit a quick patch fixing these up after this is merged.

internal/policy/policy.go
adityasaky: Will do!

internal/policy/verify.go
adityasaky: Yes, this needs documentation. It's a little related to getting `nil, nil` for when a commit isn't seen before, so the latest policy ought to be used. I'll apply changes to both docs.

adityasaky: Done!

PR #115, opened 2023-09-14
Sync interface

adityasaky:
Adds gittuf clone, push, pull.

This largely supersedes #99.

### DIFF
--- a/internal/gitinterface/sync.go
+++ b/internal/gitinterface/sync.go
@@ -8,7 +8,6 @@ import (
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
-	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/go-git/go-git/v5/storage/memory"
 )
 
@@ -57,6 +56,9 @@ func Push(ctx context.Context, repo *git.Repository, remoteName string, refs []s
 	return PushRefSpec(ctx, repo, remoteName, refSpecs)
 }
 
+// Pull fetches the specified Git refs and gittuf namespaces and applies their
+// changes to the local refs. If one of the specified refs is the current HEAD,
+// the worktree is updated to reflect changes to that ref.
 func Pull(ctx context.Context, repo *git.Repository, remoteName string, refs []string) error {
 	absRefs := make([]string, 0, len(refs))
 	for _, refName := range refs {
@@ -142,7 +144,7 @@ func FetchRefSpec(ctx context.Context, repo *git.Repository, remoteName string, 
 
 	err = remote.FetchContext(ctx, fetchOpts)
 	if err != nil {
-		if !errors.Is(err, transport.ErrEmptyRemoteRepository) && !errors.Is(err, git.NoErrAlreadyUpToDate) {
+		if !errors.Is(err, git.NoErrAlreadyUpToDate) {
 			return err
 		}
 	}
--- a/internal/gitinterface/sync.go
+++ b/internal/gitinterface/sync.go
@@ -57,6 +57,75 @@ func Push(ctx context.Context, repo *git.Repository, remoteName string, refs []s
 	return PushRefSpec(ctx, repo, remoteName, refSpecs)
 }
 
+func Pull(ctx context.Context, repo *git.Repository, remoteName string, refs []string) error {
+	absRefs := make([]string, 0, len(refs))
+	for _, refName := range refs {
+		absRef, err := AbsoluteReference(repo, refName)
+		if err != nil {
+			return err
+		}
+		absRefs = append(absRefs, absRef)
+	}
+
+	// 1. Fetch remote trackers
+	if err := Fetch(ctx, repo, remoteName, refs, true); err != nil {
+		return err
+	}
+
+	// 2. Check and set local references
+	// Also check if head is in the list of refs pulled to update worktree
+	headRef, err := repo.Reference(plumbing.HEAD, false)
+	if err != nil {
+		return err
+	}
+	var newHeadCommit plumbing.Hash
+	for _, refName := range absRefs {
+		ref, err := repo.Reference(plumbing.ReferenceName(refName), true)
+		if err != nil {
+			if !errors.Is(err, plumbing.ErrReferenceNotFound) {
+				return err
+			}
+			if err := repo.Storer.SetReference(plumbing.NewHashReference(plumbing.ReferenceName(refName), plumbing.ZeroHash)); err != nil {
+				return err
+			}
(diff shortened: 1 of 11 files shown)


internal/gitinterface/sync.go
```diff
@@ -72,10 +72,25 @@ func FetchRefSpec(ctx context.Context, repo *git.Repository, remoteName string,
...
+	ref, err := repo.Reference(plumbing.HEAD, false)
+	if err != nil {
+		return err
+	}
+
+	wt, err := repo.Worktree()
+	if err != nil {
+		return err
+	}
+	return wt.Checkout(&git.CheckoutOptions{
```
wlynch, line 91: I don't think we want to do this - fetch typically doesn't modify the local workspace, and you don't want a checkout to the last HEAD to possibly undo uncommitted changes.

internal/repository/sync_test.go
```diff
@@ -0,0 +1,493 @@
...
+		localTmpDir, err := os.MkdirTemp("", "gittuf")
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer os.RemoveAll(localTmpDir) //nolint:errcheck
+
+		if err := os.Chdir(localTmpDir); err != nil {
+			t.Fatal(err)
+		}
+		defer os.Chdir(currentDir) //nolint:errcheck
```
wlynch, line 131: This is repeated a lot. Pull into a helper func.

internal/repository/sync_test.go
```diff
@@ -0,0 +1,493 @@
...
+
+	localPolicyRef, err = repoLocal.r.Reference(policyRefNameTyped, true)
+	if err != nil {
+		t.Fatal(err)
+	}
+	remotePolicyRef, err = repoRemote.Reference(policyRefNameTyped, true)
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.Equal(t, localPolicyRef.Hash(), remotePolicyRef.Hash())
```
wlynch, line 371: Lots of duplication here. Pull into a helper func?

```go
func assertRemoteEqual(t *testing.T, ref string, a Repo, b Repo) {
	t.Helper()
	aRef, err = a.Reference(ref, true)
	if err != nil {
		t.Fatal(err)
	}
	aRef, err = b.Reference(ref, true)
	if err != nil {
		t.Fatal(err)
	}
	assert.Equal(t, aRef.Hash(), bRef.Hash())
}
```

internal/repository/sync_test.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+	if err := remoteRepo.r.Storer.SetReference(plumbing.NewSymbolicReference(plumbing.HEAD, plumbing.ReferenceName(refName))); err != nil {
+		t.Fatal(err)
+	}
+	if err := remoteRepo.r.Storer.SetReference(plumbing.NewHashReference(plumbing.ReferenceName(anotherRefName), commitID)); err != nil {
+		t.Fatal(err)
+	}
+	if err := remoteRepo.RecordRSLEntryForReference(anotherRefName, false); err != nil {
+		t.Fatal(err)
+	}
```
wlynch, line 429: This looks like this is needed for some initial repo setup? Can this also be pulled into a common func?

wlynch (changes requested):

internal/repository/sync.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// Clone wraps a typical git clone invocation, fetching gittuf refs in addition
+// to the standard refs. It performs a verification of the RSL against the
+// specified HEAD after cloning the repository.
+// TODO: resolve how root keys are trusted / bootstrapped.
+func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Repository, error) {
+	if dir == "" {
+		// FIXME: my understanding is backslashes are not used in URLs but I haven't dived into the RFCs to check yet
+		split := strings.Split(strings.TrimSpace(strings.ReplaceAll(remoteURL, "\\", "/")), "/")
```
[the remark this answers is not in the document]
adityasaky: I'll look into this! I wasn't sure about treating this as a URL as it can also be a file path (that's also how we test right now).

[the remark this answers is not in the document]
wlynch: What you might want to do then is feed EscapedPath into filepath.Split, which should also handle the \'s from Windows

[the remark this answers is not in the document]
adityasaky: I took a pass at using this by pulling it into a helper. I'm not sure if we benefit because (unless I'm doing something wrong), EscapedPath gives me nothing for windows file paths. I'm not yet using filepath.Split because we need the EscapedPath to work on Windows first.

internal/gitinterface/sync.go
adityasaky: I wasn't sure what to do here. Note that this is supposed to be the behaviour of pull, so fetch + merge. Without the checkout, the worktree was left unchanged so git assumed there were staged changes (a revert of whatever was just fetched).

internal/repository/sync_test.go
adityasaky: I think I cannot? Because the deferred chdir to currentDir will trigger when exiting the helper instead of the test.

internal/gitinterface/sync.go
adityasaky: I've pushed an attempt at separating Pull from Fetch in gitinterface. The behavior is still incorrect, too aggressive with the worktree.Reset even when the mode is set to MergeReset.

adityasaky:
Okay, commit history aside, I incorporated #108 to gate worktree changes if the remote and local RSLs don't agree. I think this somewhat stops situations where local work is overwritten by remote changes when there is a diverge. But we still broadly need to work on the sync interface, and how it ought to work.

internal/gitinterface/sync.go
adityasaky: @wlynch Would it make sense to dial the scope back here to focusing only on the gittuf namespaces? We have clearer rules on how those branches should work (for example, rsl is always fast forward), and we don't have to mess with worktrees at all. For now, the user'd have to do git push like normal and gittuf sync, and we work out how to clean this up using #4.

adityasaky:
Update: I'll rework this PR to implement `gittuf clone` and sync functions for the gittuf namespaces only. The latter commands are akin to plumbing commands, and we can enable user-focused commands that sync their namespaces + gittuf namespaces later. That'll help reduce the complexity of this PR and we can focus on getting the details of the sync correct.

adityasaky:
This has now been superseded by #134 and #137. I'll leave it open for easy discoverability while we iterate on those PRs.

adityasaky:
#134's been merged, #137 ought to be next. At this point, this should be okay to close.

PR #116, opened 2023-09-15
policy: Verify new policy against previous policy

adityasaky:
So far, we were verifying each policy state internally. We weren't verifying that a new policy's root role is signed by keys trusted in the previous policy's root. This adds that check, and in the process, I think we solve the issue of not verifying RSL entry signatures for policy changes.

### DIFF
--- a/internal/policy/verify.go
+++ b/internal/policy/verify.go
@@ -7,10 +7,14 @@ import (
 
 	"github.com/gittuf/gittuf/internal/gitinterface"
 	"github.com/gittuf/gittuf/internal/rsl"
+	"github.com/gittuf/gittuf/internal/signerverifier"
+	"github.com/gittuf/gittuf/internal/signerverifier/dsse"
 	"github.com/gittuf/gittuf/internal/tuf"
 	"github.com/go-git/go-git/v5"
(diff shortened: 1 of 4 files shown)


JustinCappos (approved):
I'm not familiar with Go, but the goal stated in the docs seems correct.

internal/policy/verify.go
```diff
@@ -129,6 +143,38 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
 	return nil
 }
 
+// VerifyNewPolicyUsingExistingPolicy ensures that when a new policy is
+// encountered, its root role is signed by keys trusted in the prior policy.
+func VerifyNewPolicyUsingExistingPolicy(ctx context.Context, newPolicy, currentPolicy *State) error {
```
wlynch, line 148: Would this make sense to define as a receiver on State?

e.g.

```go
func (s *State) VerifyState(ctx context.Context, new *State) error {
  ...
}
```

wlynch (approved):
One suggestion, but otherwise LGTM! (feel free to submit otherwise)

internal/policy/verify.go
adityasaky: Done!

Issue #102, opened 2023-09-06
Use annotations in RSL queries and policy verification


adityasaky:
RSL annotations are distinct from regular entries in that they apply to other RSL entries. An annotation can also be marked as "skip" to indicate the referred RSL entries have been revoked. While it's possible to record RSL annotations, they're largely not used in the RSL package, nor in the policy package.

This is also a bug because unexpected annotations right now can lead to panics, crashing gittuf.

adityasaky:
## Annotations

We need to establish some expected behaviour around annotations. Specifically, who is authorized to mark a prior entry as skipped? @JustinCappos and I have gone back and forth on this, one option is to require a skip annotation to be signed by a key authorized for the namespaces corresponding to the skip entries.

Also, should an annotation be allowed to mark a policy namespace RSL entry as skipped? My inclination is to say no, policy specific revocations must be handled by a new policy state as it probably means a key rotation / revocation is in order anyway.

Finally, can annotations mark prior annotations as to be skipped? For now, I think we should disallow this until we understand the behaviour around who can and cannot issue a skip annotation.

## Changes to the RSL package

The RSL provides several APIs for retrieving entries. Some of these must be updated to also return annotations that apply to the requested entries. Note that at this point, the RSL package can be "dumb" and return all relevant annotations without checking if the annotations are valid (see above). That should be handled at the policy layer instead as that package also has access to signing keys and so on.

1. GetFirstEntry

An annotation may have been issued for this entry, possibly revoking it. However, the expectation, though not enforced, is that the first entry is for the policy namespace. So, here we also need to consider the point above about whether policy specific entries may even be revoked.

2. GetLatestEntry

No annotations can be returned here as annotations can only occur after the current entry.

3. GetLatestEntryForRef

Annotations that refer to the latest entry for the ref must be returned here. The consumer must decide whether the entry should be used or not based on the annotations. If the consumer does not want to use the entry, the "next" latest entry for the ref can be found using GetLatestEntryForRefBefore with the current entry as the anchor.

4. GetLatestEntryForRefBefore

Annotations that refer to the target entry must be returned here. The consumer must decide whether the entry should be used or not based on the annotations. As before, the consumer can call GetLatestEntryForRefBefore with the current entry as the target to find the "next" latest entry.

5. GetLatestNonGittufEntry

[...]

PR #129, opened 2023-10-06
rsl: Support annotations in RSL APIs

adityasaky:
See: #102

This PR also includes two trivial commits. The first removes earlier TODO statements regarding RSL entry signatures. This is now expected to be handled in the policy package (in fact the verification workflows are already in place for this, but annotations are not yet handled). The second updates the Entry and Annotation methods to use pointer receivers. I'm not sure why it wasn't already doing so.

### DIFF
--- a/internal/rsl/rsl.go
+++ b/internal/rsl/rsl.go
@@ -462,6 +462,111 @@ func GetFirstEntryForCommit(repo *git.Repository, commit *object.Commit) (*Entry
 	}
 }
 
+// GetEntriesInRange returns a list of standard entries between the specified
+// range and a map of annotations that refer to each standard entry in the
+// range. The annotations map is keyed by the ID of the standard entry, with the
+// value being a list of annotations that apply to that standard entry.
+func GetEntriesInRange(repo *git.Repository, firstID, lastID plumbing.Hash) ([]*Entry, map[plumbing.Hash][]*Annotation, error) {
+	return GetEntriesInRangeForRef(repo, firstID, lastID, "")
+}
+
+// GetEntriesInRangeForRef returns a list of standard entries for the ref
+// between the specified range and a map of annotations that refer to each
+// standard entry in the range. The annotations map is keyed by the ID of the
+// standard entry, with the value being a list of annotations that apply to that
+// standard entry.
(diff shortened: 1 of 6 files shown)


adityasaky:
@wlynch this is still marked as a draft because I want to settle behaviours for the RSL APIs. Could you take a look at this and the specified behaviours in https://github.com/gittuf/gittuf/issues/102#issuecomment-1751131831 to see if: a) the behaviours make sense, and b) the implementation here matches those descriptions? Thanks!

adityasaky:
I added two functions to the RSL package to fetch relevant entries in a specified range. This also identifies the relevant annotations for the specified range and returns them. The verification workflow in the policy package is now greatly simplified: it doesn't need to walk the RSL manually, and it doesn't need to handle annotations in any special manner. It can just consume the relevant annotations passed to it.

internal/rsl/rsl.go
```diff
@@ -338,66 +390,175 @@ func GetFirstEntry(repo *git.Repository) (*Entry, error) {
...
+		// Here, all items are relevant until the one corresponding to first is
+		// found
+		switch it := iterator.(type) {
+		case *Entry:
+			if len(refName) != 0 && it.RefName != refName && !strings.HasPrefix(it.RefName, GittufNamespacePrefix) {
+				// If refName is set AND the iterator's ref is not for the
+				// specified ref AND the iterator's ref is not for a gittuf
+				// namespace, break
+				break
+			}
```
wlynch, line 513: It looks like we want to skip these, but IIUC if we encounter this we'll break out and add it to the entry stack anyway on lines 530-533. Is this intentional or do we need to do a similar check below?

wlynch (commented):
Looks good for the most part! One question.

internal/rsl/rsl.go
adityasaky: The `break` here moves execution to line 520. https://go.dev/ref/spec#Break_statements

So, line 530 is encountered only at the end of the for loop to handle the first entry. That said, there _is_ a bug because the same check isn't done there to see if it's a relevant entry.

I'm going to do two things. First, I'm going to invert this check to get rid of the break statement. Second, I'll add the same conditions at 530.

adityasaky: Done!

wlynch (approved):

adityasaky:
Thanks!

PR #131, opened 2023-10-07
Fix RSL terminology

### LINKED ISSUE #130: Settle terminology for RSL entries
The RSL can store two types of objects: an entry and an annotation. Unfortunately, we use the word "entry" in an abstract sense to refer to either of those two types. Some examples:
* the `EntryType` interface
* `GetEntry` which returns the object in the RSL for the specified ID, so this can be either an `Entry` or an `Annotation`
* other APIs

I suggest we use the word "record", possibly as a replacement for the abstract use of the word "entry". I can also see an argument to swap them around, where "entry" is generic and the type that has a ref and target ID becomes a "record". Once this terminology is settled, it should be a fairly easy, if noisy, PR to rename the corresponding APIs. The specification can also be updated at the same time to reflect the change, I foresee a smaller diff there.

adityasaky:
Closes #130

Renames:
Entry -> ~StandardEntry~ ReferenceEntry
Annotation -> AnnotationEntry
EntryType -> Entry

Some APIs like GetEntry... have been renamed depending on whether they return a standard entry or a generic entry.

### DIFF
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -97,7 +97,7 @@ func LoadState(ctx context.Context, repo *git.Repository, rslEntryID plumbing.Ha
 // LoadCurrentState returns the State corresponding to the repository's current
 // active policy.
 func LoadCurrentState(ctx context.Context, repo *git.Repository) (*State, error) {
-	e, _, err := rsl.GetLatestEntryForRef(repo, PolicyRef)
+	e, _, err := rsl.GetLatestReferenceEntryForRef(repo, PolicyRef)
 	if err != nil {
 		return nil, err
 	}
@@ -107,8 +107,8 @@ func LoadCurrentState(ctx context.Context, repo *git.Repository) (*State, error)
 
 // LoadStateForEntry returns the State for a specified RSL entry for the policy
 // namespace.
-func LoadStateForEntry(ctx context.Context, repo *git.Repository, e rsl.EntryType) (*State, error) {
-	entry, ok := e.(*rsl.Entry)
+func LoadStateForEntry(ctx context.Context, repo *git.Repository, e rsl.Entry) (*State, error) {
+	entry, ok := e.(*rsl.ReferenceEntry)
 	if !ok {
 		return nil, ErrNotRSLEntry
 	}
@@ -214,7 +214,7 @@ func LoadStateForEntry(ctx context.Context, repo *git.Repository, e rsl.EntryTyp
 // error is returned. Identifying the policy in this case is left to the calling
 // workflow.
 func GetStateForCommit(ctx context.Context, repo *git.Repository, commit *object.Commit) (*State, error) {
-	firstSeenEntry, _, err := rsl.GetFirstEntryForCommit(repo, commit)
+	firstSeenEntry, _, err := rsl.GetFirstReferenceEntryForCommit(repo, commit)
 	if err != nil {
 		if errors.Is(err, rsl.ErrNoRecordOfCommit) {
 			return nil, nil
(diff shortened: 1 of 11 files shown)


internal/policy/policy.go
```diff
@@ -84,7 +84,7 @@ type State struct {
 // LoadState returns the State of the repository's policy corresponding to the
 // rslEntryID.
 func LoadState(ctx context.Context, repo *git.Repository, rslEntryID plumbing.Hash) (*State, error) {
-	e, err := rsl.GetEntry(repo, rslEntryID)
+	e, err := rsl.GetItem(repo, rslEntryID)
```
JustinCappos, line 87: Are an Entry and an EntryType the same?!?  They both got replaced with "Item"...

internal/policy/verify.go
```diff
@@ -53,7 +53,7 @@ func VerifyRef(ctx context.Context, repo *git.Repository, target string) error {
 	}
 
 	// 2. Find latest entry for target
-	latestEntry, err := rsl.GetLatestEntryForRef(repo, target)
+	latestEntry, _, err := rsl.GetLatestEntryForRef(repo, target)
```
JustinCappos, line 56: These still use Entry...

JustinCappos (commented):
I know this is a draft PR.  I'm quite confused overall about the change and wanted to raise this confusion sooner rather than later.  I'm happy to discuss...

internal/policy/policy.go
adityasaky: No, they're not, this is fixing an earlier mistake. EntryType was supposed to be the generic type, representative of both Entry and Annotation. That's now replaced by Item. GetEntry was named incorrectly, because it can return either an Entry or an Annotation depending on what's requested. So GetEntry becomes GetItem.

internal/policy/verify.go
adityasaky: In this case, this is supposed to return an Entry and never an Annotation. So this is a correct use of "Entry".

internal/policy/policy.go
adityasaky: To summarize, this PR:
* renames EntryType to Item to avoid collision with Entry
* fixes some APIs that used Entry in the name when in fact they return an Item, previously EntryType

adityasaky: Also note that only the last commit pertains to the terminology. I built this PR off the commits from #129, I'll rebase once that one's merged but doing it this way avoids some conflicts. This is also why it's still a draft.

internal/rsl/rsl.go
```diff
@@ -77,34 +80,38 @@ type Entry struct {
 	TargetID plumbing.Hash
 }
 
-// NewEntry returns an Entry object for a normal RSL entry.
-func NewEntry(refName string, targetID plumbing.Hash) *Entry {
-	return &Entry{RefName: refName, TargetID: targetID}
+// NewStandardEntry returns a StandardEntry object for a normal RSL entry.
+func NewStandardEntry(refName string, targetID plumbing.Hash) *StandardEntry {
```
wlynch, line 84: bikeshedding a little bit here since we're renaming anyway: If it's not a StandardEntry, does that make anything else a non-standard Entry? 🤔 

Should this be a `ReferenceEntry` / `ReferenceStateEntry` / something else?

adityasaky: Fair point, I like ReferenceEntry the best. I can submit it fairly quickly.

adityasaky: Done!

wlynch (approved):

Issue #124, opened 2023-09-25
Add SPDX licence identifiers


adityasaky:
As documented in https://github.com/ossf/tac/issues/199, we need to add SPDX license identifiers to our source files. It's also one of the requirements for the next tier of the OpenSSF best practices badge.

PR #132, opened 2023-10-09
Add SPDX License Identifiers to code files

patzielinski:
Add SPDX License Identifiers with Apache 2.0 to files as per #124


adityasaky (approved):
Thanks!

PR #134, opened 2023-10-11
Clone repository with gittuf namespaces and push RSL and policy namespaces to remote

adityasaky:
Based on https://github.com/gittuf/gittuf/pull/115#issuecomment-1756038262.

This PR adds a clone command that clones a repository from the specified remote location with its gittuf namespaces. Additionally, it implements a push command for the policy and RSL namespaces. The remote subcommand for `gittuf policy` and `gittuf trust` are shared in a new package.

PR #135, opened 2023-10-12
Various fixes to gitinterface

adityasaky:
This PR introduces various fixes to gitinterface, some of which were previously seen in #115.

1. AbsoluteReference can now handle HEAD by resolving it to the target and returning that absolute path
2. Previously, there was some code to create the remote tracking ref for some ref, this has been broken out into a reusable component
3. Fetch had a trackRemote option to choose whether the fetch was for the remote tracker or not, this behavior has been updated to always fetch to the specified target ref (so `refs/heads/main:refs/heads/main`) _and_ the corresponding remote tracker
4. Some Fetch* tests missed a symbolic reference for HEAD, this is going to become an issue as we implement other sync interfaces alongside #134 
5. Some tests were using manual tmp dir handling, this has been updated to use `t.TempDir()` based on a prior review

### DIFF
--- a/internal/gitinterface/sync.go
+++ b/internal/gitinterface/sync.go
@@ -84,21 +84,21 @@ func FetchRefSpec(ctx context.Context, repo *git.Repository, remoteName string, 
 // specified remote. For more information on the Git refspec, please consult:
 // https://git-scm.com/book/en/v2/Git-Internals-The-Refspec.
 //
-// The refspecs are constructed to allow non-fast-forward fetches. This function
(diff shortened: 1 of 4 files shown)


adityasaky:
Of these, I think only number 3 is a controversial change as this is slightly different from Git's own behavior. However, we're the only consumers for this and I think we're more likely to want and use the updated behavior. LMK what you think, @wlynch!

wlynch (approved):

PR #137, opened 2023-10-12
Add fetch interfaces for RSL and policy namespaces

adityasaky:
This is complementary to #134 (and currently conflicts with it). This PR also requires #135 to be merged first, as some changes to fetch made there are needed here.

Also, there's some unexpected behavior in how fast-forward only fetches behave. It's possible there's a bug in gitinterface but I think it may also be an assumption around "branches" in go-git. The library assumes branches are in `refs/heads` which means our namespaces don't count as branches. See: https://github.com/go-git/go-git/issues/874


adityasaky:
Okay, this is now consistent with #134. Two blockers:
1. #134 itself, which will likely land first and cause conflicts that should be fairly straightforward to resolve here.
2. https://github.com/go-git/go-git/issues/874 and https://github.com/go-git/go-git/pull/875 in go-git that affect fetch specifically. I think we can't merge this without some fix for that in place, barring a LOT of warnings to the user.

adityasaky:
Another option may be to vendor go-git temporarily with the patch in that PR.

adityasaky:
I've updated this PR with #134's merge conflicts resolved.

wlynch (approved):

adityasaky:
Looks like tests on 1.20 break when go.mod is set to 1.21.x, including just the format of the go directive. I'm dropping support for 1.20 here as a result. Not updating go.mod to 1.21.x also triggers an error prompting a go mod tidy, which moves it back to 1.21.x.

adityasaky:
I'll update the version requirement docs in #144

.github/workflows/lint.yml
```diff
@@ -21,4 +21,4 @@ jobs:
         uses: golangci/golangci-lint-action@3a919529898de77ec3da873e3063ca4b10e7f5cc
         with:
           version: v1.54.2
-          args: --timeout=5m
+          args: --timeout=5m --skip-dirs "internal/third_party/"
```
wlynch, line 24: Create a .golangci.yml file instead. Otherwise there will be different behavior when running in CI vs local.

https://github.com/sigstore/gitsign/blob/main/.golangci.yml#L36-L37

internal/third_party/go-git/README.md
wlynch, line 1: Drop a FORK.md file in this folder that includes what revision / branch we forked from, why we forked, and what the differences are.

wlynch (changes requested):

.github/workflows/lint.yml
adityasaky: Done! https://github.com/gittuf/gittuf/pull/137/commits/3a6ea9c47abfb08d3781bb47d0c54eb17ca865ed

I'll add the additional linters separately as that'll need to be accompanied by unrelated-to-this-PR fixes.

internal/third_party/go-git/README.md
adityasaky: https://github.com/gittuf/gittuf/blob/04321fce171436d3221764a8253356d75243e48c/internal/third_party/go-git/FORK.md

.github/workflows/lint.yml
adityasaky: Replaced that commit with 947ccc82dd60d20a8e9e0f40ce0bfeae8dbdaf17 to fix an issue in the message

wlynch (approved):

PR #141, opened 2023-10-17
common: Updates to test helpers

adityasaky:
The `common` package has some helpers to create signed commits for use in tests. They used to default to the standard GPG key but some upcoming changes (#138) will use different keys to test for failure cases. This puts the invokers in control of the signing key to use.

The second change in this PR adds the ability to create test annotation entries, much like the existing standard entry creation function.

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -4,6 +4,7 @@ package common
 
 import (
 	"bytes"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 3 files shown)


adityasaky:
Either this or #131 will need to be rebased based on the order they land.

adityasaky:
Marking as draft as #131 looks likely to land first.

adityasaky:
Updated to resolve conflicts. This is now ready for review, @wlynch!

wlynch (approved):

PR #142, opened 2023-10-17
rsl: Look for annotations from latest entry

adityasaky:
The change adding support for annotations to RSL functions (#129) didn't correctly handle looking for relevant annotations when an anchor was provided.


adityasaky:
Looks like #131 will land first, so I'm making this a draft.

adityasaky:
Interestingly, in GetNonGittufParentForEntry we were actually testing for this. It didn't error out because assertAnnotationsReferToEntry didn't error out if the entry / annotations passed to it were nil. So by flipping that, the test now fails without the change to GetNonGittufParentForEntry introduced here.

wlynch (approved):

PR #146, opened 2023-10-23
Add version and release workflows

adityasaky:
This PR adds a version subcommand. It also overhauls the Makefile to build with some default flags and set the version information. Finally, it also adds goreleaser to automatically build binaries for our target platforms when we push a release tag to the repository.


adityasaky:
I verified `goreleaser build` generates binaries. We still need to test the workflow works correctly, possibly on a fork?

.github/workflows/release.yml
```diff
@@ -0,0 +1,23 @@
...
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe
+        with:
+          go-version: '1.21.x'
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+        with:
+          fetch-depth: 0 # fetch full history for previous tag information
+      - uses: goreleaser/goreleaser-action@7ec5c2b0c6cdda6e8bbb49444bc797dd33d74dd8
```
wlynch, line 18: Want to add sigstore signing? 👀 

https://github.com/sigstore/gitsign/blob/19fc2c57c09fe51defcb1f91ca70a34e33cc1b09/.goreleaser.yaml#L81-L93
https://github.com/sigstore/gitsign/blob/19fc2c57c09fe51defcb1f91ca70a34e33cc1b09/.github/workflows/release.yml#L33-L35

wlynch (approved):

.github/workflows/release.yml
adityasaky: Done!

wlynch (approved):

PR #149, opened 2023-10-23
rsl: Add functions to find unskipped entries

adityasaky:
Part of #138

This PR adds functions to find the latest unskipped RSL reference entry for the specified ref.

### DIFF
--- a/internal/rsl/rsl.go
+++ b/internal/rsl/rsl.go
@@ -103,6 +103,18 @@ func (e *ReferenceEntry) Commit(repo *git.Repository, sign bool) error {
 	return err
 }
 
+// Skipped returns true if any of the annotations mark the entry as
+// to-be-skipped.
+func (e *ReferenceEntry) SkippedBy(annotations []*AnnotationEntry) bool {
+	for _, annotation := range annotations {
+		if annotation.RefersTo(e.ID) && annotation.Skip {
+			return true
+		}
+	}
+
+	return false
+}
+
 func (e *ReferenceEntry) createCommitMessage() (string, error) {
 	lines := []string{
 		ReferenceEntryHeader,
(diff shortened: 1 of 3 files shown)


internal/rsl/rsl.go
```diff
@@ -411,6 +423,30 @@ func GetLatestReferenceEntryForRefBefore(repo *git.Repository, refName string, a
...
+// GetLatestUnskippedReferenceEntryForRef returns the latest reference entry for
+// the ref that does not have an annotation marking it as to-be-skipped.
+func GetLatestUnskippedReferenceEntryForRef(repo *git.Repository, refName string) (*ReferenceEntry, []*AnnotationEntry, error) {
+	return GetLatestUnskippedReferenceEntryForRefBefore(repo, refName, plumbing.ZeroHash)
+}
+
+// GetLatestUnskippedReferenceEntryForRefBefore returns the first reference
+// entry for the ref before the anchor that does not have an annotation marking
+// it as to-be-skipped.
+func GetLatestUnskippedReferenceEntryForRefBefore(repo *git.Repository, refName string, anchor plumbing.Hash) (*ReferenceEntry, []*AnnotationEntry, error) {
```
wlynch, line 435: It can be ambiguous what before means because you can interpret it as chronologically before if walking from the start of the repo, but this means iteratively before walking the repo starting from the ref.

Maybe clarify with this?

```suggestion
// GetLatestUnskippedReferenceEntryForRefBefore returns the first reference
// entry for the ref that does not have an annotation marking it as to-be-skipped.
// Entries are searched starting from the ref until the anchor is reached.
func GetLatestUnskippedReferenceEntryForRefBefore(repo *git.Repository, refName string, anchor plumbing.Hash) (*ReferenceEntry, []*AnnotationEntry, error) {
```

internal/rsl/rsl.go
```diff
@@ -411,6 +423,30 @@ func GetLatestReferenceEntryForRefBefore(repo *git.Repository, refName string, a
...
+		latestEntry, annotations, err := GetLatestReferenceEntryForRefBefore(repo, refName, anchor)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		if !latestEntry.SkippedBy(annotations) {
+			return latestEntry, annotations, nil
+		}
+
+		anchor = latestEntry.ID
```
wlynch, line 446: I think we should only do this if latest entry != anchor? IIUC if the anchor itself is skipped we'll iterate past it until we find the first non-skipped id.

wlynch (changes requested):

internal/rsl/rsl.go
adityasaky: I think I'm a little confused by the distinction between chronologically and iterative. Could you clarify? Thanks!

internal/rsl/rsl.go
adityasaky: latestEntry can't be anchor because we use `GetLatestReferenceEntryForRefBefore(repo, refName, anchor)` to get it. So we know that's always true, right?

wlynch (approved):

Issue #147, opened 2023-10-23
Ensure that git signing is viable


patzielinski:
When using gittuf, we should make sure that git signing is viable before we touch anything in the repository - we don't want cases where we've initialized something and it turns out that we can't sign anything.

adityasaky:
Thanks for filing this!

To be very precise, we want to ensure we can sign git commits. gittuf invokes this under the hood for RSL entries and the policy namespace. We ought to file early like we do for when a signing key isn't presented for policy metadata (https://github.com/gittuf/gittuf/blob/main/internal/cmd/policy/persistent/persistent.go#L19). For consistency's sake, I propose making this part of the cmd package, by invoking gitinterface.GetSigningCommand (https://github.com/gittuf/gittuf/blob/main/internal/gitinterface/keys.go#L43) in commands' PreRun / PersistentPreRun.

adityasaky:
Here are some docs on the prerun options we have: https://pkg.go.dev/github.com/spf13/cobra#Command

adityasaky:
Closed by #179

PR #152, opened 2023-10-25
Fix trust command issue with signing key not specified

patzielinski:
This PR has gittuf treat the RSL and policy with zero hashes as uninitialized. This fixes the issue where `gittuf trust init` is unable to proceed again after being invoked without a signing key specified in the user's git configuration.

Also see #147

### DIFF
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -56,11 +56,11 @@ var ErrPolicyExists = errors.New("cannot initialize Policy namespace as it exist
 // has a zero hash.
 func InitializeNamespace(repo *git.Repository) error {
 	for _, name := range []string{PolicyRef /*, PolicyStagingRef*/} {
-		if _, err := repo.Reference(plumbing.ReferenceName(name), true); err != nil {
+		if ref, err := repo.Reference(plumbing.ReferenceName(name), true); err != nil {
 			if !errors.Is(err, plumbing.ErrReferenceNotFound) {
 				return err
 			}
-		} else {
+		} else if !ref.Hash().IsZero() {
 			return ErrPolicyExists
 		}
 	}
--- a/internal/policy/policy_test.go
+++ b/internal/policy/policy_test.go
@@ -57,6 +57,15 @@ func TestInitializeNamespace(t *testing.T) {
 			t.Fatal(err)
 		}
 
+		// Check if policy with zero hash is treated as uninitialized
+		err = InitializeNamespace(repo)
+		assert.Nil(t, err)
+
+		if err := repo.Storer.SetReference(plumbing.NewHashReference(PolicyRef, gitinterface.EmptyBlob())); err != nil {
+			t.Fatal(err)
+		}
+
+		// Now with something added, validate that we cannot initialize the policy again
 		err = InitializeNamespace(repo)
 		assert.ErrorIs(t, err, ErrPolicyExists)
 	})
--- a/internal/rsl/rsl_test.go
+++ b/internal/rsl/rsl_test.go
@@ -43,6 +43,15 @@ func TestInitializeNamespace(t *testing.T) {
 			t.Fatal(err)
 		}
 
+		// Check if RSL with zero hash is treated as uninitialized
+		err = InitializeNamespace(repo)
+		assert.Nil(t, err)
+
+		if err := NewReferenceEntry("refs/heads/main", plumbing.ZeroHash).Commit(repo, false); err != nil {
+			t.Fatal(err)
+		}
+
+		// Now with something added, validate that we cannot initialize the RSL again
 		err = InitializeNamespace(repo)
 		assert.ErrorIs(t, err, ErrRSLExists)
 	})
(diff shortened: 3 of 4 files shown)


internal/policy/policy.go
```diff
@@ -56,11 +56,11 @@ var ErrPolicyExists = errors.New("cannot initialize Policy namespace as it exist
 // has a zero hash.
 func InitializeNamespace(repo *git.Repository) error {
 	for _, name := range []string{PolicyRef /*, PolicyStagingRef*/} {
-		if _, err := repo.Reference(plumbing.ReferenceName(name), true); err != nil {
+		if policy, err := repo.Reference(plumbing.ReferenceName(name), true); err != nil {
```
adityasaky, line 59:
```suggestion
		if ref, err := repo.Reference(plumbing.ReferenceName(name), true); err != nil {
```

nit: let's not call it `policy`

internal/policy/policy_test.go
```diff
@@ -57,8 +57,10 @@ func TestInitializeNamespace(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		err = InitializeNamespace(repo)
-		assert.ErrorIs(t, err, ErrPolicyExists)
+		// Check if policy with zero hash is treated as uninitialized
+		if err := InitializeNamespace(repo); err != nil {
+			t.Fatal(err)
```
adityasaky, line 62: You could just replace the assert with `assert.Nil(t, err)`. Also, let's use `Fatal` for test setup operations that fail rather than checks where we want to assert against an expected value.

internal/policy/policy_test.go
```diff
@@ -57,8 +57,10 @@ func TestInitializeNamespace(t *testing.T) {
...
+		}
```
adityasaky, line 63: I see you've done this for the RSL ref, but can you add a subsequent check here where the policy ref is set to something else? Note that it doesn't **need** to be valid policy, the ref just needs to be non zero. Even setting it to `gitinterface.EmptyBlob()` will suffice.

internal/rsl/rsl_test.go
```diff
@@ -43,6 +43,16 @@ func TestInitializeNamespace(t *testing.T) {
 			t.Fatal(err)
 		}
 
+		// Check if RSL with zero hash is treated as uninitialized
+		if err := InitializeNamespace(repo); err != nil {
+			t.Fatal(err)
+		}
+
+		if err := NewReferenceEntry("main", plumbing.ZeroHash).Commit(repo, false); err != nil {
```
adityasaky, line 51:
```suggestion
		if err := NewReferenceEntry("refs/heads/main", plumbing.ZeroHash).Commit(repo, false); err != nil {
```

I'm guilty of using `main` elsewhere in tests but I think it's better to converge on the full path to mimic actual RSL contents. We don't have to go back and fix the other instances right away but if / when we modify those lines, we can take care of it.

internal/rsl/rsl_test.go
```diff
@@ -43,6 +43,16 @@ func TestInitializeNamespace(t *testing.T) {
...
+			t.Error(err)
```
adityasaky, line 52: You can use `assert.Nil` here for consistency.

adityasaky (changes requested):
Thanks for opening this! I think it's quite close, a couple of quick changes and we should be good to merge.

internal/policy/policy.go
```diff
@@ -56,11 +56,11 @@ var ErrPolicyExists = errors.New("cannot initialize Policy namespace as it exist
 // has a zero hash.
 func InitializeNamespace(repo *git.Repository) error {
 	for _, name := range []string{PolicyRef /*, PolicyStagingRef*/} {
-		if _, err := repo.Reference(plumbing.ReferenceName(name), true); err != nil {
+		if ref, err := repo.Reference(plumbing.ReferenceName(name), true); err != nil {
 			if !errors.Is(err, plumbing.ErrReferenceNotFound) {
 				return err
 			}
-		} else {
+		} else if !policy.Hash().IsZero() {
```
adityasaky, line 63: Looks like you missed replacing policy with ref here.

adityasaky (changes requested):

internal/rsl/rsl_test.go
adityasaky: Actually my bad, this should be t.Fatal.

internal/rsl/rsl_test.go
```diff
@@ -43,6 +43,16 @@ func TestInitializeNamespace(t *testing.T) {
 			t.Fatal(err)
 		}
 
+		// Check if RSL with zero hash is treated as uninitialized
+		if err := InitializeNamespace(repo); err != nil {
+			t.Fatal(err)
```
adityasaky, line 48: Here, you'd use an assert because this is the second Initialize. So, it's a check rather than the setup for this test.

adityasaky (approved):
Thanks!

PR #154, opened 2023-10-25
Remove module replace in favor of a package fork.

wlynch:
This removes the sub-go.mod in favor of updating all references to the third_party/go-git package.

This lets users `go install` from outside of the module.


adityasaky (approved):
Thanks!

Issue #145, opened 2023-10-20
Remove vendored go-git module


adityasaky:
#137 vendors go-git with the patch from https://github.com/go-git/go-git/pull/875 in the internal/third_party namespace. Once https://github.com/go-git/go-git/pull/875 is merged and released, we need to remove this copy of go-git.

adityasaky:
https://github.com/go-git/go-git/pull/875 has been merged. We can remove the vendored go-git library with the next release. I'm tracking upstream releases to prepare for it.

adityasaky:
https://github.com/go-git/go-git/releases/tag/v5.10.1 has that PR merged, so we can go ahead and remove go-git now.

PR #159, opened 2023-10-27
Update go-git dependencies and imports

naveensrinivasan:
- Changed receiver when both were mixed https://go.dev/doc/faq#methods_on_values_or_pointers
- Removed import alias which was the same.
- Removed  hash.CryptoType != crypto.SHA256 which will never be true.
- Remove `crypto` and `hash` imports in `repository.go`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 dist/*
+/gittuf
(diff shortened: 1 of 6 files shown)


adityasaky:
@naveensrinivasan Thanks for opening this! Please note that these changes (minus .gitignore) are all to https://github.com/go-git/go-git, which we've temporarily forked (#145, #154). In the next release, we'll be removing it as our upstream patch has been merged by the go-git maintainers. Perhaps you could submit these changes there?

PR #162, opened 2023-10-30
Added fmt target in the Makefile

neilnaveen:
- Added fmt target in the Makefile
- Changed install target prerequisite from test to build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,15 +4,18 @@ GIT_VERSION ?= $(shell git describe --tags --always --dirty)
 
 LDFLAGS=-buildid= -X github.com/gittuf/gittuf/internal/version.gitVersion=$(GIT_VERSION)
 
-.PHONY : build test install
+.PHONY : build test install fmt
 
 default : install
 
 build : test
-	CGO_ENABLED=0 go build -trimpath -ldflags "$(LDFLAGS)" .
+	CGO_ENABLED=0 go build -trimpath -ldflags "$(LDFLAGS)"  -o dist/gittuf .
 
 install : test
 	CGO_ENABLED=0 go install -trimpath -ldflags "$(LDFLAGS)" github.com/gittuf/gittuf
 
 test :
 	go test -v ./...
+
+fmt :
+	go fmt ./...


Makefile
```diff
@@ -11,8 +11,13 @@ default : install
 build : test
 	CGO_ENABLED=0 go build -trimpath -ldflags "$(LDFLAGS)" .
 
-install : test
+install : build
```
adityasaky, line 14: I'm not enough of a gopher so I'll also sub @wlynch in.

My understanding is that `go install` compiles the binary in addition to installing it (with some caching of dependency builds as well). If the install target invokes the build target prior to building it again, are we just duplicating the build?

Makefile
```diff
@@ -11,8 +11,13 @@ default : install
...
 	CGO_ENABLED=0 go build -trimpath -ldflags "$(LDFLAGS)" .
 
-install : test
+install : build 
 	CGO_ENABLED=0 go install -trimpath -ldflags "$(LDFLAGS)" github.com/gittuf/gittuf
 
 test :
 	go test -v ./...
+
+.PHONY: fmt
```
adityasaky, line 20: Can you put this in the .PHONY line at the top of the file with the other targets?

Makefile
```diff
@@ -11,8 +11,13 @@ default : install
 build : test
```
adityasaky, line 11: Somewhat unrelated and this was actually in #159, but .gitignore could be updated to ignore the build binary. Alternatively, the build command here could be updated to place the binary in `dist/` which is already ignored. I lean towards adding a new .gitignore entry so `dist/` is mostly just used with goreleaser.

Makefile
```diff
@@ -11,8 +11,13 @@ default : install
...
+install : build 
 	CGO_ENABLED=0 go install -trimpath -ldflags "$(LDFLAGS)" github.com/gittuf/gittuf
 
 test :
 	go test -v ./...
+
+.PHONY: fmt
+
+fmt:
+	@find . -type d \( -name third_party -prune \) -o -name '*.go' -print | grep -v /third_party/ | xargs -n 1 dirname | sort -u | xargs go fmt
```
adityasaky, line 23: After #145 is closed, this is going to be a lot simpler right? It's unfortunate that `go fmt` doesn't include an option to ignore certain paths / packages. I'm wary of how portable this is on, say, Windows. WDYT of simplifying this to just `go fmt ./...` once the third_party dir is removed? That should be fairly soon.

adityasaky (changes requested):
Hi @neilnaveen, thanks for opening this PR. I think it's useful to have `fmt` in the Makefile. Some thoughts / comments below. :)

Makefile
neilnaveen: Fixed

Makefile
neilnaveen: Fixed

Makefile
neilnaveen: Fixed

Makefile
```diff
@@ -4,15 +4,19 @@ GIT_VERSION ?= $(shell git describe --tags --always --dirty)
...
 LDFLAGS=-buildid= -X github.com/gittuf/gittuf/internal/version.gitVersion=$(GIT_VERSION)
 
-.PHONY : build test install
+.PHONY : build test install fmt
 
 default : install
 
 build : test
-	CGO_ENABLED=0 go build -trimpath -ldflags "$(LDFLAGS)" .
+	mkdir -p ./dist
```
adityasaky, line 12: You don't need mkdir, go build will create the directory if needed. Also note that this may again not be portable, I'm not aware if mkdir is supported as an alias on Windows.

Makefile
```diff
@@ -4,15 +4,19 @@ GIT_VERSION ?= $(shell git describe --tags --always --dirty)
...
+	CGO_ENABLED=0 go build -trimpath -ldflags "$(LDFLAGS)"  -o ./dist
```
adityasaky, line 13:
```suggestion
	CGO_ENABLED=0 go build -trimpath -ldflags "$(LDFLAGS)"  -o dist/gittuf .
```

adityasaky (changes requested):

Makefile
adityasaky: I think dist is fine for now. Thanks for making that change :)

Makefile
```diff
@@ -4,15 +4,19 @@ GIT_VERSION ?= $(shell git describe --tags --always --dirty)
...
 
 default : install
 
 build : test
-	CGO_ENABLED=0 go build -trimpath -ldflags "$(LDFLAGS)" .
+	mkdir -p ./dist
+	CGO_ENABLED=0 go build -trimpath -ldflags "$(LDFLAGS)"  -o dist/gittuf .
 
-install : test
+install : test
```
adityasaky, line 15: nit: I think the only diff on this line now is the trailing space, please remove that.

Makefile
neilnaveen: I took out the mkdir

Makefile
neilnaveen: Fixed

adityasaky (approved):

adityasaky:
Thanks, this looks good to me! Could you squash your commits? And feel free to drop the automatic co-authored by on the intermediate commit for the review suggestion.

neilnaveen:
> Thanks, this looks good to me! Could you squash your commits? And feel free to drop the automatic co-authored by on the intermediate commit for the review suggestion.

I have squashed them. Thanks for the feedback.

adityasaky (approved):
LGTM!

PR #166, opened 2023-11-01
Double Initial Capacity of reSpecs Slice in Fetch

neilnaveen:
- Double initial capacity of reSpecs slice in Fetch for higher efficiency in append operations


adityasaky:
@neilnaveen could you make the linter happy? Good catch otherwise, thanks!

neilnaveen:
> @neilnaveen could you make the linter happy? Good catch otherwise, thanks!

I have linted it.

adityasaky (approved):
Thanks!

PR #168, opened 2023-11-02
(WIP) Preserve verification context to support thresholds

adityasaky:
This PR is a first pass at preserving the specific rule used to verify an RSL entry / commit signature. This is required to support threshold signatures in specific subgroups in the delegation graph.


adityasaky:
cc @JustinCappos

internal/policy/policy.go
```diff
@@ -366,6 +367,63 @@ func (s *State) FindPublicKeysForPath(ctx context.Context, path string) ([]*tuf.
 	}
 }
 
+func (s *State) FindVerifiersForPath(ctx context.Context, path string) ([]*VerificationContext, error) {
```
adityasaky, line 370: This will replace `FindPublicKeysForPath` once we have the verification context behavior nailed down for all scenarios. Right now, for example, we need to see how it affects `verifyTagEntry` where all keys from a policy are used if no specific trusted keys are found.

adityasaky:
This needs some rebasing but it's fundamentally ready for review. I plan to rebase after #205 and #138 are merged once and for all.

wlynch (approved):

PR #164, opened 2023-10-31
(WIP) Add early support for attestations

adityasaky:
This is a prototype of adding support for attestations. It starts with a very simple authorization attestation meant to accompany an RSL reference entry.

PR #170, opened 2023-11-02
attestations: Add plumbing for attestations

adityasaky:
This PR adds the basic plumbing for attestations from #164.

### DIFF
--- a/internal/attestations/attestations.go
+++ b/internal/attestations/attestations.go
@@ -14,10 +14,10 @@ import (
 )
 
 const (
-	Ref                         = "refs/gittuf/attestations"
-	authorizationsTreeEntryName = "authorizations"
-	initialCommitMessage        = "Initial commit"
-	defaultCommitMessage        = "Update attestations"
+	Ref                                  = "refs/gittuf/attestations"
+	referenceAuthorizationsTreeEntryName = "reference-authorizations"
+	initialCommitMessage                 = "Initial commit"
+	defaultCommitMessage                 = "Update attestations"
 )
 
 var ErrAttestationsExist = errors.New("cannot initialize attestations namespace as it exists already")
@@ -45,14 +45,14 @@ func InitializeNamespace(repo *git.Repository) error {
 
 // Attestations tracks all the attestations in a gittuf repository.
 type Attestations struct {
-	// authorizations maps each authorized action to the blob ID of the
+	// referenceAuthorizations maps each authorized action to the blob ID of the
 	// attestation. The key is a path of the form
 	// `<ref-path>/<from-id>-<to-id>`, where `ref-path` is the absolute ref path
 	// such as `refs/heads/main` and `from-id` and `to-id` determine how the ref
 	// in question moved. For example, the key
 	// `refs/heads/main/<commit-A>-<commit-B>` indicates the authorization is
 	// for the action of moving `refs/heads/main` from `commit-A` to `commit-B`.
-	authorizations map[string]plumbing.Hash
+	referenceAuthorizations map[string]plumbing.Hash
 }
 
 // LoadCurrentAttestations inspects the repository's attestations namespace and
@@ -97,7 +97,7 @@ func LoadAttestationsForEntry(repo *git.Repository, entry *rsl.ReferenceEntry) (
 
 	var authorizationsTreeID plumbing.Hash
 	for _, e := range attestationsRootTree.Entries {
-		if e.Name == authorizationsTreeEntryName {
+		if e.Name == referenceAuthorizationsTreeEntryName {
 			authorizationsTreeID = e.Hash
 		}
 	}
@@ -107,9 +107,9 @@ func LoadAttestationsForEntry(repo *git.Repository, entry *rsl.ReferenceEntry) (
 		return nil, err
 	}
 
-	attestations := &Attestations{authorizations: map[string]plumbing.Hash{}}
+	attestations := &Attestations{referenceAuthorizations: map[string]plumbing.Hash{}}
 
-	attestations.authorizations, err = gitinterface.GetAllFilesInTree(authorizationsTree)
+	attestations.referenceAuthorizations, err = gitinterface.GetAllFilesInTree(authorizationsTree)
 	if err != nil {
 		return nil, err
 	}
@@ -129,12 +129,12 @@ func (a *Attestations) Commit(repo *git.Repository, commitMessage string, signCo
 	treeBuilder := gitinterface.NewTreeBuilder(repo)
 
 	// Add authorizations tree
-	authorizationsTreeID, err := treeBuilder.WriteRootTreeFromBlobIDs(a.authorizations)
+	authorizationsTreeID, err := treeBuilder.WriteRootTreeFromBlobIDs(a.referenceAuthorizations)
 	if err != nil {
 		return err
 	}
 	attestationsTreeEntries = append(attestationsTreeEntries, object.TreeEntry{
-		Name: authorizationsTreeEntryName,
+		Name: referenceAuthorizationsTreeEntryName,
 		Mode: filemode.Dir,
 		Hash: authorizationsTreeID,
 	})
--- a/internal/attestations/attestations.go
+++ b/internal/attestations/attestations.go
@@ -109,11 +109,8 @@ func LoadAttestationsForEntry(repo *git.Repository, entry *rsl.ReferenceEntry) (
 
 	attestations := &Attestations{authorizations: map[string]plumbing.Hash{}}
 
-	filesIter := authorizationsTree.Files()
-	if err := filesIter.ForEach(func(f *object.File) error {
-		attestations.authorizations[f.Name] = f.Blob.Hash
-		return nil
-	}); err != nil {
+	attestations.authorizations, err = gitinterface.GetAllFilesInTree(authorizationsTree)
+	if err != nil {
 		return nil, err
 	}
 
--- a/internal/attestations/attestations.go
+++ b/internal/attestations/attestations.go
@@ -89,9 +89,8 @@ func LoadAttestationsForEntry(repo *git.Repository, entry *rsl.ReferenceEntry) (
 
 	if len(attestationsRootTree.Entries) == 0 {
 		// This happens in the initial commit for the attestations namespace,
-		// with no entries in the tree at all
-		// An RSL entry isn't created by default for that entry but if one is,
-		// we don't want to panic.
+		// where there are no entries in the tree yet.
+		// This is expected, and there is nothing more to check so return a zero Attestations state.
 		return &Attestations{}, nil
 	}
 
--- a/internal/attestations/attestations.go
+++ b/internal/attestations/attestations.go
@@ -7,10 +7,10 @@ import (
 
 	"github.com/gittuf/gittuf/internal/gitinterface"
 	"github.com/gittuf/gittuf/internal/rsl"
-	"github.com/gittuf/gittuf/internal/third_party/go-git"
-	"github.com/gittuf/gittuf/internal/third_party/go-git/plumbing"
-	"github.com/gittuf/gittuf/internal/third_party/go-git/plumbing/filemode"
-	"github.com/gittuf/gittuf/internal/third_party/go-git/plumbing/object"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/filemode"
+	"github.com/go-git/go-git/v5/plumbing/object"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


adityasaky:
I'll add more tests for the authorizations specific APIs.

internal/attestations/attestations.go
```diff
@@ -0,0 +1,161 @@
...
+		return err
+	}
+
+	_, err = gitinterface.Commit(repo, treeHash, Ref, initialCommitMessage, false)
+	return err
+}
+
+// Attestations tracks all the attestations in a gittuf repository.
+type Attestations struct {
+	authorizations map[string]plumbing.Hash // refPath/from-to -> blob ID
```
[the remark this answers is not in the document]
wlynch: It would be also useful to add more documentation (either README.md or godoc) on the Attestation format and how refs, commits, ReferenceEntries, and attestations relate to each other.

internal/attestations/authorization.go
```diff
@@ -0,0 +1,174 @@
...
+	predicateStruct, err := structpb.NewStruct(*predicateInterface)
+	if err != nil {
+		return nil, err
+	}
+
+	return &ita.Statement{
+		Type: ita.StatementTypeUri,
+		Subject: []*ita.ResourceDescriptor{
+			{
+				Digest: map[string]string{digestGitCommitKey: toTargetID},
```
wlynch, line 67:
```suggestion
				Digest: map[string]string{"sha1": toTargetID},
```

This should map to the algorithm, no? https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md

internal/attestations/authorization.go
```diff
@@ -0,0 +1,174 @@
...
+type Authorization struct {
+	TargetRef    string `json:"targetRef"`
+	FromTargetID string `json:"fromTargetID"`
+	ToTargetID   string `json:"toTargetID"`
+}
+
+// NewAuthorizationAttestation creates a new authorization for the provided
+// information. The authorization is embedded in an in-toto "statement" and
+// returned with the appropriate "predicate type" set.
+func NewAuthorizationAttestation(targetRef, fromTargetID, toTargetID string) (*ita.Statement, error) {
```
wlynch, line 41: Document what we mean by target id - I assume this means commit sha?

internal/attestations/authorization.go
```diff
@@ -0,0 +1,174 @@
...
+		return ErrAuthorizationNotFound
+	}
+
+	delete(a.authorizations, authPath)
+	return nil
+}
+
+// GetAuthorizationAttestationFor returns the requested authorization
+// attestation (with its signatures).
+func (a *Attestations) GetAuthorizationAttestationFor(repo *git.Repository, refName, fromID, toID string) (*sslibdsse.Envelope, error) {
```
wlynch, line 115: We probably need to tweak how we're storing attestations - I'd expect there to possibly be multiple attestations present for a given commit to/from pair. We can't rely on a single attestation per pair since these may need to be signed by different identities.

internal/attestations/authorization.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return err
+	}
+
+	attestation := &ita.Statement{}
+	if err := json.Unmarshal(payload, attestation); err != nil {
+		return err
+	}
+
+	if attestation.Subject[0].Digest["gitCommit"] != toID {
```
wlynch, line 155: I'd iterate through here, since there could be multiple digests.

internal/attestations/authorization.go
wlynch: I also don't know if we should be keying on ref, since I would expect "give me all attestations for this commit" to be a fairly common workflow - having to scan all refs for possible commits would be costly as the repo grows.

wlynch, line 115: We should add documentation about merge branches and the expected behavior of attestations.

```mermaid
    gitGraph
       commit id: "A"
       branch develop
       commit id:"B"
       commit id:"C"
       checkout main
       merge develop id:"D"
       commit id:"E"
```

e.g. If there's an attestation added to `develop:B->C`, this means that when it's merged to `main` the attestation won't be searchable via `main`, even though the commits are now part of the `main` branch. It makes sense when you realize that Authorization attestations were created w.r.t. a particular ref, but that may not be something that's immediately obvious.

wlynch (changes requested):

internal/attestations/authorization.go
naveensrinivasan, line 155: This could potentially cause a panic in Go. If attestation.Subject is an empty slice.

internal/attestations/authorization.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	if predicate["fromTargetID"] != fromID {
+		return ErrInvalidAuthorization
+	}
+
+	if predicate["targetRef"] != refName {
+		return ErrInvalidAuthorization
+	}
+
```
naveensrinivasan, line 172: The issue here is that the code assumes that the values in the predicate map are strings.

 If they are not, this will cause a panic.

internal/attestations/attestations.go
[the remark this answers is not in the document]
adityasaky: Updated the comment to be more complete of the specific implementation. I'll add authorizations more generally to the design doc, overriding #68.

internal/attestations/attestations.go
```diff
@@ -0,0 +1,161 @@
...
+	attestationsRootTree, err := gitinterface.GetTree(repo, attestationsCommit.TreeHash)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(attestationsRootTree.Entries) == 0 {
+		// This happens in the initial commit for the attestations namespace,
+		// with no entries in the tree at all
+		// An RSL entry isn't created by default for that entry but if one is,
+		// we don't want to panic.
```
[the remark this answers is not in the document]
adityasaky: Do you mean the `Commit()` and other methods on `Attestations`?

internal/attestations/attestations.go
```diff
@@ -0,0 +1,161 @@
...
+
+	attestations := &Attestations{authorizations: map[string]plumbing.Hash{}}
+
+	filesIter := authorizationsTree.Files()
+	if err := filesIter.ForEach(func(f *object.File) error {
+		attestations.authorizations[f.Name] = f.Blob.Hash
+		return nil
+	}); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
adityasaky: You're right, this is expensive. I hadn't peeked enough but this reads the blobs in questions. I was expressly trying to avoid this and just get the filenames by reading the trees rather than the blobs as well, reading the blob in question when the specific authorization is needed. Let me see how I can rework this.

internal/attestations/authorization.go
adityasaky: https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#gitcommit-gittree-gitblob-gittag

The spec supports `git` specific objects as well for an additional hint!

internal/attestations/authorization.go
adityasaky: Yes, most of the time. This is in-line with a recent RSL change where we switched from calling it `commitID` to `targetID` to support pointing to tag objects as well. I'll update the documentation.

internal/attestations/authorization.go
adityasaky: > We probably need to tweak how we're storing attestations - I'd expect there to possibly be multiple attestations present for a given commit to/from pair. We can't rely on a single attestation per pair since these may need to be signed by different identities.

On this one, I disagree. In the `repository` package, the handler for adding an authorization for the same action when someone else signed one loads the attestation that exists and attaches a second signature to the DSSE envelope. That workflow isn't in this PR though. The predicate doesn't change, so that ought to be okay?

> I also don't know if we should be keying on ref, since I would expect "give me all attestations for this commit" to be a fairly common workflow - having to scan all refs for possible commits would be costly as the repo grows.

That's a good point and needs more thinking about. As a start I modeled this on the RSL reference entry as these are essentially additional signatures for an RSL. In the RSL we don't need to explicitly track the from ID, which is the only difference in the recorded information. I'll think about the data model some more.

internal/attestations/authorization.go
adityasaky: Happy to add that here. However, do we foresee allowing users to attach attestations generated outside of gittuf for these workflows? The reason I didn't add this is because I expected gittuf to be the sole producer.

adityasaky: Related: https://github.com/gittuf/gittuf/pull/170#discussion_r1390289843

Right now I expected gittuf to be the sole producer of this attestation, so we could be assured of the fields. But more checks may be in order depending on what we decide in the thread. Thanks for flagging this!

internal/attestations/authorization.go
adityasaky: Also related to whether we can assume we're the creators of this specific attestation (which also has a gittuf.dev predicate type).

internal/attestations/authorization.go
adityasaky: I _think_ that's fine? The authorization is for `develop` to move from B to C. On the `main` branch, the authorization (if one's needed) is to move from `A` to `D`. For a parallel, if we map "authorization" to PR approval, we care that the PR moving `main` from A to D was approved and less so a sub-PR to the feature branch?

adityasaky: Note that this is _strictly_ for authorizations as they work now. Perhaps it needs to be called `reference-state-authorization` or something to make it more clear?

internal/attestations/attestations.go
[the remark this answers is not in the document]
adityasaky: https://github.com/gittuf/gittuf/pull/195

internal/attestations/authorization.go
adityasaky: https://github.com/gittuf/gittuf/pull/195

I've documented this better here, it might help clarify why we may not want this to move with changes being merged / subsumed into other branches.

internal/attestations/attestations.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/gittuf/gittuf/internal/gitinterface"
+	"github.com/gittuf/gittuf/internal/rsl"
+	"github.com/gittuf/gittuf/internal/third_party/go-git"
+	"github.com/gittuf/gittuf/internal/third_party/go-git/plumbing"
+	"github.com/gittuf/gittuf/internal/third_party/go-git/plumbing/filemode"
+	"github.com/gittuf/gittuf/internal/third_party/go-git/plumbing/object"
+)
+
+const (
+	Ref                                  = "refs/gittuf/attestations"
```
wlynch, line 17: Non blocking, but I think we should think about how we can get this to play nice with existing git attestations like [`gitsign attest`](https://github.com/sigstore/gitsign/blob/f915adc8d4b3ccc417c322071d3dbb4bbd3bf1da/internal/commands/attest/README.md) 🤔  (I'll take an AI to dig into this more)

wlynch: In particular - we should figure out sooner rather than later on breaking name changes (e.g. `refs/gittuf/attestations` vs `refs/attestations/commits`) as well as breaking storage formats.

internal/attestations/attestations.go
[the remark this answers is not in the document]
wlynch: I think my original thought was that this was an unexpected case, but re-reading it seems like this might just be the initial state?

maybe rephrase this as:

```go
	if len(attestationsRootTree.Entries) == 0 {
		// This happens in the initial commit for the attestations namespace,
		// where there are no entries in the tree yet.
		// This is expected, and there is nothing more to check so return a zero Attestations state.
		return &Attestations{}, nil
	}
```

(added as suggestion)

internal/attestations/attestations.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(attestationsRootTree.Entries) == 0 {
+		// This happens in the initial commit for the attestations namespace,
+		// with no entries in the tree at all
+		// An RSL entry isn't created by default for that entry but if one is,
+		// we don't want to panic.
+		return &Attestations{}, nil
```
wlynch, line 95:
```suggestion
	if len(attestationsRootTree.Entries) == 0 {
		// This happens in the initial commit for the attestations namespace,
		// where there are no entries in the tree yet.
		// This is expected, and there is nothing more to check so return a zero Attestations state.
		return &Attestations{}, nil
```

wlynch (approved):
One small comment tweak, but otherwise lgtm (feel free to submit)

internal/attestations/attestations.go
adityasaky: I see #237 is open to track this!

PR #172, opened 2023-11-03
Handle Merge Commit in GetFilePathsChangedByCommit

neilnaveen:
- Changed handling of merge commits to check if one of the parent commits matches the current commit

### DIFF
--- a/internal/gitinterface/changes.go
+++ b/internal/gitinterface/changes.go
@@ -3,6 +3,7 @@
 package gitinterface
 
 import (
+	"container/heap"
 	"fmt"
 	"sort"
 
@@ -36,17 +37,58 @@ func GetCommitFilePaths(commit *object.Commit) ([]string, error) {
 
 // GetFilePathsChangedByCommit returns the paths changed by the commit relative
 // to its parent commit. If the commit is a merge commit, i.e., it has more than
-// one parent, no changes are returned.
-//
-// Currently, this function does not verify that the tree for a merge commit
-// matches one of its parents. In a future version, this behavior may change and
-// return an error if a multi-parent commit seems invalid.
+// one parent, check if the commit is the same as at least one of its parents.
+// If there is a matching parent, we return no changes. If there is no matching
+// parent commit, we return the changes between the commit and each of its parents.
 func GetFilePathsChangedByCommit(repo *git.Repository, commit *object.Commit) ([]string, error) {
 	if len(commit.ParentHashes) > 1 {
-		// merge commits are expected not to introduce changes themselves
-		// TODO: should we check that the merge commit's tree matches one of its
-		// parents (usually the last)?
-		return nil, nil
+		// Merge commit: compare with each parent and aggregate changes
+
+		// Create a map to store all changes
+
+		contains := make(map[string]bool)
+
+		// keeping a heap of diffs so that we can pop them in sorted order
+
+		diffs := &diffHeap{}
+		heap.Init(diffs)
+
+		// We are iterating backwards since if there is a matching parent commit,
+		// it is likely to be the last one.
+		for parentHashIndex := len(commit.ParentHashes) - 1; parentHashIndex >= 0; parentHashIndex-- {
+			parentCommit, err := GetCommit(repo, commit.ParentHashes[parentHashIndex])
+			if err != nil {
+				return nil, err
+			}
+			// If the commit tree hash is the same as the parent tree hash, there are no changes
+			if commit.TreeHash == parentCommit.TreeHash {
+				return nil, nil
+			}
+
+			// Get the diff file paths between the commit and its parent
+			diff, err := GetDiffFilePaths(commit, parentCommit)
+			if err != nil {
+				return nil, err
+			}
+
+			for _, change := range diff {
+				// if we have not already added this change
+				if !contains[change] {
+					heap.Push(diffs, change)
+				}
+				// Add changes to the map
+				contains[change] = true
+			}
+		}
+
+		// Convert the heap to a slice
+		changes := make([]string, len(contains))
+
+		for pos := 0; diffs.Len() > 0; pos++ {
+			changes[pos] = heap.Pop(diffs).(string)
+		}
+
+		return changes, nil
 	}
 
 	if len(commit.ParentHashes) == 0 {
@@ -118,4 +160,24 @@ func diff(treeA, treeB *object.Tree) ([]string, error) {
 	})
 
 	return paths, nil
+}
+
+type diffHeap []string
+
+func (h diffHeap) Len() int           { return len(h) }
+func (h diffHeap) Less(i, j int) bool { return h[i] < h[j] }
+func (h diffHeap) Swap(i, j int)      { h[i], h[j] = h[j], h[i] }
+
+func (h *diffHeap) Push(x any) {
+	// Push and Pop use pointer receivers because they modify the slice's length,
+	// not just its contents.
+	*h = append(*h, x.(string))
+}
+
+func (h *diffHeap) Pop() any {
+	old := *h
+	n := len(old)
+	x := old[n-1]
+	*h = old[0 : n-1]
+	return x
 }
--- a/internal/gitinterface/changes_test.go
+++ b/internal/gitinterface/changes_test.go
@@ -597,4 +597,138 @@ func TestGetFilePathsChangedByCommit(t *testing.T) {
 		assert.Nil(t, err)
 		assert.Equal(t, []string{"a"}, diffs)
 	})
+
+	t.Run("merge commit with commit matching parent", func(t *testing.T) {
+		treeA, err := WriteTree(repo, []object.TreeEntry{{Name: "a", Mode: filemode.Regular, Hash: blobIDs[0]}})
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		treeB, err := WriteTree(repo, []object.TreeEntry{{Name: "a", Mode: filemode.Regular, Hash: blobIDs[1]}})
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		cA := CreateCommitObject(testGitConfig, treeA, []plumbing.Hash{plumbing.ZeroHash}, "Test commit", testClock)
+		cAID, err := WriteCommit(repo, cA)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		cB := CreateCommitObject(testGitConfig, treeB, []plumbing.Hash{cAID}, "Test commit", testClock)
+		cBID, err := WriteCommit(repo, cB)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Create a merge commit with two parents
+		cM := CreateCommitObject(testGitConfig, treeB, []plumbing.Hash{cAID, cBID}, "Merge commit", testClock)
+
+		cMID, err := WriteCommit(repo, cM)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		commit, err := GetCommit(repo, cMID)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		diffs, err := GetFilePathsChangedByCommit(repo, commit)
+		assert.Nil(t, err)
+		assert.Nil(t, diffs)
+	})
+
+	t.Run("merge commit with no matching parent", func(t *testing.T) {
+		treeA, err := WriteTree(repo, []object.TreeEntry{{Name: "a", Mode: filemode.Regular, Hash: blobIDs[0]}})
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		treeB, err := WriteTree(repo, []object.TreeEntry{{Name: "b", Mode: filemode.Regular, Hash: blobIDs[1]}})
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		treeC, err := WriteTree(repo, []object.TreeEntry{{Name: "c", Mode: filemode.Regular, Hash: blobIDs[2]}})
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		cA := CreateCommitObject(testGitConfig, treeA, []plumbing.Hash{plumbing.ZeroHash}, "Test commit", testClock)
+		cAID, err := WriteCommit(repo, cA)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		cB := CreateCommitObject(testGitConfig, treeB, []plumbing.Hash{cAID}, "Test commit", testClock)
+		cBID, err := WriteCommit(repo, cB)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Create a merge commit with two parents and a different tree
+		cM := CreateCommitObject(testGitConfig, treeC, []plumbing.Hash{cAID, cBID}, "Merge commit", testClock)
+
+		cMID, err := WriteCommit(repo, cM)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		commit, err := GetCommit(repo, cMID)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		diffs, err := GetFilePathsChangedByCommit(repo, commit)
+		assert.Nil(t, err)
+		assert.Equal(t, []string{"a", "b", "c"}, diffs)
+	})
+
+	t.Run("merge commit with overlapping parent trees", func(t *testing.T) {
+		treeA, err := WriteTree(repo, []object.TreeEntry{{Name: "a", Mode: filemode.Regular, Hash: blobIDs[0]}})
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		treeB, err := WriteTree(repo, []object.TreeEntry{{Name: "a", Mode: filemode.Regular, Hash: blobIDs[1]}})
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		treeC, err := WriteTree(repo, []object.TreeEntry{{Name: "a", Mode: filemode.Regular, Hash: blobIDs[2]}})
+		if err != nil {
+			t.Fatal(err)
+		}
+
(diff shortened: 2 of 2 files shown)


neilnaveen:
Currently the code is erroring out if their are no matches, but what if their was a conflict in the merge commit, and it was resolved manually. This might include some new changes, making it not match any of its parents.

Is this the expected behavior, or would something like this be better?

``` golang
if len(commit.ParentHashes) > 1 {
	// Merge commit: compare with each parent and aggregate changes

	// Create a map to store all changes
	allChanges := make(map[string]bool)

	for _, parentHash := range commit.ParentHashes {
		parentCommit, err := repo.CommitObject(parentHash)
		if err != nil {
			return nil, err
		}

		// Get the diff file paths between the commit and its parent
		diff, err := GetDiffFilePaths(commit, parentCommit)
		if err != nil {
			return nil, err
		}

		// Add changes to the map
		for _, change := range diff {
			allChanges[change] = true
		}
	}

	// Convert the map to a slice
	changes := make([]string, 0, len(allChanges))
	for change := range allChanges {
		changes = append(changes, change)
	}

	// Sort the changes
	sort.Strings(changes)

	return changes, nil
}
```

adityasaky:
I want to double check my thinking (and possible do some merge math) but I think we can return a list of changed file paths for the merge commit as well if it doesn't match any of the parent trees. Trivially, if we diff it against every parent tree, we should be able to construct a list of paths that had a conflict I think? cc @wlynch

adityasaky:
Ah I missed you had an alternative solution here. I think what you have in the message is in-line with my previous suggestion.

internal/gitinterface/changes_test.go
```diff
@@ -597,4 +597,102 @@ func TestGetFilePathsChangedByCommit(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		// Create a merge commit with two parents and a different tree
+		cM := &object.Commit{
+			Author:       cA.Author,
+			Committer:    cA.Committer,
+			Message:      "Merge commit",
+			TreeHash:     treeC, // This tree is different from its parents
+			ParentHashes: []plumbing.Hash{cAID, cBID},
```
adityasaky, line 680: I've wanted to support > 1 parents in the commit APIs. How do you feel about supporting that in a separate PR and then using that here?

neilnaveen: Thanks for the feedback. Just to clarify, in commit.go you want me to create a function that creates a merge commit.

adityasaky: No, for now I think we ought to just _support_ multiple parents in https://github.com/gittuf/gittuf/blob/main/internal/gitinterface/commit.go#L106-L125 and its dependents. The `Commit()` workflow still expects a single parent and that's fine. The `CreateCommitObject` could then be used in this PR with support for multiple parents.

Creating a merge commit is something I think we want to handle at the go-git level. I think some merge workflows aren't supported there just yet. If / when that requirement arises here, we can explore contributing that directly to go-git.

neilnaveen:
> I want to double check my thinking (and possible do some merge math) but I think we can return a list of changed file paths for the merge commit as well if it doesn't match any of the parent trees. Trivially, if we diff it against every parent tree, we should be able to construct a list of paths that had a conflict I think? cc @wlynch

Would this be something on the lines of what you were thinking? We still return `nil, nil` for a case with a match, but without any, we will return all the changes.
``` golang 
if len(commit.ParentHashes) > 1 {
	// Merge commit: compare with each parent and aggregate changes

	// Create a map to store all changes
	allChanges := make(map[string]bool)

	for _, parentHash := range commit.ParentHashes {
		parentCommit, err := repo.CommitObject(parentHash)
		if err != nil {
			return nil, err
		}

		// Get the diff file paths between the commit and its parent
		diff, err := GetDiffFilePaths(commit, parentCommit)
		if err != nil {
			return nil, err
		}
		
		if len(diff) == 0 {
			return nil, nil
		}
		// Add changes to the map
		for _, change := range diff {
			allChanges[change] = true
		}

	}

	// Convert the map to a slice
	changes := make([]string, 0, len(allChanges))
	for change := range allChanges {
		changes = append(changes, change)
	}
	
	// Sort the changes
	sort.Strings(changes)

	return changes, nil
}
```

adityasaky:
Yes, I think that's roughly correct. I have a couple suggestions I'll submit as a review once you add this to the PR.

internal/gitinterface/changes_test.go
neilnaveen: Fixed in #173

internal/gitinterface/changes_test.go
```diff
@@ -597,4 +597,101 @@ func TestGetFilePathsChangedByCommit(t *testing.T) {
 		assert.Nil(t, err)
 		assert.Equal(t, []string{"a"}, diffs)
 	})
+	t.Run("valid merge commit", func(t *testing.T) {
```
adityasaky, line 600: nit: new line before this line

internal/gitinterface/changes_test.go
```diff
@@ -597,4 +597,101 @@ func TestGetFilePathsChangedByCommit(t *testing.T) {
...
+		}
+
+		cB := CreateCommitObject(testGitConfig, treeB, cAID, "Test commit", testClock)
+		cBID, err := WriteCommit(repo, cB)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Create a merge commit with two parents
+		cM := &object.Commit{
```
adityasaky, line 624: This can now use #173's changes!

internal/gitinterface/changes_test.go
```diff
@@ -597,4 +597,101 @@ func TestGetFilePathsChangedByCommit(t *testing.T) {
...
+		commit, err := repo.CommitObject(cMID)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		diffs, err := GetFilePathsChangedByCommit(repo, commit)
+		assert.Nil(t, err)
+		assert.Nil(t, diffs)
+	})
+	t.Run("merge commit with no matching parent", func(t *testing.T) {
```
adityasaky, line 646: nit: new line above.

internal/gitinterface/changes_test.go
```diff
@@ -597,4 +597,101 @@ func TestGetFilePathsChangedByCommit(t *testing.T) {
...
+			TreeHash:     cB.TreeHash,
+			ParentHashes: []plumbing.Hash{cAID, cBID},
+		}
+
+		cMID, err := WriteCommit(repo, cM)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		commit, err := repo.CommitObject(cMID)
```
adityasaky, line 637: `GetCommit`

internal/gitinterface/changes_test.go
```diff
@@ -597,4 +597,101 @@ func TestGetFilePathsChangedByCommit(t *testing.T) {
...
+		}
+
+		cB := CreateCommitObject(testGitConfig, treeB, cAID, "Test commit", testClock)
+		cBID, err := WriteCommit(repo, cB)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Create a merge commit with two parents and a different tree
+		cM := &object.Commit{
```
adityasaky, line 675: #173

internal/gitinterface/changes_test.go
```diff
@@ -597,4 +597,101 @@ func TestGetFilePathsChangedByCommit(t *testing.T) {
...
+			TreeHash:     treeC, // This tree is different from its parents
+			ParentHashes: []plumbing.Hash{cAID, cBID},
+		}
+
+		cMID, err := WriteCommit(repo, cM)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		commit, err := repo.CommitObject(cMID)
```
adityasaky, line 688: `GetCommit`

internal/gitinterface/changes.go
```diff
@@ -3,13 +3,34 @@
...
 import (
+	"container/heap"
 	"fmt"
 	"sort"
 
 	"github.com/gittuf/gittuf/internal/third_party/go-git"
 	"github.com/gittuf/gittuf/internal/third_party/go-git/plumbing/object"
 )
 
+type diffHeap []string
```
adityasaky, line 14: Can you move this to the end of the file? We try to have helpers at the end or in a helpers.go file in the package...

I'll also double check the licensing around this, as I think it's based off the standard heap example in the docs.

adityasaky (changes requested):
Some more comments on this one, but I think it's close!

internal/gitinterface/changes.go
neilnaveen: I moved it to the end of the file

internal/gitinterface/changes_test.go
neilnaveen: Fixed

internal/gitinterface/changes_test.go
neilnaveen: Fixed

internal/gitinterface/changes_test.go
neilnaveen: Fixed

internal/gitinterface/changes_test.go
neilnaveen: Fixed

internal/gitinterface/changes_test.go
neilnaveen: Fixed

internal/gitinterface/changes_test.go
neilnaveen: Fixed

internal/gitinterface/changes.go
```diff
@@ -36,17 +57,55 @@ func GetCommitFilePaths(commit *object.Commit) ([]string, error) {
...
+		heap.Init(diffs)
+
+		for _, parentHash := range commit.ParentHashes {
+			parentCommit, err := repo.CommitObject(parentHash)
+			if err != nil {
+				return nil, err
+			}
+
+			// Get the diff file paths between the commit and its parent
+			diff, err := GetDiffFilePaths(commit, parentCommit)
```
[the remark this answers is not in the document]
neilnaveen: Made the iteration happen in reverse.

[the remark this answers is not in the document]
neilnaveen: Is their a reason for why the matching parent is most likely the last parent?

internal/gitinterface/changes.go
```diff
@@ -36,17 +57,55 @@ func GetCommitFilePaths(commit *object.Commit) ([]string, error) {
...
+
+		contains := make(map[string]bool)
+
+		// keeping a heap of diffs so that we can pop them in sorted order
+
+		diffs := &diffHeap{}
+		heap.Init(diffs)
+
+		for _, parentHash := range commit.ParentHashes {
+			parentCommit, err := repo.CommitObject(parentHash)
```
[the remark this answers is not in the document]
neilnaveen: Fixed

internal/gitinterface/changes.go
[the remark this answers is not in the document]
adityasaky: In a simple merge commit with two parents, the first commit is typically the tip of the base branch with the second parent the tip of the feature branch being merged into the base branch. If there isn't a conflict, the tree of the second parent is therefore the tree of the merge commit as well. Note that this is per my informal model of merges and might differ in implementations.

internal/gitinterface/changes.go
```diff
@@ -36,17 +37,57 @@ func GetCommitFilePaths(commit *object.Commit) ([]string, error) {
 
 // GetFilePathsChangedByCommit returns the paths changed by the commit relative
 // to its parent commit. If the commit is a merge commit, i.e., it has more than
-// one parent, no changes are returned.
-//
-// Currently, this function does not verify that the tree for a merge commit
-// matches one of its parents. In a future version, this behavior may change and
-// return an error if a multi-parent commit seems invalid.
+// one parent, check if the commit is the same as at least one of its parents.
+
```
adityasaky, line 41: This needs to be updated as we now return a list of changes if the tree doesn't match.

internal/gitinterface/changes.go
```diff
@@ -36,17 +37,57 @@ func GetCommitFilePaths(commit *object.Commit) ([]string, error) {
...
+		// Create a map to store all changes
+
+		contains := make(map[string]bool)
+
+		// keeping a heap of diffs so that we can pop them in sorted order
+
+		diffs := &diffHeap{}
+		heap.Init(diffs)
+
+		// We are iterating backwards since if their is a matching parent commit,
```
adityasaky, line 55:
```suggestion
		// We are iterating backwards since if there is a matching parent commit,
```

internal/gitinterface/changes.go
```diff
@@ -36,17 +37,57 @@ func GetCommitFilePaths(commit *object.Commit) ([]string, error) {
...
+		// it is likely to be the last one.
+		for parentHashIndex := len(commit.ParentHashes) - 1; parentHashIndex >= 0; parentHashIndex-- {
+			parentCommit, err := GetCommit(repo, commit.ParentHashes[parentHashIndex])
+			if err != nil {
+				return nil, err
+			}
+
+			// Get the diff file paths between the commit and its parent
```
adityasaky, line 63: Since you have the current commit and parent commit, we can return early here if the tree IDs match. We don't have to fetch both trees etc (which is what happens down that call stack) and then return on `len(diff) == 0`.

internal/gitinterface/changes_test.go
```diff
@@ -597,4 +597,91 @@ func TestGetFilePathsChangedByCommit(t *testing.T) {
 		assert.Nil(t, err)
 		assert.Equal(t, []string{"a"}, diffs)
 	})
+
+	t.Run("valid merge commit", func(t *testing.T) {
```
adityasaky, line 601: nit: "valid" implies others are invalid.

internal/gitinterface/changes_test.go
```diff
@@ -597,4 +597,91 @@ func TestGetFilePathsChangedByCommit(t *testing.T) {
...
+		}
+
+		commit, err := GetCommit(repo, cMID)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		diffs, err := GetFilePathsChangedByCommit(repo, commit)
+		assert.Nil(t, err)
+		assert.Equal(t, []string{"a", "b", "c"}, diffs)
```
adityasaky, line 685: So here, because of how the trees are set, it looks like the merge commit removes `a` and `b`, and adds `c`. This test is fine, but could you add another where the merge commit's tree overlaps with its parents? This is representative of actual merge commit conflict fixes.

internal/gitinterface/changes.go
```diff
@@ -119,3 +160,23 @@ func diff(treeA, treeB *object.Tree) ([]string, error) {
 
 	return paths, nil
 }
+
+type diffHeap []string
```
adityasaky, line 164: This isn't something we want to do in this PR but the other funcs here could probably also use this. We can discuss this separately if you want to work on that as well.

adityasaky (changes requested):
I think these are the final set of changes, after which we should be good to go.

internal/gitinterface/changes.go
neilnaveen: Added. If you think anything else should be added, so it makes more sense, let me know.

internal/gitinterface/changes.go
neilnaveen: Fixed

internal/gitinterface/changes.go
neilnaveen: I have added a check before checking the diffs to check if the commit and parent tree hash are equal.

internal/gitinterface/changes_test.go
neilnaveen: Changed the test name

internal/gitinterface/changes_test.go
neilnaveen: I have created a test that mutates the same file.

internal/gitinterface/changes.go
neilnaveen: I would be open to creating a PR to make this into a general function for use in other parts of the project.

neilnaveen: Would a generalized heap that could take in any type, and a comparator function, be more useful than this heap that only accepts strings and is a min heap, since it would be more generalized and could be used anywhere in the project for any type of datastructure?

internal/gitinterface/changes.go
adityasaky:
```suggestion
```

internal/gitinterface/changes.go
adityasaky: Perhaps an analysis of where / why / what we sort in gittuf is in order first? :smile:

internal/gitinterface/changes_test.go
```diff
@@ -597,4 +597,137 @@ func TestGetFilePathsChangedByCommit(t *testing.T) {
...
+	})
+	t.Run("merge commit with overlapping parent trees", func(t *testing.T) {
```
adityasaky, line 687:
```suggestion

	t.Run("merge commit with overlapping parent trees", func(t *testing.T) {
```

adityasaky (approved):
LGTM minus formatting nits. Could you fix them and squash your commits? Thanks!

neilnaveen:
> Perhaps an analysis of where / why / what we sort in gittuf is in order first? 😄

I don't exactly know the real reason for this, I assumed it was for continuity when calling the function with the same data, the same result would come out. Is their any other reason for the sorting, that I should include?

adityasaky (approved):
Thanks! I'm going to merge this now, but we'll need to keep an eye on merge commit behavior where conflicts occur...

adityasaky:
> I don't exactly know the real reason for this, I assumed it was for continuity when calling the function with the same data, so that the same result would come out. Is their any other reason for the sorting that I should include?

I meant more in the sense of checking if we're always sorting strings, why and where sorting really matters and so on.

PR #174, opened 2023-11-06
Support Loading Key for RSA

neilnaveen:
- Support loading RSA keys into a securesystemlib key for LoadKeyFromBytes.

### DIFF
--- a/internal/tuf/tuf.go
+++ b/internal/tuf/tuf.go
@@ -7,9 +7,13 @@ package tuf
 // however, is inspired by or cloned from the go-tuf implementation.
 
 import (
+	"crypto/rsa"
 	"crypto/sha256"
+	"crypto/x509"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"path"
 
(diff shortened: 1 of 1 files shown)


neilnaveen:
@adityasaky What should the Scheme be for the RSA key?

adityasaky:
Hi @neilnaveen, thanks for trying to address the FIXME. I'm half wondering if we ought to get rid of the function altogether and leave it to go-securesystemslib. Note that once upon a time, tuf.Key wasn't just an alias to the go-securesystemslib key definition, and we needed this method. WDYT?

Also see: https://github.com/secure-systems-lab/go-securesystemslib/blob/main/signerverifier/rsa.go#L97-L141. This is for a signerverifier specifically but general key loading functions can also be supported there I suspect.

adityasaky:
On a separate note, we ought to go back and open issues to track some of the TODO / FIXME items in the codebase. It's a good idea to take stock of them again so that there's space for discussion before a bunch of work and time is put into a solution that may not be applicable anymore. :thinking:

neilnaveen:
> Hi @neilnaveen, thanks for trying to address the FIXME. I'm half wondering if we ought to get rid of the function altogether and leave it to go-securesystemslib. Note that once upon a time, tuf.Key wasn't just an alias to the go-securesystemslib key definition, and we needed this method. WDYT?
> 
> Also see: https://github.com/secure-systems-lab/go-securesystemslib/blob/main/signerverifier/rsa.go#L97-L141. This is for a signerverifier specifically but general key loading functions can also be supported there I suspect.

We should probably use this https://github.com/secure-systems-lab/go-securesystemslib/blob/8c0fa8d1979fda745dae5325708417965522313a/signerverifier/utils.go#L27-L45.

The only problem is that this is not a exported func. I am thinking about creating a PR to make this an exported func, and push a version of these changes I have made in this PR to go-securesystemslib, for that function.

adityasaky:
> We should probably use this https://github.com/secure-systems-lab/go-securesystemslib/blob/8c0fa8d1979fda745dae5325708417965522313a/signerverifier/utils.go#L27-L45.

That doesn't handle PEM encoded RSA keys, we'd have the same issue. That said, I think we generally need to solve this in go-sslib rather than here. in-toto, TUF etc. could then benefit as well. cc @mnm678

neilnaveen:
> > We should probably use this https://github.com/secure-systems-lab/go-securesystemslib/blob/8c0fa8d1979fda745dae5325708417965522313a/signerverifier/utils.go#L27-L45.
> 
> That doesn't handle PEM encoded RSA keys, we'd have the same issue. That said, I think we generally need to solve this in go-sslib rather than here. in-toto, TUF etc. could then benefit as well. cc @mnm678

Do you mean that this should be fixed directly in go-sslib, for [this function](https://github.com/secure-systems-lab/go-securesystemslib/blob/8c0fa8d1979fda745dae5325708417965522313a/signerverifier/utils.go#L27-L45.)  and we should use that function instead of LoadKeyFromBytes. If so, I was meaning that I would push a PR to go-sslib to add this feature there, and then use that function instead of LoadKeyFromBytes here.

neilnaveen:
Just created a PR for this in go-sslib https://github.com/secure-systems-lab/go-securesystemslib/pull/60

adityasaky:
Thanks, I'll review it but I'm at KubeCon so I may be a bit delayed. I don't know that we can use it right away either because we use a separate branch for sigstore compat right now. I'll leave the PR open so we can repurpose it later for this if you're still interested in tracking the issue.

neilnaveen:
@adityasaky Now that this https://github.com/secure-systems-lab/go-securesystemslib/pull/60 has been merged in, should I implement these changes in this PR?

adityasaky:
Not quite yet. 1) We use go-sslib off a custom branch at the moment for sigstore support. 2) We need to cut a release with the RSA key change you submitted.

We need to reconcile both of those to see if we pull in the sigstore WIP changes into the release as well.

adityasaky:
There's a new go-sslib version that we're using for gittuf too, so this should be unblocked.

adityasaky:
I've lost context on the motivation for this one though.

adityasaky:
#257 and some upstream changes to go-securesystemslib means that this is handled beyond just the RSA case. We won't be using the custom format for any key type for much longer except inline in metadata. Thanks for working on this and on go-securesystemslib!

PR #179, opened 2023-11-07
Check that git signing is viable

patzielinski:
This PR addresses #147 where we'd like to ensure that we can sign git commits before we touch anything in the repository. 

With this change, invoking `gittuf rsl record` will cause gittuf to check that a signing key is specified and usable before continuing. For now, only `record.go` is modified - we can move this up to `rsl.go` if we want all rsl commands to inherit this check, as well as add this check to other commands if need be.

### DIFF
--- a/internal/cmd/rsl/record/record.go
+++ b/internal/cmd/rsl/record/record.go
@@ -44,10 +44,11 @@ func (o *options) Run(_ *cobra.Command, args []string) error {
 func New() *cobra.Command {
 	o := &options{}
 	cmd := &cobra.Command{
-		Use:   "record",
-		Short: "Record latest state of a Git reference in the RSL",
-		Args:  cobra.ExactArgs(1),
-		RunE:  o.Run,
+		Use:     "record",
+		Short:   "Record latest state of a Git reference in the RSL",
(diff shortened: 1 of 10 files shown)


internal/cmd/rsl/record/record.go
```diff
@@ -24,6 +25,12 @@ func (o *options) AddFlags(cmd *cobra.Command) {
 	)
 }
 
+func (o *options) CheckSigningKeys(_ *cobra.Command, arg []string) error {
+	_, _, err := gitinterface.GetSigningCommand()
```
adityasaky, line 29: I'd make a generic "is signing viable" function that uses this and put it in cmd/common. Then we can also plug this into the `trust` and `policy` commands that also record RSL entries but don't use this exact command.

adityasaky:
@wlynch WDYT of this? This looks like it achieves the end goal but do you have a more manageable suggestion for when we have more commands?

adityasaky (approved):
I think this is fine and I'm happy to proceed with this. That said, I wonder if there's a more maintainable way of tagging subcommands that require this check...

Also, could you clean up the commit history? Thanks!

adityasaky (approved):

Issue #54, opened 2023-05-30
Enable logging


adityasaky:
gittuf's implementations should include verbose logging for the benefit of users. Actions within the `internal/repository` package in particular should log at various checkpoints what's happening. Further, the user should be able to see these logs via a `--verbose` flag they can invoke from the CLI.

adityasaky:
Relevant: https://pkg.go.dev/log/slog

adityasaky:
I think slog is overkill for gittuf's workflow. But otherwise, I agree with what you're proposing. Go for it!

spectre10:
> I think slog is overkill for gittuf's workflow.

yeah, I though so too. I just made it with the `log` package. Let me know your thoughts. (#180 )

naveensrinivasan:
> I think slog is overkill for gittuf's workflow. But otherwise, I agree with what you're proposing. Go for it!

Why do think so? I would recommend using slog with an interface so that it provides an ideal option to be able to swap.

adityasaky:
When I was first considering what we wanted, I was thinking of it very much in the context of messages presented to users. I expected the log data's  consumers to be exclusively users passing in `--verbose`. Thinking about it some more, it's probably a good idea after all to start off with something structured which is then presented to the user as regular text. The JSON formatted logs might be useful elsewhere such as in a [CI context](https://github.com/gittuf/ci-demo) or any other "hosted" gittuf verification context. Thanks for your input @naveensrinivasan!

naveensrinivasan:
> When I was first considering what we wanted, I was thinking of it very much in the context of messages presented to users. I expected the log data's consumers to be exclusively users passing in `--verbose`. Thinking about it some more, it's probably a good idea after all to start off with something structured which is then presented to the user as regular text. The JSON formatted logs might be useful elsewhere such as in a [CI context](https://github.com/gittuf/ci-demo) or any other "hosted" gittuf verification context. Thanks for your input @naveensrinivasan!

Thanks for the update 👍

Dev-Voldemort:
Hey, I would like to work on adding log statements as I am currently exploring the code-base. Can I DM you on slack for specific queries? @adityasaky

adityasaky:
Sure! Feel free to use the #gittuf channel too.

adityasaky:
Closed by https://github.com/gittuf/gittuf/pull/302.

PR #180, opened 2023-11-07
Add setup for logging with slog. Add 'verbose' flag for logging at debug level.

spectre10:
This PR adds basic setup for logging with slog. And also adds '--verbose' flag for enabling debug level logging. (relevent: #54)

### DIFF
--- a/internal/cmd/root/root.go
+++ b/internal/cmd/root/root.go
@@ -3,6 +3,9 @@
 package root
 
 import (
+	"log/slog"
+	"os"
+
 	"github.com/gittuf/gittuf/internal/cmd/clone"
 	"github.com/gittuf/gittuf/internal/cmd/policy"
 	"github.com/gittuf/gittuf/internal/cmd/rsl"
@@ -14,11 +17,31 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type options struct {
+	verbose bool
+}
+
+func (o *options) AddFlags(cmd *cobra.Command) {
+	cmd.PersistentFlags().BoolVar(&o.verbose, "verbose", false, "enable verbose logging")
+}
+
 func New() *cobra.Command {
+	o := &options{}
 	cmd := &cobra.Command{
 		Use:   "gittuf",
 		Short: "A security layer for Git repositories, powered by TUF",
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			level := slog.LevelInfo
+
+			if o.verbose {
+				level = slog.LevelDebug
+			}
+
+			slog.SetDefault(slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: level})))
+		},
 	}
+
+	o.AddFlags(cmd)
 
 	cmd.AddCommand(clone.New())
 	cmd.AddCommand(trust.New())


internal/logging/logging.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"io"
+	"log"
+	"os"
+)
+
+var (
+	logFile *os.File
+	logFilePath = "info.log"
```
naveensrinivasan, line 11: I would recommend use `slog` package that was introduced in 1.21.

1. Structured Logging: slog allows the code to add key-value pairs to your log messages. This can make your logs easier to parse and analyze. For example, you could modify your setOutput function to include additional context information:

2. Log Levels: slog provides functions for different log levels, such as Info, Debug, Warn, and Error. You could use these functions to log messages at different levels of severity.

3. Custom Handlers: slog allows you to define custom handlers for your logs. You could use this feature to customize how your logs are formatted or where they are sent. For example, you could create a handler that sends critical logs to a separate file or a remote logging service.

4. Performance: slog is designed to be fast and efficient. By using slog, you can improve the performance of your logging package.

``` go
import "log/slog"

type SlogLogger struct {
    logger slog.Logger
}

func NewSlogLogger() *SlogLogger {
    return &SlogLogger{
        logger: slog.Default(),
    }
}

func (l *SlogLogger) Info(message string, keyvals ...any) {
    l.logger.Info(message, keyvals...)
}

func (l *SlogLogger) Error(message string, keyvals ...any) {
    l.logger.Error(message, keyvals...)
}
```
https://go.dev/blog/slog

naveensrinivasan (commented):
👍 

A few suggestions.

internal/logging/logging.go
```diff
@@ -0,0 +1,42 @@
+package logging
```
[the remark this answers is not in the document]
spectre10: That is true! Thank you for the feedback.

internal/logging/logging.go
spectre10: In the beginning, I actually thought it might be overkill for the gittuf workflow (as discussed here #54). However, I am currently working on a slog implementation. I am just in the process of writing a custom handler to format key-val pair logs into human readable output to show on terminal.

naveensrinivasan: Awesome! Thanks for taking the suggestion. Appreciate that!

adityasaky:
Oh one more thing: @spectre10 this isn't compulsory by any means but we also chat about gittuf on the #gittuf channel on the OpenSSF slack workspace: https://join.slack.com/t/openssf/shared_invite/zt-22dfsrz1x-VirRpydvBZCXuTaGSTPWFA. Feel free to drop by if you have any questions and so on.

internal/logutils/logutils.go
```diff
@@ -0,0 +1,18 @@
+package logutils
+
+import (
+	"io"
+	"os"
+
+	"github.com/gittuf/gittuf/internal/logutils/slogger"
+	"github.com/mattn/go-colorable"
```
adityasaky, line 8: Overall, I think I lean towards dropping this. I've not been convinced in the past on colouring log output and I lean towards not introducing the dependency. Is this something others feel strongly about? cc @wlynch as well.

internal/logutils/logger.go
```diff
@@ -0,0 +1,21 @@
+package logutils
```
naveensrinivasan, line 1: > Avoid package names like base, util, or common

https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

internal/logutils/slogger/handler.go
```diff
@@ -0,0 +1,80 @@
...
+	buf = append(buf, r.Message...)
+
+	r.Attrs(func(a slog.Attr) bool {
+		buf = h.appendAttr(buf, a)
+		return true
+	})
+	buf = append(buf, string(h.withAttrs)...)
+	buf = append(buf, "\n"...)
+
+	h.Mu.Lock()
```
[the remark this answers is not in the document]
spectre10: We can use channels, however, [official handler guide](https://github.com/golang/example/blob/master/slog-handler-guide/README.md) uses mutex. And writing to os.Stdout should not be a bottleneck in our case. However, we can use channels if we want to do some processing which can be off-loaded to a different goroutines via channels, as described [here](https://github.com/golang/example/blob/master/slog-handler-guide/README.md#speed). Let me know your thoughts.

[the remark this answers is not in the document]
naveensrinivasan: 👍

internal/logutils/logger.go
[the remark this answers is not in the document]
spectre10: Done!

internal/logutils/logutils.go
spectre10: I have dropped the colors, as everybody agrees on it.

internal/logutils/logger.go
spectre10: Would `internal/logging` name be appropriate?

spectre10:
@adityasaky @naveensrinivasan @patzielinski is adding a source package name from which the log function was called similar to below example, desirable?
example

adityasaky:
I think it's fine to skip that.

wlynch:
Swooping in a little late here, but it wasn't clear to me from the PR description - why implement a new slog handler instead of using the existing TextHandler + top level slog funcs? 🤔 

I don't think we're doing anything *too* special that would require the specialized formatting, and we could simplify this by doing something like:

```go
level := slog.LevelInfo
if verbose {
  level = slog.LevelDebug
}
slog.SetDefault(slog.New(slog.NewTextHandler(os.Stderr, slog.HandlerOptions{Level: level}))
```

We can add a flag for text vs json output later if needed.
For log file teeing, since this will usually be invoked from the command line, can we rely on standard stream redirection for most use cases?

spectre10:
> why implement a new slog handler instead of using the existing TextHandler + top level slog funcs? 🤔

1. The format of the logs shown on Stdout will be key-val pairs only, if we use the existing TextHandler. And that might have unnecessary information like time (although they can be removed with `ReplaceAttr` handler option, we might still want them when logging to the file, as opposed to Stdout.), key names (like level="" msg="") etc. which take up considerable length of the typical log output and are also not easy to glance at. Additionally we might want to write the log source in the log file, but not on Stdout. This type of segregation and formatting for different outputs is not possible with existing handlers. (also useful in CI/CD, as described by @adityasaky [here](https://github.com/gittuf/gittuf/issues/54#issuecomment-1810501107) )
<b> The goal in the end is to have Stdout output similar to the one with `slog.Default()`, but with much more control (see point-2) in our hands.</b>
2. As @naveensrinivasan mentioned earlier, we can extend the functionality of the handler later on. For example, send critical logs to a separate file or a remote logging service. This type of segregation of logs based on levels is also not possible with existing handlers. (can be done via `ReplaceAttr` func, however it is meant to only modify/sanitize the keys/values, and we would also not be able to format them.)
3. Considering the above points, custom handlers seem like the most direct approach. And the complexity of writing custom handlers is also not that high as they have a great official [guide](https://github.com/golang/example/blob/master/slog-handler-guide/README.md) and also a good set of default best practices for it.

As for using top level funcs, we can use them as well. The abstraction is just for convenience as to make it easy to swap the underlying methods/functions and for testability reasons. And also for providing formatted methods, like
`log.Infof("skipped %v issues", numOfIssues)` vs `log.Info(fmt.Sprintf("skipped %v issues", numOfIssues))`.
 Let me know your thoughts! @wlynch

adityasaky:
@spectre10 I understand we want this to be extensible for future cases but I'm also wary of prematurely building up these capabilities in the codebase. I think using slog itself allows for extending when the time comes. For this PR, I suggest we keep it small and stick to the defaults.

spectre10:
@adityasaky I understand. If we were to go according to @wlynch's suggestions, wouldn't it be better to just initialize the default logger in cobra's PreRun or Run funcs? 
```
func New() *cobra.Command {
	cmd := &cobra.Command{
		Use:   "gittuf",
		Short: "A security layer for Git repositories, powered by TUF",
		PersistentPreRun: func(cmd *cobra.Command, args []string) {
			level := slog.LevelInfo
			if b, _ := cmd.Flags().GetBool("verbose"); b {
				level = slog.LevelDebug
			}
			slog.SetDefault(slog.New(slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{Level: level})))
		},
	}
	return cmd
}
```
or should we keep the `internal/logging` package to return a slog.logger instance?

adityasaky:
I'll leave that call to @wlynch. :)

spectre10:
Apologies for the delay (due to academic constraints). Updated the PR according to suggestions. I can rebase the PR after reviews.

internal/cmd/root/root.go
```diff
@@ -14,12 +17,36 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type options struct {
+	Verbose bool
```
adityasaky, line 21:
```suggestion
	verbose bool
```

internal/cmd/root/root.go
```diff
@@ -14,12 +17,36 @@ import (
...
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			level := slog.LevelInfo
+			isVerbose, err := cmd.Flags().GetBool("verbose")
+			if err != nil {
+				return err
+			}
+
+			if isVerbose {
+				level = slog.LevelDebug
+			}
```
adityasaky, line 41:
```suggestion
			}

```

internal/cmd/root/root.go
```diff
@@ -14,12 +17,36 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type options struct {
+	Verbose bool
+}
+
+func (o *options) AddFlags(cmd *cobra.Command) {
+	cmd.PersistentFlags().BoolVar(&o.Verbose, "verbose", false, "Enable verbose logging")
```
adityasaky, line 25:
```suggestion
	cmd.PersistentFlags().BoolVar(&o.Verbose, "verbose", false, "enable verbose logging")
```

internal/cmd/root/root.go
```diff
@@ -14,12 +17,36 @@ import (
...
+}
+
 func New() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "gittuf",
 		Short: "A security layer for Git repositories, powered by TUF",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			level := slog.LevelInfo
+			isVerbose, err := cmd.Flags().GetBool("verbose")
```
adityasaky, line 34: if `o := &options{}` is moved up above the `cmd` definition, we can just use `o.verbose` right?

adityasaky: We shouldn't have an error then and we can just use `PersistentPreRun`?

adityasaky:
```
	o := &options{}
	cmd := &cobra.Command{
		Use:   "gittuf",
		Short: "A security layer for Git repositories, powered by TUF",
		PersistentPreRun: func(cmd *cobra.Command, args []string) {
			level := slog.LevelInfo
			if o.verbose {
				level = slog.LevelDebug
			}

			slog.SetDefault(slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: level})))
		},
	}
	o.AddFlags(cmd)
```

adityasaky (changes requested):
I think this is really close!

internal/cmd/root/root.go
spectre10: Yeah. Updated and rebased.

internal/cmd/root/root.go
```diff
@@ -14,12 +17,32 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type option struct {
```
adityasaky, line 20: apologies, i missed this earlier. For consistency, can you rename this `options`?

spectre10: Done.

adityasaky (approved):
Thanks!

PR #184, opened 2023-11-09
Restrict push workflows to main to prevent double running CI 

wlynch:
This can happen commonly for dependabot and other PRs that are opened from inside the same repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: test
 on:
   push:
+    branches: ['main']
     paths-ignore:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,7 @@
 name: test
```
adityasaky, line 1: Should this be added to lint.yml as well?

adityasaky (approved):

adityasaky:
@wlynch I just applied the same change to the lint workflow as well, merging. Thanks!

Issue #171, opened 2023-11-03
[Bug]: gittuf policy remote commands require a signing key


adityasaky:
### What happened?

The remote subcommands of `gittuf policy` inherit the persistent signing key flag that's required for other policy subcommands. This flag should not be required for push and pull operations as no new metadata signatures are issued.

### Relevant log output

```Shell
$ gittuf policy remote push origin
Error: required flag(s) "signing-key" not set
Usage:
  gittuf policy remote push <remote> [flags]

Flags:
  -h, --help   help for push

Global Flags:
  -k, --signing-key string   signing key to use to sign policy file
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adityasaky:
I was able to get around this by using a signing key (which is never actually loaded in this case), but we ought to roll out a fix for this.

adityasaky:
Closed by #186

PR #186, opened 2023-11-13
fix: signing-key required on policy remote sub-commands.

spectre10:
Addresses #171

### DIFF
--- a/internal/cmd/trust/trust.go
+++ b/internal/cmd/trust/trust.go
@@ -21,8 +21,12 @@ func New() *cobra.Command {
 
 	cmd.AddCommand(i.New(o))
 	cmd.AddCommand(addpolicykey.New(o))
-	cmd.AddCommand(remote.New())
 	cmd.AddCommand(removepolicykey.New(o))
+
+	remoteCmd := remote.New()
+	cmd.AddCommand(remoteCmd)
+	// set signing-key as not required for remote command
+	remoteCmd.InheritedFlags().SetAnnotation("signing-key", cobra.BashCompOneRequiredFlag, []string{"false"}) // nolint:errcheck
 
 	return cmd
 }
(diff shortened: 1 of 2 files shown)


internal/cmd/trustpolicy/remote/remote.go
```diff
@@ -12,6 +12,10 @@ func New() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "remote",
 		Short: "Tools for managing remote policies",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			err := cmd.Flags().SetAnnotation("signing-key", cobra.BashCompOneRequiredFlag, []string{"false"})
```
adityasaky, line 16: Can this instead happen when the child command is added to the parent? https://github.com/gittuf/gittuf/blob/main/internal/cmd/trust/trust.go#L24 and the equivalent for `policy` command, basically. That way, it happens once rather than on prerun every time the command is invoked.

adityasaky (changes requested):
Thanks for taking a stab at this!

internal/cmd/trustpolicy/remote/remote.go
spectre10: Done!

internal/cmd/trust/trust.go
```diff
@@ -21,8 +21,13 @@ func New() *cobra.Command {
 
 	cmd.AddCommand(i.New(o))
 	cmd.AddCommand(addpolicykey.New(o))
-	cmd.AddCommand(remote.New())
 	cmd.AddCommand(removepolicykey.New(o))
 
+	remoteCmd := remote.New()
+	cmd.AddCommand(remoteCmd)
+
```
adityasaky, line 28:
```suggestion
```

nit: consistency with policy.go.

adityasaky (approved):
LGTM, I'll wait for @patzielinski to confirm again as well. Can you squash your commits in the meantime? Thanks!

adityasaky (approved):

PR #187, opened 2023-11-13
Created Generic Heap Datastructure

neilnaveen:
- Changed heap implementation in changes.go to use new generic heap
- Based on this https://github.com/gittuf/gittuf/pull/172#discussion_r1389608478


adityasaky:
General thought (I haven't had a chance to review this yet): all of this is because we're missing a generic set, right? IIRC most sorts we perform are for determinism in testing and aren't actually functionality-critical. I want to think about this some more, let's hang back on implementing things in this direction for now until we decide what's the best way forward.

Issue #128, opened 2023-10-06
Project Governance Documents


hythloda:
Hello gittuf! 

The Linux Foundation has an initial set of documents for you to review that we think are useful for gittuf:

2. [Technical Charter Draft](https://docs.google.com/open?id=1c35OxpKws0jsj0Cf5lyyr-apyhM1bLzOaOS5Yj4YR94)
3. (NOT NEEDED) [Contribution Agreement Draft](https://docs.google.com/open?id=18F6tg3ohyaa1f08W-jijpAwyA1nJevfxJ02g1naD_A8) (NOT NEEDED)
4. (NOT NEEDED)[Series Agreement Draft](https://docs.google.com/open?id=1cEZqAp5-MtUQmNFPWgKidpe5s0uEHtAjZa4r7TbVCCc)(NOT NEEDED)

Please review these materials and let us know if you have any questions. To give you some background on these materials:

## **Technical Project Charter Template**

A technical charter is created for all new projects to define both the project operations and the IP policy.  This document is used to set up an LLC entity, and this becomes the core governance of that LLC.  We have proposed technical oversight for the project falls to a “Technical Steering Committee” made up of the project’s committers. At a later date the TSC is free to evolve how membership on the TSC is determined to accommodate project growth and the evolution of its governance. 

If you like this doc as is let me know and we can adopt it, if you have changes then come to an agreement in together and then we can look at it too. 

## **Project Contribution Agreement**

In order to enable collaboration across organizations, the Linux Foundation and its project hosting entities hold project names, logos and key accounts (e.g., domain names, GitHub accounts, etc.) for the benefit of the project community.  The way we do this is through the execution of the Project Contribution Agreement with the project’s founding organization or, in some cases, individual.  While companies transferring project names and logos to us can continue use those names and logos in accordance with our trademark usage guidelines (please see https://lfprojects.org/policies/trademark-policy/), project names should not be used as part of a commercial product name.  If you are currently using the name of the project as part of a product name, please let us know.

While the Project Contribution Agreement will ask about any trademark registrations, please note your organization may have common law trademark rights in a project name even if a registration has not been filed. Therefore, unless the project name has never been used publicly, we have Project Contribution Agreements signed even if there are no trademark registrations.

## **Series Agreement**

[...]

PR #188, opened 2023-11-14
Update license header

adityasaky:
Per #128, this PR updates the license file's header. The same header is used in the gittuf GitHub Action (https://github.com/gittuf/gittuf-installer/blob/main/LICENSE).


JustinCappos (approved):
LGTM

PR #189, opened 2023-11-14
Created test in log.go for merge commits

neilnaveen:
- Created test with a git tree containing a merge commit
- Added additional edge cases for regular git tree

### DIFF
--- a/internal/gitinterface/log_test.go
+++ b/internal/gitinterface/log_test.go
@@ -36,19 +36,9 @@ func TestGetCommitsBetweenRange(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	treeHashes := make([]plumbing.Hash, 0, 5)
-	for i := 1; i <= 5; i++ {
-		objects := make([]object.TreeEntry, 0, i)
-		for j := 0; j < i; j++ {
-			objects = append(objects, object.TreeEntry{Name: fmt.Sprintf("%d", j+1), Hash: emptyBlobHash})
-		}
-
-		treeHash, err := WriteTree(repo, objects)
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		treeHashes = append(treeHashes, treeHash)
+	treeHashes := createTestTrees(t, repo, emptyBlobHash, 5)
+	if err != nil {
+		t.Fatal(err)
 	}
 
 	commitIDs := []plumbing.Hash{}
@@ -66,15 +56,21 @@ func TestGetCommitsBetweenRange(t *testing.T) {
 		commitIDs = append(commitIDs, ref.Hash())
 	}
 
-	allCommits := make([]*object.Commit, 0, len(commitIDs))
-	for _, commitID := range commitIDs {
-		commit, err := GetCommit(repo, commitID)
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		allCommits = append(allCommits, commit)
+	allCommits, err := GetCommitsFromCommitIDs(commitIDs, repo)
+	if err != nil {
+		t.Fatal(err)
 	}
+	// Git tree structure with their commit trees and their values:
+	//
+	// Commit1 <- Commit2 <- Commit3 <- Commit4 <- Commit5
+	//   |          |           |           |         |
+	// Tree1      Tree2       Tree3       Tree4     Tree5
+	//   |          |           |           |         |
+	// Blob1      Blob1       Blob1       Blob1      Blob1
+	//            Blob2       Blob2       Blob2      Blob2
+	//                        Blob3       Blob3      Blob3
+	//                                    Blob4      Blob4
+	//                                               Blob5
 
 	t.Run("Check range between commits 1 and 5", func(t *testing.T) {
 		commits, err := GetCommitsBetweenRange(repo, commitIDs[4], commitIDs[0])
@@ -102,4 +98,207 @@ func TestGetCommitsBetweenRange(t *testing.T) {
 		})
 		assert.Equal(t, expectedCommits, commits)
 	})
+
+	t.Run("Get commits from invalid range", func(t *testing.T) {
+		commits, err := GetCommitsBetweenRange(repo, plumbing.ZeroHash, plumbing.ZeroHash)
+		assert.ErrorIs(t, err, plumbing.ErrObjectNotFound)
+		assert.Nil(t, commits)
+	})
+	t.Run("Get commits from non-existent commit", func(t *testing.T) {
+		nonExistentHash := EmptyBlob()
+		commits, err := GetCommitsBetweenRange(repo, nonExistentHash, plumbing.ZeroHash)
+		assert.Nil(t, err)
+		assert.Equal(t, commits, []*object.Commit{})
+	})
+}
+
+func TestGetCommitsBetweenRangeForMergeCommits(t *testing.T) {
+	// Creating a tree with merge commits
+	commitIDs := make([]plumbing.Hash, 0, 6)
+	repo, err := git.Init(memory.NewStorage(), memfs.New())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	refName := plumbing.ReferenceName("refs/heads/main")
+
+	if err := repo.Storer.SetReference(plumbing.NewHashReference(refName, plumbing.ZeroHash)); err != nil {
+		t.Fatal(err)
+	}
+
+	ref, err := repo.Reference(refName, true)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	emptyBlobHash, err := WriteBlob(repo, []byte{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	treeHashes := createTestTrees(t, repo, emptyBlobHash, 6)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// creating the first commit
+
+	commit := CreateCommitObject(testGitConfig, treeHashes[0], []plumbing.Hash{ref.Hash()}, fmt.Sprintf("Test commit %v", 1), testClock)
+	commitHash, err := WriteCommit(repo, commit)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	commitIDs = append(commitIDs, commitHash)
+
+	// creating two children from the first commit
+
+	// in the visual, these will be commit 2 and commit 3
+
+	children := createChildrenCommits(t, repo, treeHashes, commitHash, 2)
+
+	commitIDs = append(commitIDs, children...)
+
+	// creating a child for commit 2, which in the visual will be commit 4
+
+	commit = CreateCommitObject(testGitConfig, treeHashes[3], []plumbing.Hash{children[0]}, fmt.Sprintf("Test commit %v", 4), testClock)
+
+	commitHash, err = WriteCommit(repo, commit)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	commitIDs = append(commitIDs, commitHash)
+
+	// creating a merge commit from the two children, which in the visual will be commit 5
+
+	commit = CreateCommitObject(testGitConfig, treeHashes[4], children, fmt.Sprintf("Test commit %v", 5), testClock)
+
+	commitHash, err = WriteCommit(repo, commit)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	commitIDs = append(commitIDs, commitHash)
+
+	// creating a child for commit 3, which in the visual will be commit 6
+
+	commit = CreateCommitObject(testGitConfig, treeHashes[5], []plumbing.Hash{children[1]}, fmt.Sprintf("Test commit %v", 6), testClock)
+
+	commitHash, err = WriteCommit(repo, commit)
+	if err != nil {
+		t.Fatal(err)
+	}
+	commitIDs = append(commitIDs, commitHash)
+
+	// Git tree with merge commit structure without its commit trees and its values:
+	//
+	//  commit 4       commit 5         commit 6
+	//    │              │  │              │
+	//    └─► commit 2 ◄─┘  └─► commit 3 ◄─┘
+	//            │              │
+	//            └─► commit 1 ◄─┘
+
+	allCommits, err := GetCommitsFromCommitIDs(commitIDs, repo)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Run("Test commit 1", func(t *testing.T) {
+		// commit 1 is the first commit, so it should be the only commit returned
+		commits, err := GetCommitsBetweenRange(repo, commitIDs[0], plumbing.ZeroHash)
+		assert.Nil(t, err)
+		expectedCommits := []*object.Commit{allCommits[0]}
+		assert.Equal(t, expectedCommits, commits)
+	})
+
+	t.Run("Test commit 2", func(t *testing.T) {
+		// commit 2 is the first child of commit 1, so only it and commit 1 should be returned
+		commits, err := GetCommitsBetweenRange(repo, commitIDs[1], plumbing.ZeroHash)
+		assert.Nil(t, err)
+		expectedCommits := []*object.Commit{allCommits[1], allCommits[0]}
+		assert.Equal(t, expectedCommits, commits)
+	})
+
+	t.Run("Test commit 3", func(t *testing.T) {
+		// commit 3 is the second child of commit 1, so only it and commit 1 should be returned
+		commits, err := GetCommitsBetweenRange(repo, commitIDs[2], plumbing.ZeroHash)
+		assert.Nil(t, err)
+		expectedCommits := []*object.Commit{allCommits[0], allCommits[2]}
+		assert.Equal(t, expectedCommits, commits)
+	})
+
+	t.Run("Test commit 4", func(t *testing.T) {
+		// commit 4 is the child of commit 2, so only it, commit 2, and commit 2's parent commit 1 should be returned
+		commits, err := GetCommitsBetweenRange(repo, commitIDs[3], plumbing.ZeroHash)
+		assert.Nil(t, err)
+		expectedCommits := []*object.Commit{allCommits[1], allCommits[0], allCommits[3]}
+		assert.Equal(t, expectedCommits, commits)
+	})
+
+	t.Run("Test commit 5, the merge commit", func(t *testing.T) {
+		// commit 5 is the merge commit of commit 2 and commit 3, so it should return commit 5, commit 2, commit 3, and commit 1 (the parent of commit 2 and commit 3)
+		commits, err := GetCommitsBetweenRange(repo, commitIDs[4], plumbing.ZeroHash)
+		assert.Nil(t, err)
+		expectedCommits := []*object.Commit{allCommits[4], allCommits[1], allCommits[0], allCommits[2]}
+		assert.Equal(t, expectedCommits, commits)
+	})
+
+	t.Run("Test commit 6", func(t *testing.T) {
+		// commit 6 is the child of commit 3, so it should return commit 6, commit 3, and commit 1 (the parent of commit 3)
+		commits, err := GetCommitsBetweenRange(repo, commitIDs[5], plumbing.ZeroHash)
+		assert.Nil(t, err)
+		expectedCommits := []*object.Commit{allCommits[0], allCommits[5], allCommits[2]}
+		assert.Equal(t, expectedCommits, commits)
+	})
+}
+
+func GetCommitsFromCommitIDs(commitIDs []plumbing.Hash, repo *git.Repository) ([]*object.Commit, error) {
+	allCommits := make([]*object.Commit, 0, len(commitIDs))
+	for _, commitID := range commitIDs {
+		commit, err := GetCommit(repo, commitID)
+		if err != nil {
+			return nil, err
+		}
+
+		allCommits = append(allCommits, commit)
+	}
+
+	return allCommits, nil
+}
+
+func createTestTrees(t *testing.T, repo *git.Repository, emptyBlobHash plumbing.Hash, num int) []plumbing.Hash {
+	t.Helper()
+	treeHashes := make([]plumbing.Hash, 0, num)
+	for i := 1; i <= num; i++ {
+		objects := make([]object.TreeEntry, 0, i)
+		for j := 0; j < i; j++ {
+			objects = append(objects, object.TreeEntry{Name: fmt.Sprintf("%d", j+1), Hash: emptyBlobHash})
+		}
+
+		treeHash, err := WriteTree(repo, objects)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		treeHashes = append(treeHashes, treeHash)
+	}
+	return treeHashes
+}
+
+func createChildrenCommits(t *testing.T, repo *git.Repository, treeHashes []plumbing.Hash, parentHash plumbing.Hash, numChildren int) []plumbing.Hash {
+	t.Helper()
+
+	children := make([]plumbing.Hash, 0, numChildren)
+
+	for i := 1; i <= numChildren; i++ {
+		commit := CreateCommitObject(testGitConfig, treeHashes[i], []plumbing.Hash{parentHash}, fmt.Sprintf("Test commit %v", i+1), testClock)
+
+		commitHash, err := WriteCommit(repo, commit)
+		if err != nil {
+			t.Fatal(err)
+		}
+		children = append(children, commitHash)
+	}
+	return children
 }


adityasaky:
cc @patzielinski

internal/gitinterface/log_test.go
```diff
@@ -66,15 +66,21 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		allCommits = append(allCommits, commit)
+	allCommits, err := getCommitsFromCommitIds(commitIDs, repo)
+	if err != nil {
+		t.Fatal(err)
 	}
+	// Git tree structure with their commit trees and their values :
```
adityasaky, line 73:
```suggestion
	// Git tree structure with their commit trees and their values:
```

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
 		})
 		assert.Equal(t, expectedCommits, commits)
 	})
+	t.Run("Get commits from invalid range", func(t *testing.T) {
```
adityasaky, line 111:
```suggestion

	t.Run("Get commits from invalid range", func(t *testing.T) {
```

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+		commits, err := GetCommitsBetweenRange(repo, plumbing.ZeroHash, plumbing.ZeroHash)
+		assert.NotNil(t, err)
+		assert.Nil(t, commits)
+	})
```
adityasaky, line 115:
```suggestion

	})
```

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
 		})
 		assert.Equal(t, expectedCommits, commits)
 	})
+	t.Run("Get commits from invalid range", func(t *testing.T) {
+		commits, err := GetCommitsBetweenRange(repo, plumbing.ZeroHash, plumbing.ZeroHash)
+		assert.NotNil(t, err)
```
adityasaky, line 113: Can you `assert.ErrorIs` here?

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+		assert.Nil(t, commits)
+	})
+	t.Run("Get commits from non-existent commit", func(t *testing.T) {
+		nonExistentHash := plumbing.NewHash("1234567890123456789012345678901234567890")
```
adityasaky, line 117: Let's use `EmptyBlob` or something here.

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+		commits, err := GetCommitsBetweenRange(repo, nonExistentHash, plumbing.ZeroHash)
+		assert.NotNil(t, err)
```
adityasaky, line 119: `assert.ErrorIs`?

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	emptyBlobHash, err := WriteBlob(repo, []byte{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	treeHashes := make([]plumbing.Hash, 0, 6)
```
adityasaky, line 149: Can this be moved into a helper that accepts `n` for number of trees? Then that can be reused by the previous test too.

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+	})
+	t.Run("Get commits from non-existent commit", func(t *testing.T) {
+		nonExistentHash := plumbing.NewHash("1234567890123456789012345678901234567890")
+		commits, err := GetCommitsBetweenRange(repo, nonExistentHash, plumbing.ZeroHash)
+		assert.NotNil(t, err)
+		assert.Nil(t, commits)
+	})
+}
+
+func TestGetCommitsBetweenRangeForMergeCommits(t *testing.T) {
```
adityasaky, line 124: I see why this is a separate test but a part of me wonders about keeping them organized by making this a `t.Run` in the previous one. :thinking:

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+	//    │              │  │              │
+	//    └─► commit 2 ◄─┘  └─► commit 3 ◄─┘
+	// 		      │              │
+	// 		 	  └─► commit 1 ◄─┘
+
+	allCommits, err := getCommitsFromCommitIds(commitIDs, repo)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Run("Get all commits from merge commit 5", func(t *testing.T) {
```
adityasaky, line 231:
```suggestion

	t.Run("Get all commits from merge commit 5", func(t *testing.T) {
```

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+	}
+
+	commitIDs = append(commitIDs, ref.Hash())
+
+	// Git tree with merge commit structure without its commit trees and its values :
+	//
+	//  commit 4       commit 5         commit 6
+	//    │              │  │              │
+	//    └─► commit 2 ◄─┘  └─► commit 3 ◄─┘
+	// 		      │              │
```
adityasaky, line 224: I think the spacing on this diagram is off for commit 1 to commit 2

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+		childRefName := plumbing.ReferenceName(fmt.Sprintf("refs/heads/child%d", i))
+		childRef, err := applyCommitAndRefreshRef(repo, commit, plumbing.NewHashReference(childRefName, plumbing.ZeroHash))
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		commitIDs = append(commitIDs, childRef.Hash())
+		children = append(children, childRef.Hash())
+	}
+	// creating a merge commit from the two children
```
adityasaky, line 189:
```suggestion

	// creating a merge commit from the two children
```

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+	commit = CreateCommitObject(testGitConfig, treeHashes[5], []plumbing.Hash{children[1]}, fmt.Sprintf("Test commit %v", 6), testClock)
+
+	ref, err = applyCommitAndRefreshRef(repo, commit, ref)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	commitIDs = append(commitIDs, ref.Hash())
+
+	// Git tree with merge commit structure without its commit trees and its values :
```
adityasaky, line 219:
```suggestion
	// Git tree with merge commit structure without its commit trees and its values:
```

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+		assert.Nil(t, err)
+		expectedCommits := []*object.Commit{allCommits[3], allCommits[2], allCommits[1], allCommits[0]}
+		sort.Slice(expectedCommits, func(i, j int) bool {
+			return expectedCommits[i].ID().String() < expectedCommits[j].ID().String()
+		})
+
+		assert.Equal(t, expectedCommits, commits)
+		assert.Nil(t, err)
+	})
+	t.Run("Get all commits from merge commit 6", func(t *testing.T) {
```
adityasaky, line 242:
```suggestion

	t.Run("Get all commits from merge commit 6", func(t *testing.T) {
```

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+		commits, err := GetCommitsBetweenRange(repo, commitIDs[3], plumbing.ZeroHash)
```
adityasaky, line 243: This invocation looks identical to the one for merge commit 5

adityasaky: Also, is 6 a merge commit? In the diagram its parent seems to be commit 3 only?

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+		expectedCommits := []*object.Commit{allCommits[3], allCommits[2], allCommits[1], allCommits[0]}
+		sort.Slice(expectedCommits, func(i, j int) bool {
+			return expectedCommits[i].ID().String() < expectedCommits[j].ID().String()
+		})
+		assert.Equal(t, expectedCommits, commits)
+		assert.Nil(t, err)
+	})
+}
+
+func getCommitsFromCommitIds(commitIDs []plumbing.Hash, repo *git.Repository) ([]*object.Commit, error) {
```
adityasaky, line 254: I suspect this could be a wrapper around `GetCommit` and actually exposed as part of gitinterface.

internal/gitinterface/log_test.go
```diff
@@ -102,4 +108,171 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		allCommits = append(allCommits, commit)
+	}
+	return allCommits, nil
+}
+
+func applyCommitAndRefreshRef(repo *git.Repository, commit *object.Commit, ref *plumbing.Reference) (*plumbing.Reference, error) {
```
adityasaky, line 267: 1. Do we need to return a new plumbing.Reference? Since we're passing in the pointer of the ref, just updating it should do?
2. If this is only used in tests, let's pass in `t`, register this as a helper `t.Helper()`, and abort in case of errors `t.Fatal`. If 1 holds and with this change, we don't need to return anything at all here.

adityasaky: Also, it'd be neat to track other places this helper could be used. I think there are certainly other tests that could reuse this.

adityasaky: They don't need to be updated in this PR though.

adityasaky (changes requested):
Thanks @neilnaveen! Some questions around the new tests and some formatting suggestions.

internal/gitinterface/log_test.go
adityasaky: It's off in https://github.com/gittuf/gittuf/blob/94b5e2a50f068974d37bdd0a6e0ed8c03206e158/internal/gitinterface/log_test.go too though :\

The issue is there's a mix of `\t` and spaces. We can't control tab width, so I suggest using spaces throughout for these diagrams.

internal/gitinterface/log_test.go
neilnaveen: Yes, this is the wrong test name, I was thinking about something on the line of, "Get commits from leaf commit 6 to root commit 1"

internal/gitinterface/log_test.go
neilnaveen: Fixed

internal/gitinterface/log_test.go
neilnaveen: I have made this an exported func

internal/gitinterface/log_test.go
neilnaveen: Fixed, changed the test name, and fixed the test to actually reflect test name. Merge commit 5's test has also been fixed, before it was not actually testing the merge commit, but something else.

internal/gitinterface/log_test.go
neilnaveen: fixed

internal/gitinterface/log_test.go
neilnaveen: fixed

internal/gitinterface/log_test.go
neilnaveen: fixed

internal/gitinterface/log_test.go
neilnaveen: fixed

internal/gitinterface/log_test.go
neilnaveen: fixed

internal/gitinterface/log_test.go
neilnaveen: fixed

internal/gitinterface/log_test.go
neilnaveen: fixed

internal/gitinterface/log_test.go
neilnaveen: fixed

internal/gitinterface/log_test.go
neilnaveen: fixed

internal/gitinterface/log_test.go
neilnaveen: fixed

internal/gitinterface/log_test.go
neilnaveen: fixed

internal/gitinterface/log_test.go
```diff
@@ -102,4 +98,182 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+		assert.Nil(t, err)
+		expectedCommits := []*object.Commit{allCommits[5], allCommits[2], allCommits[0]}
+		sort.Slice(expectedCommits, func(i, j int) bool {
+			return expectedCommits[i].ID().String() < expectedCommits[j].ID().String()
+		})
+		assert.Equal(t, expectedCommits, commits)
+	})
+}
+
+func GetCommitsFromCommitIds(commitIDs []plumbing.Hash, repo *git.Repository) ([]*object.Commit, error) {
```
adityasaky, line 217:
```suggestion
func GetCommitsFromCommitIDs(commitIDs []plumbing.Hash, repo *git.Repository) ([]*object.Commit, error) {
```

Couple other thoughts with this one:
1. For consistency, please reorder the arguments so we pass in the repo first. This aligns with other APIs in this package.
2. We shouldn't have an exported function in a `_test` file. This should go in `commit.go`.
3. We really ought to be testing this function.

internal/gitinterface/log_test.go
adityasaky: I think I've come around on this one given the difference in setup.

internal/gitinterface/log_test.go
```diff
@@ -102,4 +98,182 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+	applyCommitAndRefreshRef(t, repo, commit, ref)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	commitIDs = append(commitIDs, ref.Hash())
+
+	// creating two children from the first commit
+
+	childrenCommitIds, children := createChildrenCommits(t, repo, ref, treeHashes, 2)
```
adityasaky, line 157:
```suggestion
	childrenCommitIDs, children := createChildrenCommits(t, repo, ref, treeHashes, 2)
```

internal/gitinterface/log_test.go
```diff
@@ -102,4 +98,182 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+	treeHashes, err := createTreeHashes(repo, emptyBlobHash, 6)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// creating the first commit
+
+	commit := CreateCommitObject(testGitConfig, treeHashes[0], []plumbing.Hash{ref.Hash()}, fmt.Sprintf("Test commit %v", 1), testClock)
+
+	applyCommitAndRefreshRef(t, repo, commit, ref)
```
adityasaky, line 148: Strictly speaking, we only want the commits, no? We don't need to actually apply them to a ref? The commits can just live in the object store.

internal/gitinterface/log_test.go
adityasaky: Also, for consistency and readability, please update `children` (the second return value) to indicate what it is. I'm confused because it seems to be identical to `childrenCommitIDs`?

internal/gitinterface/log_test.go
```diff
@@ -102,4 +98,182 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+	// creating a merge commit from the two children
+
+	commit = CreateCommitObject(testGitConfig, treeHashes[4], children, fmt.Sprintf("Test commit %v", 5), testClock)
+
+	applyCommitAndRefreshRef(t, repo, commit, ref)
+	commitIDs = append(commitIDs, ref.Hash())
+
+	// creating children for each of the two children
+
+	commit = CreateCommitObject(testGitConfig, treeHashes[3], []plumbing.Hash{children[0]}, fmt.Sprintf("Test commit %v", 4), testClock)
```
adityasaky, line 170: Could you indicate in a comment what each index in `children` maps to in the diagram? It's hard to keep track atm.

internal/gitinterface/log_test.go
```diff
@@ -102,4 +98,182 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+	//            │              │
+	//            └─► commit 1 ◄─┘
+
+	allCommits, err := GetCommitsFromCommitIds(commitIDs, repo)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Run("Get all commits from merge commit 5", func(t *testing.T) {
+		commits, err := GetCommitsBetweenRange(repo, commitIDs[4], plumbing.ZeroHash)
```
adityasaky, line 196: Here as well, a comment to indicate what each index maps to would be helpful to know what commits are returned in the diagram. That should give us a good idea of whether the behavior aligns with our expectations.

internal/gitinterface/log_test.go
```diff
@@ -102,4 +98,182 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+	allCommits := make([]*object.Commit, 0, len(commitIDs))
+	for _, commitID := range commitIDs {
+		commit, err := GetCommit(repo, commitID)
+		if err != nil {
+			return nil, err
+		}
+
+		allCommits = append(allCommits, commit)
+	}
+	return allCommits, nil
```
adityasaky, line 227:
```suggestion

	return allCommits, nil
```

internal/gitinterface/log_test.go
```diff
@@ -102,4 +98,182 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+}
+
+func applyCommitAndRefreshRef(t *testing.T, repo *git.Repository, commit *object.Commit, ref *plumbing.Reference) {
+	t.Helper()
+	if _, err := ApplyCommit(repo, commit, ref); err != nil {
```
adityasaky, line 232:
```suggestion

	if _, err := ApplyCommit(repo, commit, ref); err != nil {
```

internal/gitinterface/log_test.go
```diff
@@ -102,4 +98,182 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+
+	for i := 1; i <= numChildren; i++ {
+		commit := CreateCommitObject(testGitConfig, treeHashes[i], []plumbing.Hash{ref.Hash()}, fmt.Sprintf("Test commit %v", i+1), testClock)
+		childRefName := plumbing.ReferenceName(fmt.Sprintf("refs/heads/child%d", i))
+		childRef := plumbing.NewHashReference(childRefName, commit.Hash)
+
+		applyCommitAndRefreshRef(t, repo, commit, childRef)
+
+		commitIDs = append(commitIDs, childRef.Hash())
+		children = append(children, childRef.Hash())
```
adityasaky, line 276: Are you returning this twice to separate handling children at during invocation? I wonder if some things could be moved around there instead?

adityasaky (changes requested):
Some more comments, I think this is getting a lot closer!

internal/gitinterface/log_test.go
neilnaveen: I did this, and just to clear things up, I made a test for every single commit, so people can understand what connects to what, and how it works.

internal/gitinterface/log_test.go
neilnaveen: Done

internal/gitinterface/log_test.go
neilnaveen: Fixed

internal/gitinterface/log_test.go
neilnaveen: Fixed

internal/gitinterface/log_test.go
neilnaveen: I had refactored the creation of children into the function, and I overlooked that, the result is now just one slice of hashes that is be appended to the commitID's slice in the main test.

internal/gitinterface/log_test.go
neilnaveen: Fixed

internal/gitinterface/log_test.go
neilnaveen: would you like me to move this function to commit.go and in a later PR test it, or should I just keep it as an non-exported function for now?

internal/gitinterface/log_test.go
```diff
@@ -36,19 +36,9 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
-			objects = append(objects, object.TreeEntry{Name: fmt.Sprintf("%d", j+1), Hash: emptyBlobHash})
-		}
-
-		treeHash, err := WriteTree(repo, objects)
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		treeHashes = append(treeHashes, treeHash)
+	treeHashes, err := createTreeHashes(repo, emptyBlobHash, 5)
```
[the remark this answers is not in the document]
neilnaveen: Fixed

internal/gitinterface/log_test.go
adityasaky: bump

neilnaveen:
The test are failing, because when I am calling GetCommitsBetweenRange from a given commit. I am getting it with the second parameter as a zero hash, I am not getting the first commit in the commit tree, which is expected. Am I creating the first commit incorrectly? @adityasaky

internal/gitinterface/log_test.go
```diff
@@ -102,4 +99,218 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	commitIDs = append(commitIDs, commitHash)
+
+	// creating two children from the first commit
+
+	// in the visual, these will be commit 2 and commit 3
+
+	children := createChildrenCommits(t, repo, ref, treeHashes, 2)
```
adityasaky, line 159: `ref` is never updated here, so commit 2 and 3 are created without a parent, right?

adityasaky: I suggest ripping `ref` out entirely, we only need the commits themselves.

neilnaveen: I forgot about that, will change

neilnaveen: Fixed

internal/gitinterface/log_test.go
```diff
@@ -102,4 +99,218 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+	return treeHashes
+}
+
+func createChildrenCommits(t *testing.T, repo *git.Repository, treeHashes []plumbing.Hash, numChildren int) []plumbing.Hash {
+	t.Helper()
+
+	children := make([]plumbing.Hash, 0, numChildren)
+
+	for i := 1; i <= numChildren; i++ {
+		commit := CreateCommitObject(testGitConfig, treeHashes[i], []plumbing.Hash{}, fmt.Sprintf("Test commit %v", i+1), testClock)
```
adityasaky, line 307: this also ends up with no parents, you want to pass in the parent commit ID(s) here.

neilnaveen: @adityasaky I have done this, is this the change you mean?

internal/gitinterface/log_test.go
```diff
@@ -102,4 +99,218 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+			t.Errorf("Expected and actual commits do not match. Expected: %v, Got: %v", expectedCommits, commits)
+		}
+	})
+
+	t.Run("Test commit 3", func(t *testing.T) {
+		commits, err := GetCommitsBetweenRange(repo, commitIDs[2], plumbing.ZeroHash)
+		assert.Nil(t, err)
+		// commit 3 is the second child of commit 1, so only it and commit 1 should be returned
+		expectedCommits := []*object.Commit{allCommits[0], allCommits[2]}
+		if !reflect.DeepEqual(expectedCommits, commits) {
```
adityasaky, line 232: any specific reason for not using `assert.Equal` here? I think this was originally using assert?

internal/gitinterface/log_test.go
```diff
@@ -102,4 +99,218 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+	//    │              │  │              │
+	//    └─► commit 2 ◄─┘  └─► commit 3 ◄─┘
+	//            │              │
+	//            └─► commit 1 ◄─┘
+
+	allCommits, err := GetCommitsFromCommitIDs(commitIDs, repo)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Run("Test commit 1", func(t *testing.T) {
```
adityasaky, line 207:
```suggestion

	t.Run("Test commit 1", func(t *testing.T) {
```

as a rule of thumb, please insert a newline before every new "block" like a new `t.Run` or whatever.

adityasaky (commented):
final comments I think!

internal/gitinterface/log_test.go
```diff
@@ -102,4 +99,218 @@ func TestGetCommitsBetweenRange(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	commitIDs = append(commitIDs, commitHash)
+
+	// creating two children from the first commit
+
+	// in the visual, these will be commit 2 and commit 3
+
+	children := createChildrenCommits(t, repo, treeHashes, commit.Hash, 2)
```
adityasaky, line 159: Hmm, I'm guessing here but I think you want to use `commitHash` here instead of `.Hash`. This is because `commit` is not updated after `WriteCommit` so it has no notion of its hash in the storer.

internal/gitinterface/log_test.go
neilnaveen: Assert errors out for this test, at least in my editor because it cannot compare expected commits with the returned commits, since the commits have unexported fields, which assert cannot process.

internal/gitinterface/log_test.go
neilnaveen: Fixed

internal/gitinterface/log_test.go
neilnaveen: Done

neilnaveen:
> The test are failing, because when I am calling GetCommitsBetweenRange from a given commit. I am getting it with the second parameter as a zero hash, I am not getting the first commit in the commit tree, which is expected. Am I creating the first commit incorrectly? @adityasaky

I am guessing that they were just flakey test, because now they are running properly 😆

internal/gitinterface/log_test.go
neilnaveen: Taken out

internal/gitinterface/log_test.go
adityasaky: Can you paste the error here? We don't need to assert private fields, and we use assert with similar inputs in other tests.

neilnaveen: @adityasaky I just ran it again, with assert, and it worked. I am guessing my test were flakey. Changed the reflect.DeepEqual to assert.Equal

adityasaky:
Can you squash your commits? :)

neilnaveen:
> Can you squash your commits? :)

Squashed

PR #194, opened 2023-11-21
policy: Account for non Git signing keys in verify

adityasaky:
This PR adds a check for cases where a key is technically trusted to sign for some namespace but it can only be used with gittuf policies, not Git objects. We have this check elsewhere.

Long term, this check can be pulled out because the signing method aligns with SSH keys. So I suggest we add this check in now and strip them out when we have support for Git <-> SSH keys.

### DIFF
--- a/internal/policy/verify.go
+++ b/internal/policy/verify.go
@@ -213,8 +213,8 @@ func VerifyCommit(ctx context.Context, repo *git.Repository, ids ...string) map[
 			}
 
 			if errors.Is(err, gitinterface.ErrUnknownSigningMethod) {
-				// We encounter this for key types that can be used for metadata
-				// but not Git objects
+				// We encounter this for key types that can be used for gittuf
+				// policy metadata but not Git objects
 				continue
 			}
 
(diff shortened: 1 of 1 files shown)


internal/policy/verify.go
```diff
@@ -375,6 +375,11 @@ func verifyEntry(ctx context.Context, repo *git.Repository, policy *State, entry
 			gitNamespaceVerified = true
 			break
 		}
+		if errors.Is(err, gitinterface.ErrUnknownSigningMethod) {
+			// We encounter this for key types that can be used for metadata
```
wlynch, line 379: It's not immediate clear what metadata means here - can we add more details?

wlynch (approved):
LGTM, but we should add a bit more detail (feel free to submit after updating).

internal/policy/verify.go
adityasaky: fixed!

adityasaky:
Thanks @wlynch!

Also, cc @bradcush.

PR #199, opened 2023-11-21
Added a termination check for delegation search

neilnaveen:
- Added a termination check for delgation search
- Changed the type of search from Breadth-First Search to Depth-First Search because if a terminating delegation is found, in BFS we would already have considered its siblings, but in DFS if we found a terminating delegation, we would have not already searched its other siblings which is expected behavior.

### DIFF
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -674,15 +674,15 @@ func (s *State) findDelegationEntry(roleName string) (tuf.Delegation, error) {
 		delegationTargetsMetadata[name] = targetsMetadata
 	}
 
-	delegationsStack := topLevelTargetsMetadata.Delegations.Roles
+	delegationsToCheck := topLevelTargetsMetadata.Delegations.Roles
 
 	for {
-		if len(delegationsStack) == 0 {
+		if len(delegationsToCheck) == 0 {
 			return tuf.Delegation{}, ErrDelegationNotFound
 		}
 
-		delegation := delegationsStack[0]
-		delegationsStack = delegationsStack[1:]
+		delegation := delegationsToCheck[0]
+		delegationsToCheck = delegationsToCheck[1:]
 
 		if delegation.Name == roleName {
 			return delegation, nil
@@ -690,9 +690,9 @@ func (s *State) findDelegationEntry(roleName string) (tuf.Delegation, error) {
 
 		if s.HasTargetsRole(delegation.Name) {
 			if delegation.Terminating {
-				delegationsStack = delegationTargetsMetadata[delegation.Name].Delegations.Roles
+				delegationsToCheck = delegationTargetsMetadata[delegation.Name].Delegations.Roles
 			}
-			delegationsStack = append(delegationTargetsMetadata[delegation.Name].Delegations.Roles, delegationsStack...)
+			delegationsToCheck = append(delegationTargetsMetadata[delegation.Name].Delegations.Roles[:len(delegationTargetsMetadata[delegation.Name].Delegations.Roles)-1], delegationsToCheck...)
 		}
 	}
 }
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -3,10 +3,9 @@
 package policy
 
 import (
-	"errors"
-
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


internal/policy/policy.go
```diff
@@ -675,23 +674,25 @@ func (s *State) findDelegationEntry(roleName string) (tuf.Delegation, error) {
...
 			return delegation, nil
 		}
 
 		if s.HasTargetsRole(delegation.Name) {
-			// TODO: clarifying terminating / reachability
-			delegationsQueue = append(delegationsQueue, delegationTargetsMetadata[delegation.Name].Delegations.Roles...)
+			if delegation.Terminating {
+				delegationsStack = delegationTargetsMetadata[delegation.Name].Delegations.Roles
+			}
+			delegationsStack = append(delegationTargetsMetadata[delegation.Name].Delegations.Roles, delegationsStack...)
```
adityasaky, line 695: I think we'll want to do all-but-the-last-one here. See: https://github.com/gittuf/gittuf/blob/main/internal/policy/policy.go#L359-L362

internal/policy/policy.go
```diff
@@ -675,23 +674,25 @@ func (s *State) findDelegationEntry(roleName string) (tuf.Delegation, error) {
 		delegationTargetsMetadata[name] = targetsMetadata
 	}
 
-	delegationsQueue := topLevelTargetsMetadata.Delegations.Roles
+	delegationsStack := topLevelTargetsMetadata.Delegations.Roles
```
adityasaky, line 677: It's not really a stack. It's not a typical queue either but it's closer IMO. Note that we may add to the front but we still add multiple entries that are processed in order.

adityasaky: Life may be simpler if we avoid `queue`/`stack` in the name hehe. (I'm aware I started using that in the name in the first place)

adityasaky (changes requested):

adityasaky:
I haven't had a chance to look but I'm nervous about whether we have sufficient test coverage here, directly or indirectly, to be sure we aren't breaking something else. Can you take a look? :smile:

neilnaveen:
> I haven't had a chance to look but I'm nervous about whether we have sufficient test coverage here, directly or indirectly, to be sure we aren't breaking something else. Can you take a look? 😄

I can write some test for this function, since currently this function has no coverage.

internal/policy/policy.go
neilnaveen: I changed it to delegationsToCheck

adityasaky:
>I can write some test for this function, since currently this function has no coverage.

That'd be great, thanks!

internal/policy/policy.go
```diff
@@ -675,23 +674,25 @@ func (s *State) findDelegationEntry(roleName string) (tuf.Delegation, error) {
...
 
 		if delegation.Name == roleName {
 			return delegation, nil
 		}
 
 		if s.HasTargetsRole(delegation.Name) {
-			// TODO: clarifying terminating / reachability
-			delegationsQueue = append(delegationsQueue, delegationTargetsMetadata[delegation.Name].Delegations.Roles...)
+			if delegation.Terminating {
+				delegationsToCheck = delegationTargetsMetadata[delegation.Name].Delegations.Roles
```
adityasaky, line 693: FYI: we've been discussing terminating delegation behavior in #168 and #206.

neilnaveen: @adityasaky Should I first create a PR just for testing this function, and later on, I can change this PR according to how termination is going to be handled?

neilnaveen: @adityasaky I have added test in #209

adityasaky: Having looked at that, I think it can land regardless and we can work out terminating delegation in this context separately.

adityasaky:
This is basically good to go right? It needs a rebase to fix conflicts? I see tests were added to `findDelegationEntry`.

adityasaky:
Actually, I was thinking about this some more, and I think we have a separate issue (previously seen in python-tuf in https://github.com/theupdateframework/python-tuf/issues/660). In the short term, I will close this because we shouldn't terminate since we don't know the namespace we're finding the delegation for. In the long term, I think we want to remove `state.Verify()` altogether and check for validity only while walking the delegation graph. We already do this to an extent in using the new verifier in #168.

PR #201, opened 2023-11-22
Check that the tip of a ref matches latest corresponding RSL entry

adityasaky:
This updates the repository package to check that the current state of the tested ref matches the latest RSL entry.

### DIFF
--- a/internal/policy/verify.go
+++ b/internal/policy/verify.go
@@ -43,44 +43,46 @@ var (
 )
 
 // VerifyRef verifies the signature on the latest RSL entry for the target ref
-// using the latest policy.
-func VerifyRef(ctx context.Context, repo *git.Repository, target string) error {
+// using the latest policy. The expected Git ID for the ref in the latest RSL
+// entry is returned if the policy verification is successful.
+func VerifyRef(ctx context.Context, repo *git.Repository, target string) (plumbing.Hash, error) {
 	// 1. Get latest policy entry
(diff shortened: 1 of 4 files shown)


internal/policy/verify.go
```diff
@@ -40,43 +40,43 @@ var ErrUnauthorizedSignature = errors.New("unauthorized signature")
 
 // VerifyRef verifies the signature on the latest RSL entry for the target ref
 // using the latest policy.
```
wlynch, line 42: Update godoc to explain what the returned hash value represents.

internal/repository/verify.go
```diff
@@ -4,22 +4,36 @@ package repository
...
 import (
 	"context"
+	"errors"
 
 	"github.com/gittuf/gittuf/internal/gitinterface"
 	"github.com/gittuf/gittuf/internal/policy"
+	"github.com/gittuf/gittuf/internal/third_party/go-git/plumbing"
 )
 
+var ErrRefStateDoesNotMatchRSL = errors.New("Git reference's current state does not match latest RSL entry") //nolint:stylecheck
```
wlynch, line 14: Add godoc explaining what this means / how a repo can get into this state / how to resolve?

wlynch (changes requested):
LGTM! We should add some more godoc though. 🙏

adityasaky:
Done!

wlynch (approved):

Issue #202, opened 2023-11-24
[Bug]: clone does not handle slashes at the end of repository path.


spectre10:
### What happened?

the Clone func added in #134 does not handle slashes at the end of repository path. 
```go
func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Repository, error) {
	if dir == "" {
		// FIXME: my understanding is backslashes are not used in URLs but I haven't dived into the RFCs to check yet
		split := strings.Split(strings.TrimSpace(strings.ReplaceAll(remoteURL, "\\", "/")), "/")
		dir = strings.TrimSuffix(split[len(split)-1], ".git")
	}
	..
	..
	if err := os.Mkdir(dir, 0755); err != nil {
		return nil, errors.Join(ErrCloningRepository, err)
	}
	..
}
```
Here, if we provide slashes at the end of remoteURL, split[len(split)-1] results in an empty string. Consequently, dir also results in an empty string. Afterwards, when dir is provided to Mkdir, it results in an error.

### Relevant log output

```Shell
$ ./gittuf clone https://github.com/gittuf/ci-demo.git/
Error: unable to clone repository
mkdir : no such file or directory
Usage:
  gittuf clone [flags]

Flags:
  -b, --branch string   Specify branch to check out
  -h, --help            help for clone
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

spectre10:
might be relevent #140

adityasaky:
This is simpler than #140. Can you submit a patch to strip trailing slashes?

adityasaky:
I'd probably strip it for the dirname rather than the URL as a whole first, and let go-git's URL clean up handle it after

spectre10:
> This is simpler than #140. Can you submit a patch to strip trailing slashes?

Sure. #203

adityasaky:
Fixed in #203

PR #203, opened 2023-11-24
repository: Fix clone not handling trailing slashes in repository path.

spectre10:
Addresses #202

### DIFF
--- a/internal/repository/sync.go
+++ b/internal/repository/sync.go
@@ -26,7 +26,10 @@ var (
 func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Repository, error) {
 	if dir == "" {
 		// FIXME: my understanding is backslashes are not used in URLs but I haven't dived into the RFCs to check yet
-		split := strings.Split(strings.TrimSpace(strings.ReplaceAll(remoteURL, "\\", "/")), "/")
+		modifiedURL := strings.ReplaceAll(remoteURL, "\\", "/")
+		modifiedURL = strings.TrimRight(strings.TrimSpace(modifiedURL), "/") // Trim spaces and trailing slashes if any
+
+		split := strings.Split(modifiedURL, "/")
 		dir = strings.TrimSuffix(split[len(split)-1], ".git")
 	}
 	_, err := os.Stat(dir)
--- a/internal/repository/sync_test.go
+++ b/internal/repository/sync_test.go
@@ -202,4 +202,32 @@ func TestClone(t *testing.T) {
 		_, err = Clone(context.Background(), remoteTmpDir, dirName, "")
 		assert.ErrorIs(t, err, ErrDirExists)
 	})
+
+	t.Run("successful clone without specifying dir, with trailing slashes in repository path", func(t *testing.T) {
+		localTmpDir := t.TempDir()
+
+		if err := os.Chdir(localTmpDir); err != nil {
+			t.Fatal(err)
+		}
+		defer os.Chdir(currentDir) //nolint:errcheck
+
+		repo, err := Clone(context.Background(), remoteTmpDir+"//", "", "")
+		assert.Nil(t, err)
+		head, err := repo.r.Head()
+		if err != nil {
+			t.Fatal(err)
+		}
+		assert.Equal(t, commitID, head.Hash())
+
+		localRSLRef, err := repo.r.Reference(plumbing.ReferenceName(rsl.Ref), true)
+		if err != nil {
+			t.Fatal(err)
+		}
+		assert.Equal(t, remoteRSLRef.Hash(), localRSLRef.Hash())
+		localPolicyRef, err := repo.r.Reference(plumbing.ReferenceName(policy.PolicyRef), true)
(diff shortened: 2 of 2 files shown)


internal/repository/sync.go
```diff
@@ -26,9 +26,13 @@ var (
 func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Repository, error) {
 	if dir == "" {
 		// FIXME: my understanding is backslashes are not used in URLs but I haven't dived into the RFCs to check yet
-		split := strings.Split(strings.TrimSpace(strings.ReplaceAll(remoteURL, "\\", "/")), "/")
+		modifiedURL := strings.ReplaceAll(remoteURL, "\\", "/")
+		modifiedURL = strings.TrimRight(strings.TrimSpace(modifiedURL), "/") // Trim space and trailing slashes if present
+
+		split := strings.Split(modifiedURL, "/")
 		dir = strings.TrimSuffix(split[len(split)-1], ".git")
```
adityasaky, line 33: 1. Can you trim the `/` _specifically_ for the `dir` rather than the URL?
2. Can you add a test with a trailing `/` and see if it passes?

I ask because I also want us to see if go-git's behavior matches git's. I think it does, the issue you opened failed at the directory creation.

adityasaky (changes requested):

internal/repository/sync.go
spectre10: I am not sure if I understand correctly. ModifiedURL is only used to create split. The original remoteURL does not get modified.

spectre10: > I ask because I also want us to see if go-git's behavior matches git's. I think it does, the issue you opened failed at the directory creation.

I just checked and it does not in some cases. If there are two trailing slashes go-git is unable to clone, while git in-fact does.

internal/repository/sync_test.go
```diff
@@ -202,4 +202,65 @@ func TestClone(t *testing.T) {
...
+		}
+		assert.Equal(t, remoteRSLRef.Hash(), localRSLRef.Hash())
+		localPolicyRef, err := repo.r.Reference(plumbing.ReferenceName(policy.PolicyRef), true)
+		if err != nil {
+			t.Fatal(err)
+		}
+		assert.Equal(t, remotePolicyRef.Hash(), localPolicyRef.Hash())
+	})
+
+	t.Run("successful clone with dir. With trailing slashes in repository path.", func(t *testing.T) {
```
spectre10, line 234: This will always succeed, because here the remoteURL is a (temp.) directory path. Go-git handles trailing slashes in directory path, but does not handle two trailing slashes in http url as git does. Should I clone [ci-demo](https://github.com/gittuf/ci-demo) in a temp dir to test that?

internal/repository/sync.go
adityasaky: > I am not sure if I understand correctly. ModifiedURL is only used to create split. The original remoteURL does not get modified.

Ah I see. I missed that, my bad. That should be okay then, I suggest we pass in the URL as-is to gitinterface for the most part.

adityasaky: > I just checked and it does not in some cases. If there are two trailing slashes go-git is unable to clone, while git in-fact does.

That sounds like a fairly straightforward bug to fix in go-git :eyes:

internal/repository/sync_test.go
```diff
@@ -202,4 +202,65 @@ func TestClone(t *testing.T) {
 		_, err = Clone(context.Background(), remoteTmpDir, dirName, "")
 		assert.ErrorIs(t, err, ErrDirExists)
 	})
+
+	t.Run("successful clone without specifying dir, with trailing slashes in repository path.", func(t *testing.T) {
```
adityasaky, line 206:
```suggestion
	t.Run("successful clone without specifying dir, with trailing slashes in repository path", func(t *testing.T) {
```

internal/repository/sync_test.go
adityasaky: Interesting. I want to minimize the number of "live" tests that make a network call, so I'm leaning no. I suggest we just submit a patch to go-git.

adityasaky, line 234:
```suggestion

	t.Run("successful clone with dir, with trailing slashes in repository path", func(t *testing.T) {
```

adityasaky: This applies to both of the new tests right? If we aren't really testing any behavior here at the moment, let's take it out.

spectre10: Done.

spectre10: > Interesting. I want to minimize the number of "live" tests that make a network call, so I'm leaning no. I suggest we just submit a patch to go-git.

I think that handling more than single slash is a trivial matter. Even git does not handle more than two slashes, while other commands such as curl handles any number of slashes. In any case, our tests do not fail, so I think we can submit a patch later on, if we need to.

adityasaky (approved):

adityasaky:
Thanks!

PR #205, opened 2023-11-25
common: Export test config and clock

adityasaky:
We have two copies of these vars in the repo. One is in `common`, the other in `gitinterface`. We can't avoid defining the second one as its a dependency for `common`. I'm also leaning against defining it just once in `gitinterface` as `common` basically just holds test specific helpers / data, so it's the more natural home. This PR exports the `common` one for other packages to use.

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -29,7 +29,7 @@ const (
 )
 
 var (
-	testGitConfig = &config.Config{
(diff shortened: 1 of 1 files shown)


adityasaky:
Used in a test for #168

wlynch (approved):

PR #207, opened 2023-11-27
design: Flesh out on-behalf-of workflows

adityasaky:
This is a first pass at noting how a gittuf user signing an RSL entry on behalf of a non gittuf user can authenticate the user.

At this point, the doc doesn't talk about the data model for how the RSL entry is signed (if any changes are needed) and how the evidence is logged in the repository. My inclination is to use the attestation model being developed in other open PRs. I think that bit can wait until later, once we have that ready to go, and in the meantime the change in this PR can flesh out the general mechanism.

### DIFF
--- a/docs/design-document.md
+++ b/docs/design-document.md
@@ -294,6 +294,52 @@ Reference authorizations are stored in a directory called
 must have the in-toto predicate type:
 `https://gittuf.dev/reference-authorization/v<VERSION>`.
 
+#### Authentication Evidence Attestations
+
+In certain workflows, it is necessary to authenticate an actor outside of the
+context of gittuf. For example, later in this document is a description of a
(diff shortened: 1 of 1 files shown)


adityasaky:
> I like it overall, but I wonder if you want to enumerate more about what these statements actually look like in practice.

I'm happy to do that, but it might have to wait until we have attestation support to capture the evidence here. Unless that's not what you meant?

adityasaky:
@JustinCappos I started working on evidence attestation type for this. Can you take a look? I think we can nail this down only further into building it out, but it might get the idea across for now.

JustinCappos (approved):
Looks fine to me.  I do worry that complex types would cause potential parsing issues for languages like C that handle them less well.  However, since we're a code-base-first project, I think this is fine here.

PR #209, opened 2023-11-27
Added Test for findDelegationEntry in policy.go

neilnaveen:
- Added Test for findDelegationEntry in policy.go

### DIFF
--- a/internal/policy/policy_test.go
+++ b/internal/policy/policy_test.go
@@ -339,94 +339,28 @@ func TestGetStateForCommit(t *testing.T) {
 	assert.Equal(t, firstState, state)
 }
 
-func TestFindDelegationEntry(t *testing.T) {
+func TestStateFindDelegationEntry(t *testing.T) {
+	// Test case when no delegation is found
 	t.Run("no delegation", func(t *testing.T) {
 		state := createTestStateWithPolicy(t)
-
 		entry, err := state.findDelegationEntry("random")
 		assert.ErrorIs(t, err, ErrDelegationNotFound)
-		assert.Equal(t, entry, tuf.Delegation{})
+		assert.Nil(t, entry)
 	})
+
+	// Test case for a simple delegation
 	t.Run("simple delegation", func(t *testing.T) {
-		state := createTestStateWithPolicy(t)
-		state.DelegationEnvelopes = map[string]*sslibdsse.Envelope{}
-
-		topLevelDelegations := &tuf.Delegations{Roles: []tuf.Delegation{{Name: "1"}}}
-		topLevelMetadata := tuf.TargetsMetadata{Delegations: topLevelDelegations}
-
-		topLevelEnv, err := dsse.CreateEnvelope(topLevelMetadata)
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		state.TargetsEnvelope = topLevelEnv
-		state.DelegationEnvelopes["targets"] = topLevelEnv
-
-		secondLevelDelegations := &tuf.Delegations{}
-		secondLevelMetadata := tuf.TargetsMetadata{Delegations: secondLevelDelegations}
-		secondLevelEnv, err := dsse.CreateEnvelope(secondLevelMetadata)
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		state.DelegationEnvelopes["1"] = secondLevelEnv
-
+		state := createTestStateWithDelegatedPolicies(t)
 		entry, err := state.findDelegationEntry("1")
 		assert.Nil(t, err)
-		assert.Equal(t, tuf.Delegation{Name: "1"}, entry)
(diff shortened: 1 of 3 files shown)


internal/policy/policy_test.go
```diff
@@ -337,3 +336,95 @@ func TestGetStateForCommit(t *testing.T) {
...
+
+func TestFindDelegationEntry(t *testing.T) {
+	t.Run("no delegation", func(t *testing.T) {
+		state := createTestStateWithPolicy(t)
+
+		entry, err := state.findDelegationEntry("random")
+		assert.ErrorIs(t, err, ErrDelegationNotFound)
+		assert.Equal(t, entry, tuf.Delegation{})
+	})
+	t.Run("simple delegation", func(t *testing.T) {
```
adityasaky, line 348:
```suggestion

	t.Run("simple delegation", func(t *testing.T) {
```

internal/policy/policy_test.go
```diff
@@ -337,3 +336,95 @@ func TestGetStateForCommit(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, firstState, state)
 }
+
+func TestFindDelegationEntry(t *testing.T) {
```
adityasaky, line 340:
```suggestion
func TestStateFindDelegationEntry(t *testing.T) {
```

internal/policy/policy_test.go
```diff
@@ -337,3 +336,95 @@ func TestGetStateForCommit(t *testing.T) {
...
+	t.Run("no delegation", func(t *testing.T) {
+		state := createTestStateWithPolicy(t)
+
+		entry, err := state.findDelegationEntry("random")
+		assert.ErrorIs(t, err, ErrDelegationNotFound)
+		assert.Equal(t, entry, tuf.Delegation{})
```
adityasaky, line 346: Hmm, here's what it returns for an error:

```
	if err != nil {
		return tuf.Delegation{}, err
	}
```

I think it's okay to sneak a change in to turn all of those returns into nil first, before the tests commit here. Could you?

internal/policy/policy_test.go
```diff
@@ -337,3 +336,95 @@ func TestGetStateForCommit(t *testing.T) {
...
+	})
+	t.Run("simple delegation", func(t *testing.T) {
+		state := createTestStateWithPolicy(t)
```
adityasaky, line 349: I wonder if we can have a helper like this one `createTestStateWithDelegatedPolicies`.

adityasaky: That would do away with all the temporary policies that manually modify `State`. Also, you could have actual policies there that we could use in other tests, say to verify signatures using delegated policies.

internal/policy/policy_test.go
```diff
@@ -337,3 +336,95 @@ func TestGetStateForCommit(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		state.DelegationEnvelopes["1"] = secondLevelEnv
+
+		entry, err := state.findDelegationEntry("1")
+		assert.Nil(t, err)
+		assert.Equal(t, tuf.Delegation{Name: "1"}, entry)
+	})
+	t.Run("delegation with multiple roles", func(t *testing.T) {
```
adityasaky, line 376:
```suggestion

	t.Run("delegation with multiple roles", func(t *testing.T) {
```

internal/policy/policy_test.go
```diff
@@ -337,3 +336,95 @@ func TestGetStateForCommit(t *testing.T) {
...
+		state := createTestStateWithPolicy(t)
```
adityasaky, line 377: same general comment here, I don't love modifying `state` like this as `findDelegationEntry` assumes `state.Verify()` has passed in the invoking function. It's not the end of the world but a clean helper with real policies would be nice. It could be as simple as targets delegating to role A and B, with role A delegating to role C and D for some simple branch / tag / file rules.

adityasaky (approved):
This is a solid start! I have some comments on how the test cases can be created in a more reusable manner.

internal/policy/policy_test.go
neilnaveen: I have moved it to a different helper func in helpers_test.go

internal/policy/policy_test.go

internal/policy/policy_test.go

internal/policy/policy_test.go
neilnaveen: I have turned them into nil's, but to do that, the return value had to be changed into a pointer

internal/policy/policy_test.go
neilnaveen: fixed

internal/policy/policy_test.go

adityasaky:
Can you fix conflicts?

adityasaky (approved):

neilnaveen:
> Can you fix conflicts?

Fixed the conflicts

adityasaky (approved):

Issue #121, opened 2023-09-18
Support applying file policies in verify-commit


adityasaky:
Add `applyFilePolicies` flag that uses the commitPolicy to check that the commit signature is from a key trusted for all the paths modified by the commit.

#101

adityasaky:
We've dropped verify-commit since this was opened.

PR #214, opened 2023-12-01
Added flag to VerifyCommit for extra verification

neilnaveen:
- Added flag to VerifyCommit so that it can check if a given commit is allowed to modify the files it is modifying.
- Solves issue #121 

Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/internal/policy/verify.go
+++ b/internal/policy/verify.go
@@ -1025,14 +1025,14 @@ func checkCommitAgainstModifiedPaths(ctx context.Context, repo *git.Repository, 
 
 		for _, verifier := range verifiers {
 			err := verifier.Verify(ctx, commit, referenceAttesation)
-			if err == nil {
+			switch {
+			case err == nil:
 				goodKey = true
-				break
-			} else if errors.Is(err, gitinterface.ErrUnknownSigningMethod) {
+			case errors.Is(err, gitinterface.ErrUnknownSigningMethod):
 				// We encounter this for key types that can be used for gittuf
 				// policy metadata but not Git objects
 				continue
-			} else if !errors.Is(err, gitinterface.ErrIncorrectVerificationKey) {
+			case !errors.Is(err, gitinterface.ErrIncorrectVerificationKey):
 				// Unexpected error
 				break
 			}
--- a/internal/policy/verify.go
+++ b/internal/policy/verify.go
@@ -30,6 +30,7 @@ const (
 	unableToLoadPolicyMessageFmt      = "unable to load applicable gittuf policy: %s"
 	unableToFindPolicyMessage         = "unable to find applicable gittuf policy"
 	goodSignatureMessageFmt           = "good signature from key '%s:%s'"
+	goodApplyFilePoliciesMessage      = "commit is allowed to modify all paths it is modifying"
 	goodTagSignatureMessage           = "good signature for RSL entry and tag"
 	goodSignatureMessageForRSLEntry   = "good signature for RSL entry"
 	badSignatureMessageForRSLEntry    = "bad signature for RSL entry"
@@ -353,7 +354,7 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
 // consumed directly by the user, as this is used for a special, user-invoked
 // workflow. gittuf's other verification workflows are currently not expected to
 // use this function.
-func VerifyCommit(ctx context.Context, repo *git.Repository, ids ...string) map[string]string {
+func VerifyCommit(ctx context.Context, repo *git.Repository, applyFilePolicies bool, ids ...string) map[string]string {
 	status := make(map[string]string, len(ids))
 	commits := make(map[string]*object.Commit, len(ids))
 
@@ -400,10 +401,6 @@ func VerifyCommit(ctx context.Context, repo *git.Repository, ids ...string) map[
 			continue
 		}
 
-		// TODO: Add `applyFilePolicies` flag that uses the commitPolicy to
-		// check that the commit signature is from a key trusted for all the
-		// paths modified by the commit.
-
 		keys, err := commitPolicy.PublicKeys()
 		if err != nil {
 			status[id] = fmt.Sprintf(unableToLoadPolicyMessageFmt, err.Error())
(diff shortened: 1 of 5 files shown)


neilnaveen:
I currently have not refactored these lines https://github.com/gittuf/gittuf/blob/d3b0ae454e35ea1c6331c6ff95793c5dbc931127/internal/policy/verify.go#L204-L225 to use the helper function verifyKeys, becasue [FindPublicKeysForPath](https://github.com/gittuf/gittuf/blob/d3b0ae454e35ea1c6331c6ff95793c5dbc931127/internal/policy/policy.go#L318-L367) returns a slice of keys, while just retrieving all public keys using [policy.PublicKeys()](https://github.com/gittuf/gittuf/blob/d3b0ae454e35ea1c6331c6ff95793c5dbc931127/internal/policy/policy.go#L236C1-L282) for a policy returns a map.

internal/policy/verify.go
```diff
@@ -643,3 +649,46 @@ func getChangedPaths(repo *git.Repository, entry *rsl.ReferenceEntry) ([]string,
...
+	filePaths, err := gitinterface.GetFilePathsChangedByCommit(repo, commit)
+	if err != nil {
+		return []string{}, err
+	}
+
+	badPaths := []string{}
+	for _, path := range filePaths {
+		keys, err := commitPolicy.FindPublicKeysForPath(ctx, fmt.Sprintf("file:%s", path))
+		if err != nil {
+			return []string{}, err
```
adityasaky, line 663:
```suggestion
			return nil, err
```

internal/policy/verify.go
```diff
@@ -643,3 +649,46 @@ func getChangedPaths(repo *git.Repository, entry *rsl.ReferenceEntry) ([]string,
...
+		}
+
+		// Verify the keys associated with the commit. If verification fails, add the path to the list of bad paths.
+		err = verifyKeys(ctx, commit, keys)
```
adityasaky, line 667: I appreciate using a helper but given we're consolidating this in #168, you can probably skip this.

internal/policy/verify.go
```diff
@@ -643,3 +649,46 @@ func getChangedPaths(repo *git.Repository, entry *rsl.ReferenceEntry) ([]string,
...
+		err := gitinterface.VerifyCommitSignature(ctx, commit, key)
+		if err == nil {
+			return fmt.Errorf(goodSignatureMessageFmt, key.KeyType, key.KeyID)
+		}
+
+		if errors.Is(err, gitinterface.ErrUnknownSigningMethod) {
+			// We encounter this for key types that can be used for gittuf
+			// policy metadata but not Git objects
+			continue
+		}
```
adityasaky, line 690: You want to handle a case where it's not `ErrIncorrectVerificationKey` after this for cases where you have an unexpected error from `VerifyCommitSignature`.

internal/policy/verify_test.go
```diff
@@ -141,14 +141,14 @@ func TestVerifyCommit(t *testing.T) {
...
-	status = VerifyCommit(testCtx, repo, tagHash.String())
+	status = VerifyCommit(testCtx, repo, true, tagHash.String())
 	assert.Equal(t, expectedStatus, status)
 
 	// Add a commit but don't record it in the RSL
 	commitIDs = common.AddNTestCommitsToSpecifiedRef(t, repo, refName, 1, gpgKeyName)
 
 	expectedStatus = map[string]string{commitIDs[0].String(): unableToFindPolicyMessage}
-	status = VerifyCommit(testCtx, repo, commitIDs[0].String())
+	status = VerifyCommit(testCtx, repo, true, commitIDs[0].String())
```
adityasaky, line 151: IIRC the test helper already has file policies for "1" and "2". Can you add a test where an unauthorized commit modifies these files?

neilnaveen: I have added a failing test where a commit edits an unauthorized file path. I had to change all the test to not use the flag applyFilePolicies, since it turns out all commits are editing unauthorized file paths. I think this is being set in the at the start of the test when the repo is being created with createTestStatewithPolicy. To create a valid test, either the policy would have to be changed outside of the func, or the function itself would have to be changed.

internal/policy/verify.go
neilnaveen: Handled

internal/policy/verify.go
neilnaveen: I have removed the helper function

internal/policy/verify.go

internal/policy/verify.go
```diff
@@ -643,3 +649,46 @@ func getChangedPaths(repo *git.Repository, entry *rsl.ReferenceEntry) ([]string,
 
 	return gitinterface.GetDiffFilePaths(currentCommit, priorCommit)
 }
+
+func checkCommitAgainstModifiedPaths(ctx context.Context, repo *git.Repository, commit *object.Commit, commitPolicy *State) ([]string, error) {
+	filePaths, err := gitinterface.GetFilePathsChangedByCommit(repo, commit)
+	if err != nil {
+		return []string{}, err
```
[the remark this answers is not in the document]

internal/policy/verify.go
```diff
@@ -192,9 +192,15 @@ func VerifyCommit(ctx context.Context, repo *git.Repository, ids ...string) map[
...
-		// TODO: Add `applyFilePolicies` flag that uses the commitPolicy to
-		// check that the commit signature is from a key trusted for all the
-		// paths modified by the commit.
+		// This flag tells us to check if the file paths modified by the commit are allowed by the policy
+		if applyFilePolicies {
+			badPaths, err := checkCommitAgainstModifiedPaths(ctx, repo, commit, commitPolicy)
+			if err != nil {
+				status[id] = err.Error()
+			} else if len(badPaths) > 0 {
+				status[id] = fmt.Sprintf("commit is not allowed to modify the following paths: %s", strings.Join(badPaths, ", "))
```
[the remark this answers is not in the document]
neilnaveen: Added

adityasaky:
Blocked for now by #219

neilnaveen:
> Blocked for now by #219

@adityasaky I have rebased with the changes from 219

internal/policy/verify.go
```diff
@@ -643,3 +649,46 @@ func getChangedPaths(repo *git.Repository, entry *rsl.ReferenceEntry) ([]string,
...
+
+func checkCommitAgainstModifiedPaths(ctx context.Context, repo *git.Repository, commit *object.Commit, commitPolicy *State) ([]string, error) {
+	filePaths, err := gitinterface.GetFilePathsChangedByCommit(repo, commit)
+	if err != nil {
+		return []string{}, err
+	}
+
+	badPaths := []string{}
+	for _, path := range filePaths {
+		keys, err := commitPolicy.FindPublicKeysForPath(ctx, fmt.Sprintf("file:%s", path))
```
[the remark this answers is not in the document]
adityasaky: bump on this.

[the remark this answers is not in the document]
adityasaky: I also want us to start thinking about this in the context of attestations, because a policy might require more than 1 approval for a file change...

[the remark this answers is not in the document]
neilnaveen: I have changed it to use FindVerifiersForPath, instead of FindPublicKeysForPath, and the verification is currently using the current policies TargetEnvelope. When PR #261 lands, this will have to change, and the whole delegation tree will have to be traversed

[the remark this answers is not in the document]
neilnaveen: I have changed this to use the current attestations in the repo.

neilnaveen:
I have added attestation support for VerifyCommit since if we have access to the namespace and the rule has a higher threshold; this check would have failed in the previous version of the applyFilePolicy check. Without attestations, this PR would output false negatives instead of VerifyCommit's current state of reporting false positives.

Issue #191, opened 2023-11-17
Policy / rule examination interface


adityasaky:
gittuf currently does not have subcommands to examine a policy / rules. If a user wants to inspect what policies are applicable for some state, they have to manually examine the different envelopes (decoding the payloads in the process) tracked at `refs/gittuf/policy`. As @errordeveloper points out in #4

> One low-hanging fruit seems to be gittuf policy {list-rules,show-rule}. As a newbie, I was surprised that there isn't a way to see what policies are already defined.

adityasaky:
Closed by #221.

PR #221, opened 2023-12-08
Added list-rules to cmd

neilnaveen:
- Added the list-rules command, which list all current rules for the latest policy state.
- Fixes issue #191

### DIFF
--- a/internal/cmd/policy/listrules/listrules.go
+++ b/internal/cmd/policy/listrules/listrules.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/gittuf/gittuf/internal/cmd/common"
 	"github.com/gittuf/gittuf/internal/policy"
 	"github.com/gittuf/gittuf/internal/repository"
 	"github.com/spf13/cobra"
@@ -67,10 +66,9 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 func New() *cobra.Command {
 	o := &options{}
 	cmd := &cobra.Command{
-		Use:     "list-rules",
-		Short:   "List rules for the current state",
-		PreRunE: common.CheckIfSigningViable,
-		RunE:    o.Run,
+		Use:   "list-rules",
+		Short: "List rules for the current state",
+		RunE:  o.Run,
 	}
 	o.AddFlags(cmd)
 
--- a/internal/cmd/policy/listrules/listrules.go
+++ b/internal/cmd/policy/listrules/listrules.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/gittuf/gittuf/internal/policy"
 	"github.com/gittuf/gittuf/internal/repository"
 	"github.com/spf13/cobra"
 )
@@ -30,9 +29,6 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 	// The order is a pre-order traversal of the delegation tree, so that the parent is always before the children.
 
 	for _, curRule := range rules {
-		if curRule.Delegation.Name == policy.AllowRuleName {
-			continue
-		}
 		fmt.Printf(strings.Repeat("    ", curRule.Depth)+"Rule %s:\n", curRule.Delegation.Name)
 		gitpaths, filepaths := []string{}, []string{}
 		for _, path := range curRule.Delegation.Paths {
--- a/internal/cmd/policy/listrules/listrules.go
+++ b/internal/cmd/policy/listrules/listrules.go
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package listrules
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/gittuf/gittuf/internal/cmd/common"
+	"github.com/gittuf/gittuf/internal/policy"
+	"github.com/gittuf/gittuf/internal/repository"
+	"github.com/spf13/cobra"
+)
+
+type options struct{}
+
+func (o *options) AddFlags(_ *cobra.Command) {}
+
+func (o *options) Run(cmd *cobra.Command, _ []string) error {
+	repo, err := repository.LoadRepository()
+	if err != nil {
+		return err
+	}
+
+	rules, err := repo.ListRules(cmd.Context())
+	if err != nil {
+		return err
+	}
+
+	// Iterate through the rules, they are already in order, and the depth tells us how to indent.
+	// The order is a pre-order traversal of the delegation tree, so that the parent is always before the children.
+
+	for _, curRule := range rules {
+		if curRule.Delegation.Name == policy.AllowRuleName {
+			continue
+		}
+		fmt.Printf(strings.Repeat("    ", curRule.Depth)+"Rule %s:\n", curRule.Delegation.Name)
+		gitpaths, filepaths := []string{}, []string{}
+		for _, path := range curRule.Delegation.Paths {
+			if strings.HasPrefix(path, "git:") {
+				gitpaths = append(gitpaths, path)
+			} else {
+				filepaths = append(filepaths, path)
+			}
+		}
+		if len(filepaths) > 0 {
+			fmt.Println(strings.Repeat("    ", curRule.Depth+1) + "Paths affected:")
+			for _, v := range filepaths {
+				fmt.Printf(strings.Repeat("    ", curRule.Depth+2)+"%s\n", v)
+			}
+		}
+		if len(gitpaths) > 0 {
+			fmt.Println(strings.Repeat("    ", curRule.Depth+1) + "Refs affected:")
+			for _, v := range gitpaths {
+				fmt.Printf(strings.Repeat("    ", curRule.Depth+2)+"%s\n", v)
+			}
+		}
+
+		fmt.Println(strings.Repeat("    ", curRule.Depth+1) + "Authorized keys:")
+		for _, key := range curRule.Delegation.Role.KeyIDs {
+			fmt.Printf(strings.Repeat("    ", curRule.Depth+2)+"%s\n", key)
+		}
+	}
+	return nil
+}
+
+func New() *cobra.Command {
+	o := &options{}
+	cmd := &cobra.Command{
+		Use:     "list-rules",
+		Short:   "List rules for the current state",
+		PreRunE: common.CheckIfSigningViable,
+		RunE:    o.Run,
+	}
+	o.AddFlags(cmd)
+
+	return cmd
+}
--- a/internal/cmd/policy/policy.go
+++ b/internal/cmd/policy/policy.go
@@ -6,6 +6,7 @@ import (
 	"github.com/gittuf/gittuf/internal/cmd/policy/addkey"
 	"github.com/gittuf/gittuf/internal/cmd/policy/addrule"
 	i "github.com/gittuf/gittuf/internal/cmd/policy/init"
+	"github.com/gittuf/gittuf/internal/cmd/policy/listrules"
 	"github.com/gittuf/gittuf/internal/cmd/policy/persistent"
 	"github.com/gittuf/gittuf/internal/cmd/policy/removerule"
 	"github.com/gittuf/gittuf/internal/cmd/trustpolicy/remote"
@@ -23,6 +24,7 @@ func New() *cobra.Command {
 	cmd.AddCommand(i.New(o))
 	cmd.AddCommand(addkey.New(o))
 	cmd.AddCommand(addrule.New(o))
+	cmd.AddCommand(listrules.New())
 	cmd.AddCommand(removerule.New(o))
 
 	remoteCmd := remote.New()
--- a/internal/policy/helpers_test.go
+++ b/internal/policy/helpers_test.go
@@ -159,11 +159,6 @@ func createTestStateWithDelegatedPolicies(t *testing.T) *State {
 	if err != nil {
 		t.Fatal(err)
 	}
-	gpgkey, err := gpg.LoadGPGKeyFromBytes(gpgPubKeyBytes)
-
-	if err != nil {
-		t.Fatal(err)
-	}
 
 	rootMetadata := InitializeRootMetadata(key)
 
@@ -174,6 +169,11 @@ func createTestStateWithDelegatedPolicies(t *testing.T) *State {
 		t.Fatal(err)
 	}
 	rootEnv, err = dsse.SignEnvelope(context.Background(), rootEnv, signer)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	gpgkey, err := gpg.LoadGPGKeyFromBytes(gpgPubKeyBytes)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/internal/policy/helpers_test.go
+++ b/internal/policy/helpers_test.go
@@ -173,7 +173,7 @@ func createTestStateWithDelegatedPolicies(t *testing.T) *State {
 		t.Fatal(err)
 	}
 
-	gpgkey, err := gpg.LoadGPGKeyFromBytes(gpgPubKeyBytes)
+	gpgKey, err := gpg.LoadGPGKeyFromBytes(gpgPubKeyBytes)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -203,12 +203,12 @@ func createTestStateWithDelegatedPolicies(t *testing.T) *State {
 
 	// Create the second level of delegations
 	delegation1Metadata := InitializeTargetsMetadata()
-	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "3", []*tuf.Key{gpgkey}, []string{"file:1/subpath1/*"})
+	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "3", []*tuf.Key{gpgKey}, []string{"file:1/subpath1/*"})
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "4", []*tuf.Key{gpgkey}, []string{"file:1/subpath2/*"})
+	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "4", []*tuf.Key{gpgKey}, []string{"file:1/subpath2/*"})
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/internal/policy/helpers_test.go
+++ b/internal/policy/helpers_test.go
@@ -159,6 +159,11 @@ func createTestStateWithDelegatedPolicies(t *testing.T) *State {
 	if err != nil {
 		t.Fatal(err)
 	}
+	gpgkey, err := gpg.LoadGPGKeyFromBytes(gpgPubKeyBytes)
+
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	rootMetadata := InitializeRootMetadata(key)
 
@@ -169,11 +174,6 @@ func createTestStateWithDelegatedPolicies(t *testing.T) *State {
 		t.Fatal(err)
 	}
 	rootEnv, err = dsse.SignEnvelope(context.Background(), rootEnv, signer)
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	gpgKey, err := gpg.LoadGPGKeyFromBytes(gpgPubKeyBytes)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -203,12 +203,12 @@ func createTestStateWithDelegatedPolicies(t *testing.T) *State {
 
 	// Create the second level of delegations
 	delegation1Metadata := InitializeTargetsMetadata()
-	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "3", []*tuf.Key{gpgKey}, []string{"file:1/subpath1/*"})
+	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "3", []*tuf.Key{gpgkey}, []string{"file:1/subpath1/*"})
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "4", []*tuf.Key{gpgKey}, []string{"file:1/subpath2/*"})
+	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "4", []*tuf.Key{gpgkey}, []string{"file:1/subpath2/*"})
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -223,16 +223,6 @@ func createTestStateWithDelegatedPolicies(t *testing.T) *State {
 		t.Fatal(err)
 	}
 
-	// Create the empty delegation envelope for delegation 2, 3, and 4
-	emptyEnv, err := dsse.CreateEnvelope(InitializeTargetsMetadata())
-	if err != nil {
-		t.Fatal(err)
-	}
-	emptyEnv, err = dsse.SignEnvelope(context.Background(), emptyEnv, signer)
-	if err != nil {
-		t.Fatal(err)
-	}
-
 	curState := &State{
 		RootEnvelope:        rootEnv,
 		TargetsEnvelope:     targetsEnv,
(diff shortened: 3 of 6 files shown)


internal/policy/policy.go
```diff
@@ -695,3 +694,17 @@ func (s *State) findDelegationEntry(roleName string) (tuf.Delegation, error) {
...
 	}
 }
+
+func ListRules(ctx context.Context, repo *git.Repository) ([]tuf.Delegation, error) {
+	state, err := LoadCurrentState(ctx, repo)
+	if err != nil {
+		return nil, err
+	}
+
+	targetsMetadata, err := state.GetTargetsMetadata(TargetsRoleName)
```
adityasaky, line 704: We need to explore delegated roles as well.

internal/policy/policy.go
```diff
@@ -695,3 +694,17 @@ func (s *State) findDelegationEntry(roleName string) (tuf.Delegation, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return targetsMetadata.Delegations.Roles, nil
```
adityasaky, line 709: let's skip displaying the allow rule

adityasaky (changes requested):
Good start! I wonder what the best way to display delegated rules are. cc @wlynch

internal/policy/policy.go
neilnaveen: Added

internal/policy/policy.go
neilnaveen: I have not changed the ListRules function in policy, instead I just excluded showing it when it was being outputted.

internal/cmd/policy/listrules/listrules.go
```diff
@@ -0,0 +1,90 @@
...
+		return err
+	}
+
+	rules, err := repo.ListRules(cmd.Context())
+	if err != nil {
+		return err
+	}
+
+	for _, rule := range rules {
+		if rule.Name == "gittuf-allow-rule" {
```
adityasaky, line 34:
```suggestion
		if rule.Name == policy.AllowRuleName {
```

internal/cmd/policy/listrules/listrules.go
```diff
@@ -0,0 +1,90 @@
...
+			continue
+		}
+
+		fmt.Printf("Rule %s:\n", rule.Name)
+		m := map[string][]string{}
+		for _, path := range rule.Paths {
+			if strings.HasPrefix(path, "git:") {
+				m["git"] = append(m["git"], path)
```
adityasaky, line 42: The slices haven't been initialized right? I expect a panic here. FWIW I don't think the map is necessary. You could just have two initialized slices for git and file rules and append to them.

internal/policy/policy.go
```diff
@@ -765,3 +765,48 @@ func (s *State) findDelegationEntry(roleName string) (*tuf.Delegation, error) {
...
+	allDelegations := topLevelTargetsMetadata.Delegations.Roles
+
+	for len(delegationsToSearch) > 0 {
+		delegation := delegationsToSearch[0]
+		delegationsToSearch = delegationsToSearch[1:]
+
+		if state.HasTargetsRole(delegation.Name) {
+			// TODO: clarifying terminating / reachability
+
+			allDelegations = append(allDelegations, delegationTargetsMetadata[delegation.Name].Delegations.Roles...)
```
adityasaky, line 806: The order matters here. I suggest placing the delegations upfront so that delegated rules are placed right after the delegating rule. I also wonder if it'd be possible indent per delegation level in the CLI output so it's immediately clear how a rule is reached. WDYT?

adityasaky (changes requested):
I think this is closer but some suggestions on how delegated rules may be communicated so the reachability is more obvious to a user?

internal/policy/policy.go
neilnaveen: I have changed the ListRules func, so that it returns a slice of delegations with their depth, so different delegations, based on their depth in the delegation tree are indented differently.

internal/cmd/policy/listrules/listrules.go
neilnaveen: Fixed

internal/cmd/policy/listrules/listrules.go
neilnaveen: Fixed

internal/cmd/policy/policy.go
```diff
@@ -24,6 +25,7 @@ func New() *cobra.Command {
 	cmd.AddCommand(addkey.New(o))
 	cmd.AddCommand(addrule.New(o))
 	cmd.AddCommand(removerule.New(o))
+	cmd.AddCommand(listrules.New(o))
```
adityasaky, line 28: nit: can you place this line above `removerule` to be alphabetical? I'm aware this isn't consistent right now but as we add subcommands, it'll be helpful to order them upfront.

internal/policy/policy.go
```diff
@@ -82,6 +82,11 @@ type State struct {
 	RootPublicKeys      []*tuf.Key
 }
 
+type DelegationDepth struct {
```
adityasaky, line 85:
```suggestion
type DelegationWithDepth struct {
```

internal/cmd/policy/listrules/listrules.go
```diff
@@ -0,0 +1,91 @@
...
+				fmt.Printf(strings.Repeat("    ", curRule.Depth+2)+"%s\n", v)
+			}
+		}
+
+		fmt.Println(strings.Repeat("    ", curRule.Depth+1) + "Authorized keys:")
+		for _, key := range curRule.Delegation.Role.KeyIDs {
+			fmt.Printf(strings.Repeat("    ", curRule.Depth+2)+"%s\n", key)
+		}
+
+		if curRule.Delegation.Custom != nil {
```
adityasaky, line 67: I'd remove this altogether. We have some ideas to eventually put something here, but then we'll know exactly what fields they are and we can represent them better.

internal/cmd/policy/listrules/listrules.go
```diff
@@ -0,0 +1,91 @@
...
+			}
+			fmt.Printf(strings.Repeat("    ", curRule.Depth+1)+"Details: %s\n", string(customString))
+		}
+
+		fmt.Println()
+	}
+	return nil
+}
+
+func New(persistent *persistent.Options) *cobra.Command {
```
adityasaky, line 80: I think we don't need `persistent` for this one.

internal/policy/policy.go
```diff
@@ -765,3 +770,54 @@ func (s *State) findDelegationEntry(roleName string) (*tuf.Delegation, error) {
 		}
 	}
 }
+
+func ListRules(ctx context.Context, repo *git.Repository) ([]*DelegationDepth, error) {
```
adityasaky, line 774: missing docstring

internal/policy/policy.go
```diff
@@ -765,3 +770,54 @@ func (s *State) findDelegationEntry(roleName string) (*tuf.Delegation, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	topLevelTargetsMetadata, err := state.GetTargetsMetadata(TargetsRoleName)
+	if err != nil {
+		return nil, err
+	}
+
+	delegationTargetsMetadata := map[string]*tuf.TargetsMetadata{}
```
adityasaky, line 785: Any reason to populate these upfront? I suggest using `GetTargetsMetadata` when `HasTargetsRole` is for some delegation. That'll limit loading to _reachable_ roles.

internal/policy/policy.go
```diff
@@ -765,3 +770,54 @@ func (s *State) findDelegationEntry(roleName string) (*tuf.Delegation, error) {
...
+		delegationsToSearch = append(delegationsToSearch, &DelegationDepth{Delegation: topLevelDelegation, Depth: 0})
+		allDelegations = append(allDelegations, &DelegationDepth{Delegation: topLevelDelegation, Depth: 0})
+	}
+
+	for len(delegationsToSearch) > 0 {
+		currentDelegation := delegationsToSearch[0]
+		delegationsToSearch = delegationsToSearch[1:]
+
+		if state.HasTargetsRole(currentDelegation.Delegation.Name) {
+			// TODO: clarifying terminating / reachability
```
adityasaky, line 813: terminating does not apply when we aren't querying for a specific namespace, so this can be taken out.

internal/policy/policy.go
adityasaky: Also, please move this function above so all the private functions are at the end of the file.

internal/policy/policy_test.go
```diff
@@ -409,3 +407,52 @@ func TestStateFindDelegationEntry(t *testing.T) {
 		assert.Equal(t, &tuf.Delegation{Name: "4", Paths: []string{"path1/subpath2/*"}, Terminating: false, Role: tuf.Role{KeyIDs: []string{"157507bbe151e378ce8126c1dcfe043cdd2db96e"}, Threshold: 1}}, entry)
 	})
 }
+
+func TestListRules(t *testing.T) {
+	repo, _ := createTestRepository(t, createTestStateWithPolicy)
```
adityasaky, line 412: Can we test with delegated roles? Keep this one and move to two `t.Run`s.

adityasaky (changes requested):
This is looking good!

internal/policy/policy.go
```diff
@@ -765,3 +770,54 @@ func (s *State) findDelegationEntry(roleName string) (*tuf.Delegation, error) {
...
+
+			for _, delegation := range delegationTargetsMetadata[currentDelegation.Delegation.Name].Delegations.Roles {
+				allDelegations = append([]*DelegationDepth{{Delegation: delegation, Depth: currentDelegation.Depth + 1}}, allDelegations...)
```
adityasaky, line 816: I see how these are different, but please add a comment why the same append (at first glance) op is called on the two different slices.

adityasaky:
Also, eventually, we may want to parametrize this _with a command_. For example `gittuf policy list-rules-for refs/heads/main` etc.

internal/cmd/policy/policy.go
neilnaveen: Moved

internal/policy/policy.go

internal/cmd/policy/listrules/listrules.go
neilnaveen: I have removed this

internal/cmd/policy/listrules/listrules.go
neilnaveen: Removed

internal/policy/policy.go
neilnaveen: I have moved it and added a docstring

internal/policy/policy.go
neilnaveen: Removed

internal/policy/policy_test.go
neilnaveen: I have added another test for delegated roles, but this meant that I had to change createTestStateWithPolicy to use the root key, since using GPG keys for delegated entries makes them fail the state.Verify check.

internal/policy/policy.go
neilnaveen: Added

internal/cmd/policy/listrules/listrules.go
```diff
@@ -0,0 +1,79 @@
...
+			for _, v := range gitpaths {
+				fmt.Printf(strings.Repeat("    ", curRule.Depth+2)+"%s\n", v)
+			}
+		}
+
+		fmt.Println(strings.Repeat("    ", curRule.Depth+1) + "Authorized keys:")
+		for _, key := range curRule.Delegation.Role.KeyIDs {
+			fmt.Printf(strings.Repeat("    ", curRule.Depth+2)+"%s\n", key)
+		}
+		fmt.Println()
```
adityasaky, line 63:
```suggestion
```

adityasaky: Let's avoid the trailing newline, perhaps this should go somewhere else if the concern is a newline at the start of a rule.

internal/policy/policy.go
```diff
@@ -567,7 +572,7 @@ func (s *State) Verify(ctx context.Context) error {
 	}
 
 	if len(delegationEnvelopes) != 0 {
-		return ErrDanglingDelegationMetadata
+		return fmt.Errorf(fmt.Sprint(delegationEnvelopes))
```
adityasaky, line 575: can you explain this change?

adityasaky: Also beware of #219...

adityasaky: I suggest leaving `Verify` alone.

internal/policy/policy_test.go
adityasaky: the leaf delegation can still point to a gpg key no?

internal/policy/helpers_test.go
```diff
@@ -203,12 +198,13 @@ func createTestStateWithDelegatedPolicies(t *testing.T) *State {
 
 	// Create the second level of delegations
 	delegation1Metadata := InitializeTargetsMetadata()
-	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "3", []*tuf.Key{gpgKey}, []string{"path1/subpath1/*"})
+
+	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "3", []*tuf.Key{key}, []string{"path1/subpath1/*"})
```
adityasaky, line 202: these paths wouldn't do much anyway, because they don't have the `file:` / `git:` path schemes.

adityasaky: perhaps that should be resolved separately though.

adityasaky (changes requested):

internal/policy/policy.go
neilnaveen: I was changing that when I was debugging, and I forgot to revert it

internal/policy/helpers_test.go
neilnaveen: So for now should it be left as it is, should I remove the file paths alltogheter, or should I format it correctly in the file: or git: format?

neilnaveen:
>the leaf delegation can still point to a gpg key no?

Are you talking about delegations 3 and 4 in the delegation tree, If so, I don't think that this is possible since verify would error out since NewSignerVerifierFromTUFKey cannot handle GPG keys?

internal/policy/policy.go

internal/cmd/policy/listrules/listrules.go
neilnaveen: Fixed

internal/policy/helpers_test.go
adityasaky: Can you submit a separate small PR focused on this helper that:
* fixes the paths
* uses gpg keys only in the leaf delegation (i.e. which does not point to another metadata)

It might be easier to merge that first and rebase this one.

neilnaveen: How would the leaf delegations be allowed to use GPG keys, wouldn't state verification fail?

neilnaveen:
> > the leaf delegation can still point to a gpg key no?
> 
> Are you talking about delegations 3 and 4 in the delegation tree, If so, I don't think that this is possible since verify would error out since NewSignerVerifierFromTUFKey cannot handle GPG keys?

This will be fixed by 219

internal/policy/helpers_test.go
neilnaveen: Will be fixed with PR #251

adityasaky: Let's reverse changes to the delegated policy here :)

neilnaveen: done

adityasaky:
@neilnaveen can you rebase this one now? It might be worth squashing commits first so you don't have to address the same conflicts multiple times.

neilnaveen:
I have rebased the changes introduced by PR #219

internal/cmd/policy/listrules/listrules.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return nil
+}
+
+func New() *cobra.Command {
+	o := &options{}
+	cmd := &cobra.Command{
+		Use:     "list-rules",
+		Short:   "List rules for the current state",
+		PreRunE: common.CheckIfSigningViable,
```
adityasaky, line 72: We don't need to be able to sign to view rules, this can be taken out.

internal/policy/helpers_test.go
```diff
@@ -156,6 +156,11 @@ func createTestStateWithDelegatedPolicies(t *testing.T) *State {
 	}
 
 	key, err := tuf.LoadKeyFromBytes(rootPubKeyBytes)
+	if err != nil {
+		t.Fatal(err)
+	}
+	gpgkey, err := gpg.LoadGPGKeyFromBytes(gpgPubKeyBytes)
```
adityasaky, line 162: We don't need this diff at all, right? Can you put gpgKey's load back where it was so this doesn't show up as a diff?

internal/policy/helpers_test.go
```diff
@@ -203,12 +203,12 @@ func createTestStateWithDelegatedPolicies(t *testing.T) *State {
 
 	// Create the second level of delegations
 	delegation1Metadata := InitializeTargetsMetadata()
-	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "3", []*tuf.Key{gpgKey}, []string{"file:1/subpath1/*"})
+	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "3", []*tuf.Key{gpgkey}, []string{"file:1/subpath1/*"})
```
adityasaky, line 206: This diff is because of the confusion with `gpgKey`. If you make it so gpgKey has no diff (as I say above), this line also should have no diff.

internal/policy/helpers_test.go
```diff
@@ -203,12 +203,12 @@ func createTestStateWithDelegatedPolicies(t *testing.T) *State {
...
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "4", []*tuf.Key{gpgKey}, []string{"file:1/subpath2/*"})
+	delegation1Metadata, err = AddOrUpdateDelegation(delegation1Metadata, "4", []*tuf.Key{gpgkey}, []string{"file:1/subpath2/*"})
```
adityasaky, line 211: same here

internal/policy/policy.go
```diff
@@ -629,6 +634,53 @@ func (s *State) HasTargetsRole(roleName string) bool {
 	return ok
 }
 
+// ListRules returns a list of all the rules as an array of the delegations in a pre order traversal of the
```
adityasaky, line 637: nit: 80 cols for docstrings.

```
// ListRules returns a list of all the rules as an array of the delegations in a
// pre order traversal of the delegation tree, with the depth of each
// delegation.
```

internal/policy/policy_test.go
```diff
@@ -427,3 +427,142 @@ func TestStateFindDelegationEntry(t *testing.T) {
 		assert.Equal(t, &tuf.Delegation{Name: "4", Paths: []string{"file:1/subpath2/*"}, Terminating: false, Role: tuf.Role{KeyIDs: []string{"157507bbe151e378ce8126c1dcfe043cdd2db96e"}, Threshold: 1}}, entry)
 	})
 }
+
+func TestListRules(t *testing.T) {
+	t.Run("t1", func(t *testing.T) {
```
adityasaky, line 432:
```suggestion
	t.Run("no delegations", func(t *testing.T) {
```

internal/policy/policy_test.go
```diff
@@ -427,3 +427,142 @@ func TestStateFindDelegationEntry(t *testing.T) {
...
+						KeyIDs:    []string{},
+						Threshold: 1,
+					},
+				},
+				Depth: 0,
+			},
+		}
+		assert.Equal(t, expectedRules, rules)
+	})
+	t.Run("t2", func(t *testing.T) {
```
adityasaky, line 480:
```suggestion
	t.Run("with delegations", func(t *testing.T) {
```

internal/policy/policy.go
```diff
@@ -629,6 +634,53 @@ func (s *State) HasTargetsRole(roleName string) bool {
...
+
+		// allDelegations will be the returned list of all the delegations in pre-order traversal, no delegations will be popped off
+		allDelegations = append(allDelegations, currentDelegation)
+		if state.HasTargetsRole(currentDelegation.Delegation.Name) {
+			currentMetadata, err := state.GetTargetsMetadata(currentDelegation.Delegation.Name)
+			if err != nil {
+				return nil, err
+			}
+
+			delegationsToSearch = append([]*DelegationWithDepth{{Delegation: currentMetadata.Delegations.Roles[len(currentMetadata.Delegations.Roles)-1], Depth: currentDelegation.Depth + 1}}, delegationsToSearch...)
```
adityasaky, line 669: Can you clarify this statement along with line 676?

internal/policy/policy.go
```diff
@@ -629,6 +634,53 @@ func (s *State) HasTargetsRole(roleName string) bool {
...
+			for _, delegation := range currentMetadata.Delegations.Roles {
+				if delegation.Name == AllowRuleName {
+					continue
+				}
+				// delegationsToSearch will contain the same delegations of allDelegations,
+				// but at every iteration, the next delegation will be popped off, leaving us at the end with no delegations.
+				delegationsToSearch = append([]*DelegationWithDepth{{Delegation: delegation, Depth: currentDelegation.Depth + 1}}, delegationsToSearch...)
```
adityasaky, line 676: Linked to comment about line 669.

internal/policy/policy_test.go
```diff
@@ -427,3 +427,142 @@ func TestStateFindDelegationEntry(t *testing.T) {
...
+					Role: tuf.Role{
+						KeyIDs:    []string{"52e3b8e73279d6ebdd62a5016e2725ff284f569665eb92ccb145d83817a02997"},
+						Threshold: 1,
+					},
+				},
+				Depth: 0,
+			},
+			{
+				Delegation: tuf.Delegation{
+					Name:        "4",
```
adityasaky, line 502: shouldn't 3 be listed before 4? Note that the order of rules is really important, as a terminating rule that appears first may mean the rule after is ignored.

internal/policy/policy_test.go
```diff
@@ -427,3 +427,142 @@ func TestStateFindDelegationEntry(t *testing.T) {
...
+					Custom:      nil,
+					Role: tuf.Role{
+						KeyIDs:    []string{"52e3b8e73279d6ebdd62a5016e2725ff284f569665eb92ccb145d83817a02997"},
+						Threshold: 1,
+					},
+				},
+				Depth: 0,
+			},
+			{
+				Delegation: tuf.Delegation{
```
adityasaky, line 553: I wonder if we should filter allow rule here rather than in the cmd package.

adityasaky: I don't like the idea of ListRules returning multiple allow rules in random spots :\

adityasaky (changes requested):

internal/policy/helpers_test.go
neilnaveen: Before it was tuf.LoadKeyFromBytes, which was failing, since it could not process a gpg key, so I had to switch it.

adityasaky: https://github.com/gittuf/gittuf/pull/221/files#diff-ec90ffc9ec0e3e5ac57fce38f8a23d4e1b139204a779ac29bf13baa4a3c6dd6dL176-L180 already was loading this key, my point is we don't need this "move" anymore. By moving this back to the original location and fixing the var name, it'll have no diff at all.

internal/policy/policy_test.go
neilnaveen: The order has been changed so that the original order is preserved.

internal/policy/policy.go
neilnaveen: I have added additional comments

internal/policy/policy.go
neilnaveen: This line is now gone, since we do not need the additional allow rules

internal/policy/policy_test.go

internal/policy/policy_test.go

internal/policy/policy.go

internal/policy/helpers_test.go
neilnaveen: Fixed

internal/policy/helpers_test.go
neilnaveen: Fixed

internal/policy/helpers_test.go
neilnaveen: I have fixed this, but it is still showing a diff, even though I cannot find one.

internal/cmd/policy/listrules/listrules.go
neilnaveen: Removed

internal/policy/policy_test.go
neilnaveen: Fixed

adityasaky (approved):
Thanks @neilnaveen!

PR #228, opened 2023-12-18
don't print cli usage on error

datosh:


adityasaky (approved):
Thanks! Could you resolve the conflict?

Issue #158, opened 2023-10-27
Support gittuf-specific Git hooks


adityasaky:
We need to define Git hooks for gittuf-specific actions that the gittuf command can apply to the specified repository. For this, we need to first define the operations that must be invoked via a Git hook, then define platform-specific hooks that get written to `.git/hooks/`. For example, a `pre-push` hook could be configured to create RSL entry for the ref being pushed. These hooks could be installed using a command like `gittuf add-hooks`.

datosh:
I will start with a PoC on this one.

I will add functionality to register hooks with the proposed `gittuf add-hooks`. 

What's the rationale behind creating a RSL entry when pushing. My intuition would have been to add an entry for each commit instead, and then a hook to `git push` to also push `refs/gittuf/reference-state-log`.

adityasaky:
Local only commits are not made available to remote users. The RSL entry represents a state of a ref that others check out and consume. If branch `feature` builds off `main` with commits `A <- B <- C` but it's only pushed at `C`, `feature` isn't expected to ever have `A` or `B` as the tip. Creating the RSL entry at push time is also more efficient since at verification, if the branch goes directly to `C`, that's what you want to verify. Finally, this is also more conducive to rewriting history locally before pushing changes.

spectre10:
I think a `pre-receive` server-side hook would be a good addition to this. It can be configured to reject the changes if the rsl verification fails, or if the changes are not recorded in rsl. If this cannot be achieved with `add-hooks`, we can maybe document it somewhere. GitHub/GitLab also support these hooks and suggest similar use-cases ([GitHub](https://docs.github.com/en/enterprise-server@3.7/admin/policies/enforcing-policy-with-pre-receive-hooks/about-pre-receive-hooks#about-pre-receive-hooks)).

adityasaky:
I think focusing on client side hoods is fine for now. Moving server side gittuf implementation details to another ticket sounds good to me.

spectre10:
> An important detail is that server-side hooks are only supported on GitHub Enterprise and self-hosted/dedicated GitLab instances. Do we have access to the right infrastructure to test this? Alternatively, it could be implemented and tested using a lightweight git server such as Gitea, which does support server side hooks (https://docs.gitea.com/administration/config-cheat-sheet#security-security).

Can't we just create a local repo and use it as remote to clone from, in another folder? We can set up the server-side hooks in local (remote) repo and push from the other (cloned) folder. @datosh

[...]

PR #234, opened 2023-12-18
implement gittuf add-hooks to configure pre-push hook

datosh:
Manage hooks as described in: #158 

Added `UpdateHook` functionality. There doesn't seem to be support for managing hooks in go-git (https://github.com/src-d/go-git/issues?q=is%3Aissue+is%3Aopen+hook) 

~~Unit test currently breaks as the `conf.Core.Worktree` value is not populated when creating a repo with `PlainInit`. Any input appreciated.~~ done

Add following tests / features before merging: 
* make sure hook is executable
* do not overwrite hook script unless force

Further considerations:
* `gittuf rsl record HEAD` should not add a record when executed multiple times. We need a new issue for this.

### DIFF
--- a/internal/cmd/addhooks/addhooks.go
+++ b/internal/cmd/addhooks/addhooks.go
@@ -1,0 +1,56 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package addhooks
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/gittuf/gittuf/internal/repository"
+	"github.com/spf13/cobra"
+)
+
+type options struct {
+	force bool
+}
+
+func (o *options) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolVarP(
+		&o.force,
+		"force",
+		"f",
+		false,
+		"Overwrite hooks, if they already exist",
+	)
+}
+
+func (o *options) Run(cmd *cobra.Command, _ []string) error {
+	repo, err := repository.LoadRepository()
+	if err != nil {
+		return err
+	}
+
+	err = repo.UpdateHook(repository.HookPrePush, prePushScript, o.force)
+	var hookErr *repository.ErrHookExists
+	if errors.As(err, &hookErr) {
+		fmt.Fprintf(
+			cmd.ErrOrStderr(),
+			"'%s' already exists. Use --force flag or merge existing hook and the following script manually:\n\n%s\n",
+			string(hookErr.HookType),
+			prePushScript,
+		)
+	}
+	return err
+}
+
+func New() *cobra.Command {
+	o := &options{}
+	cmd := &cobra.Command{
+		Use:   "add-hooks",
+		Short: "Add git hooks that automatically create and sync RSL.",
+		RunE:  o.Run,
+	}
+	o.AddFlags(cmd)
+
+	return cmd
+}
--- a/internal/cmd/addhooks/addhooks.go
+++ b/internal/cmd/addhooks/addhooks.go
@@ -20,7 +20,7 @@ func (o *options) AddFlags(cmd *cobra.Command) {
 		"force",
 		"f",
 		false,
-		"Overwrite hooks, if they already exist",
+		"overwrite hooks, if they already exist",
 	)
 }
 
--- a/internal/cmd/addhooks/prepush.go
+++ b/internal/cmd/addhooks/prepush.go
@@ -1,0 +1,22 @@
+package addhooks
+
+var prePushScript = []byte(`#!/bin/sh
+set -e
+
+remote="$1"
+url="$2"
+
+if ! command -v gittuf > /dev/null
+then
+    echo "gittuf could not be found"
+    echo "Download from: https://github.com/gittuf/gittuf/releases/latest"
+    exit 1
+fi
+
+echo "Pulling RSL from ${remote}."
+gittuf rsl remote pull ${remote}
+echo "Creating new RSL record for HEAD."
+gittuf rsl record HEAD
+echo "Pushing RSL to ${remote}."
+gittuf rsl remote push ${remote}
+`)
--- a/internal/repository/hook.go
+++ b/internal/repository/hook.go
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package repository
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path"
+)
+
+type ErrHookExists struct {
+	HookType HookType
+}
+
+func (e *ErrHookExists) Error() string {
+	return fmt.Sprintf("hook '%s' already exists", e.HookType)
+}
+
+type HookType string
+
+var HookPrePush = HookType("pre-push")
+
+// UpdateHook updates a git hook in the repositorie's .git/hooks folder.
+// Existing hook files are not overwritten, unless force flag is set.
+func (r *Repository) UpdateHook(hookType HookType, content []byte, force bool) error {
+	// TODO: rely on go-git to find .git folder, once
+	// https://github.com/go-git/go-git/issues/977 is available.
+	// Note, until then gittuf does not support separate git dir.
+	tree, err := r.r.Worktree()
+	if err != nil {
+		return fmt.Errorf("reading worktree: %w", err)
+	}
+	if tree == nil {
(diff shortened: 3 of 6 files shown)


internal/repository/hook.go
```diff
@@ -0,0 +1,29 @@
...
+var HookPrePush = HookType("pre-push")
+
+func (r *Repository) UpdateHook(hookType HookType, content []byte) error {
+	conf, err := r.r.Config()
+	if err != nil {
+		return fmt.Errorf("reading repository config: %w", err)
+	}
+
+	repoRoot := conf.Core.Worktree
+	hookFile := path.Join(repoRoot, ".git", "hooks", string(hookType))
```
adityasaky, line 22: It'd be nice to identify `.git` rather than hardcode it, for cases with detached git-dirs.

internal/repository/hook.go
```diff
@@ -0,0 +1,29 @@
...
+
+	if err := os.WriteFile(hookFile, content, 0o700); err != nil {
```
adityasaky, line 24: We should at the very least abort if a pre-push hook already exists. We don't want to nuke other hooks the user may have installed.

internal/cmd/addhooks/prepush.go
```diff
@@ -0,0 +1,23 @@
...
+
+var prePushScript = []byte(`#!/bin/sh
+
+remote="$1"
+url="$2"
+
+if ! command -v gittuf > /dev/null
+then
+    echo "gittuf could not be found"
+    echo "Download from: https://github.com/gittuf/gittuf"
```
adityasaky, line 11:
```suggestion
    echo "Download from: https://github.com/gittuf/gittuf/releases/latest"
```

internal/cmd/addhooks/prepush.go
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+echo "Updating RSL."
+gittuf rsl remote pull origin
+echo "Creating new RSL record for HEAD."
+gittuf rsl record HEAD
+
+echo "Pushing refs/gittuf/reference-state-log with ${remote}"
+# --no-verify disables pre-push hook, to prevent infinite loop.
+git push --no-verify ${remote} refs/gittuf/reference-state-log
```
adityasaky, line 22: Interesting, I wonder if `gittuf rsl remote push` bypasses the hook trigger anyway. I'm reasonably certain it should.

adityasaky (commented):
Some general comments, I like where this is going!

internal/cmd/addhooks/prepush.go
datosh: I tried, and somehow `gittuf rsl remote push` does not discover my push credentials. Maybe a bug we should file? I can do more investigation tomorrow.

adityasaky: Very strange, more details would be great for this.

internal/repository/hook.go
spectre10: As a side note, hooks may not always be scripts, they can be any executable binary. So appending to existing hooks may not work as well.

adityasaky: good point! i think we check and if so, we present the parts of the hook they ought to add to the script themself. Maybe a link to docs someplace.

internal/repository/hook.go
datosh: I am not sure what a "detached git-dir" is. Could you provide me an example so I can test against the implementation?

internal/repository/hook.go
datosh: Good point. I added the necessary checks to not overwrite existing hooks. 
Do we want to provide a `--force` flag for `gittuf add-hooks` or are we good with the basic functionality for now?

internal/cmd/addhooks/prepush.go
datosh:
```
datosh@lenovo-x1:~/code/testing-gittuf/repo$ git push --no-verify origin refs/gittuf/reference-state-log 
Everything up-to-date
datosh@lenovo-x1:~/code/testing-gittuf/repo$ gittuf rsl remote push origin
Error: unable to push RSL
authentication required
Usage:
  gittuf rsl remote push <remote> [flags]

Flags:
  -h, --help   help for push

Global Flags:
      --verbose   enable verbose logging
```

`gittuf` does not seem to provide a lot of details, even with `--verbose` flag.

As I authenticated to this repository using VSCode, and I had similar issues in the past with other tools, I suspect VSCode is doing something non-standard. I will do some investigation, but don't hesitate to provide any pointers :)

datosh:
I'll fix the DCO before merging to keep the history intact.

internal/cmd/addhooks/prepush.go
adityasaky: The `--verbose` flag doesn't do much yet. We have the logging framework merged as of a week or so ago, but now we need to go in and add Debug statements.

adityasaky: gittuf * push has worked with the standard CLI invocation where I'm authenticated using ssh-agent and so. Nothing jumps out at me atm re the VSCode thing.

internal/repository/hook.go
adityasaky: `--force` is fine to add I think, with a warning printed to stderr.

internal/cmd/addhooks/prepush.go
datosh: > Interesting, I wonder if `gittuf rsl remote push` bypasses the hook trigger anyway. I'm reasonably certain it should.

Yes, it does. I have changed the hook script to only rely on `gittuf`.

> gittuf * push has worked with the standard CLI invocation where I'm authenticated using ssh-agent and so. Nothing jumps out at me atm re the VSCode thing.

It seems to be related to my particular VSCode setup. I see the same auth issue with other tools.
On the other hand, when I checkout with ssh key everything works fine. I think we can ignore that auth error in this context.

internal/repository/hook.go
datosh: oh wow, TIL! 

Do you know if there is a better approach than using `r.r.Worktree()` to eventually get to the `hooks` folder? If not then I will add more logic to test if `.git` is dir or file etc...

datosh: Created issue (https://github.com/go-git/go-git/issues/977) and added comment.

adityasaky:
Just took another quick look. I think this is largely good to go minus `--force` and a corresponding test.

internal/cmd/addhooks/prepush.go
```diff
@@ -0,0 +1,22 @@
...
+
+if ! command -v gittuf > /dev/null
+then
+    echo "gittuf could not be found"
+    echo "Download from: https://github.com/gittuf/gittuf/releases/latest"
+    exit 1
+fi
+
+echo "Pulling RSL from ${remote}."
+gittuf rsl remote pull ${remote}
```
adityasaky, line 17: We have a slight issue with branching here where we don't check if rsl update includes changes to the branch currently being pushed. I don't think we can do that strictly within the hook, maybe using a more sophisticated gittuf command.

Since the HEAD branch is not updated but the RSL is, we can't run `verify-ref` after the pull either, because that'll fail simply because the HEAD != expected value in the RSL.

datosh:
> Just took another quick look. I think this is largely good to go minus `--force` and a corresponding test.

I implemented `--force` and squashed so the DCO check is happy!

internal/cmd/addhooks/prepush.go
datosh: I see the problem you are describing, but currently we do not run any verification in the script. Is this something we should do?

adityasaky: It's fine for now, I was also thinking aloud here.

internal/cmd/addhooks/addhooks.go
```diff
@@ -0,0 +1,56 @@
...
+	force bool
+}
+
+func (o *options) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolVarP(
+		&o.force,
+		"force",
+		"f",
+		false,
+		"Overwrite hooks, if they already exist",
```
adityasaky, line 23:
```suggestion
		"overwrite hooks, if they already exist",
```

adityasaky (approved):
LGTM, aside a minor consistency nit in the help message for the command.

PR #235, opened 2023-12-18
repository: Prevent duplicate RSL entries

adityasaky:
`gittuf rsl record` would blindly create RSL entries when instructed, even if the content was identical to the latest unskipped entry for the ref in question. This PR prevents duplicate RSL entries.

### DIFF
--- a/internal/repository/rsl.go
+++ b/internal/repository/rsl.go
@@ -33,6 +33,14 @@ func (r *Repository) RecordRSLEntryForReference(refName string, signCommit bool)
 		return err
 	}
 
+	isDuplicate, err := r.isDuplicateEntry(absRefName, ref.Hash())
+	if err != nil {
+		return err
+	}
+	if isDuplicate {
+		return nil
+	}
+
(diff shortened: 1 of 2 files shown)


internal/repository/rsl.go
```diff
@@ -33,6 +33,20 @@ func (r *Repository) RecordRSLEntryForReference(refName string, signCommit bool)
...
+	if err == nil {
+		if latestUnskippedEntry.TargetID == ref.Hash() {
+			// duplicate entry, do nothing
+			return nil
+		}
+	} else if !errors.Is(err, rsl.ErrRSLEntryNotFound) {
+		// a prior entry may not exist, so that's the only allowed error
+		return err
+	}
+
```
datosh, line 49: Given that this code is added twice (even with the same comment), would it make sense to refactor it into a `Repository` private helper function instead? Something like `isLatestEntry(refName string) bool`

adityasaky: Done!

wlynch (approved):

PR #242, opened 2023-12-21
Add interfaces for authorization attestations

adityasaky:
Builds on #170, supersedes #164.


internal/repository/attestations.go
```diff
@@ -0,0 +1,154 @@
...
+// the specified parameters. The issuer of the authorization is identified using
+// their key.
+func (r *Repository) RemoveReferenceAuthorization(keyBytes []byte, targetRef, fromID, toID string, signCommit bool) error {
+	key, err := tuf.LoadKeyFromBytes(keyBytes)
+	if err != nil {
+		return err
+	}
+	if key.KeyVal.Private == "" {
+		// The authorization removal process must be authenticated using the
+		// original signing key
```
wlynch, line 108: Out of scope for this PR, but if other admins wanted an authorization to be removed because it was now considered untrustworthy, the recommended action would be to modify the repo policy to not trust the identity, right?

internal/repository/attestations.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if len(env.Signatures) == 0 {
+		if err := allAttestations.RemoveReferenceAuthorization(targetRef, fromID, toID); err != nil {
+			return err
+		}
+	} else {
+		if err := allAttestations.SetReferenceAuthorization(r.r, env, targetRef, fromID, toID); err != nil {
+			return err
+		}
+	}
```
wlynch, line 149:
```suggestion
	if len(newSignatures) == 0 {
		// No more signatures, we can remove the ReferenceAuthorization.
		if err := allAttestations.RemoveReferenceAuthorization(targetRef, fromID, toID); err != nil {
			return err
		}
	} else {
		// There are still other signatures left, so update the ReferenceAuthorization to the new state.
		env.Signatures = newSignatures
		if err := allAttestations.SetReferenceAuthorization(r.r, env, targetRef, fromID, toID); err != nil {
			return err
		}
	}
```

Small nit here for readability - this confused me on first pass because I thought this was acting on the original signature state until I realized that this was the post-filtered state.

internal/signerverifier/dsse/dsse.go
```diff
@@ -44,15 +44,30 @@ func SignEnvelope(ctx context.Context, envelope *dsse.Envelope, signer dsse.Sign
...
+			break
+		}
+	}
+
+	if signatureIndex > -1 {
+		// Replace existing signature for the key
+		envelope.Signatures[signatureIndex] = signature
+	} else {
+		envelope.Signatures = append(envelope.Signatures, signature)
+	}
```
wlynch, line 70: Why do we need this? Is it because the envelope passed in might already have signatures attached?

At minimum we should document the assumptions we have about the envelope signatures, but I also wonder if we should bother with this and leave the existing signatures 🤔 

If you do want to enforce the exactly 1 behavior of key ids <-> signatures, then this isn't sufficient since this only updates the first instance of a key id, but if the key id appears multiple times in the existing envelope the rest remain unmodified. 🤷

wlynch (commented):
Looks good! Just one thing we need to resolve.

internal/repository/attestations.go
adityasaky: Correct. Then we'd have a new policy in place for changes that come after which wouldn't trust that signature, even if it continued to exist.

internal/repository/attestations.go
adityasaky: Done.

internal/signerverifier/dsse/dsse.go
adityasaky: I've documented in the docstring the correct behaviour. I've also updated this to scan the full list of signatures, keeping signatures from other keys and appending the new sig from the current key. This removes duplicates when they already exist in an envelope, and prevents duplicates in gittuf signed metadata (assuming key ID calculation is consistent).

PR #225, opened 2023-12-16
Added add and remove root keys to cmd

neilnaveen:
- Added functionality to cmd to add and remove root keys from, which needs a key already in the root to allow adding or removing keys from the root.
- Partially fixes issue https://github.com/gittuf/gittuf/issues/117

PR #249, opened 2023-12-23
add debug statements for repository/root.go and cmd/trust/init/init.go

Dev-Voldemort:
As discussed in #54.

### DIFF
--- a/internal/cmd/addhooks/addhooks.go
+++ b/internal/cmd/addhooks/addhooks.go
@@ -5,6 +5,7 @@ package addhooks
 import (
 	"errors"
 	"fmt"
+	"log/slog"
 
 	"github.com/gittuf/gittuf/internal/repository"
 	"github.com/spf13/cobra"
@@ -31,6 +32,9 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 	}
 
 	err = repo.UpdateHook(repository.HookPrePush, prePushScript, o.force)
+	if err == nil {
+		slog.Info("Hook updated", "type", repository.HookPrePush)
+	}
 	var hookErr *repository.ErrHookExists
 	if errors.As(err, &hookErr) {
 		fmt.Fprintf(
--- a/internal/cmd/clone/clone.go
+++ b/internal/cmd/clone/clone.go
@@ -5,6 +5,7 @@ package clone
 import (
 	"github.com/gittuf/gittuf/internal/repository"
 	"github.com/spf13/cobra"
+	"log/slog"
 )
 
 type options struct {
@@ -27,6 +28,11 @@ func (o *options) Run(cmd *cobra.Command, args []string) error {
 		dir = args[1]
 	}
 	_, err := repository.Clone(cmd.Context(), args[0], dir, o.branch)
+
+	if err == nil {
+		slog.Info("Repository cloned", "repository", args[0])
+	}
+
 	return err
 }
 
--- a/internal/cmd/rsl/annotate/annotate.go
+++ b/internal/cmd/rsl/annotate/annotate.go
@@ -3,6 +3,8 @@
 package annotate
 
 import (
+	"log/slog"
+
 	"github.com/gittuf/gittuf/internal/cmd/common"
 	"github.com/gittuf/gittuf/internal/repository"
 	"github.com/spf13/cobra"
@@ -38,7 +40,13 @@ func (o *options) Run(_ *cobra.Command, args []string) error {
 		return err
 	}
 
-	return repo.RecordRSLAnnotation(args, o.skip, o.message, true)
+	err = repo.RecordRSLAnnotation(args, o.skip, o.message, true)
+	if err != nil {
+		return err
+	}
+	slog.Info("Added RSL annotation")
+
+	return nil
 }
 
 func New() *cobra.Command {
--- a/internal/repository/hook.go
+++ b/internal/repository/hook.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"log/slog"
 	"os"
 	"path"
 )
@@ -35,11 +36,13 @@ func (r *Repository) UpdateHook(hookType HookType, content []byte, force bool) e
 	if tree == nil {
 		return fmt.Errorf("worktree is nil, can't update hooks")
 	}
+	slog.Debug("Fetched repository's worktree")
 	repoRoot := tree.Filesystem.Root()
 	hookFolder := path.Join(repoRoot, ".git", "hooks")
 	if err := os.MkdirAll(hookFolder, 0o750); err != nil {
 		return fmt.Errorf("making sure folder exist: %w", err)
 	}
+	slog.Debug("Created .git/hooks directory in root")
 
 	hookFile := path.Join(hookFolder, string(hookType))
 	hookExists, err := doesFileExist(hookFile)
--- a/internal/repository/root.go
+++ b/internal/repository/root.go
@@ -7,6 +7,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"log/slog"
 
 	"github.com/gittuf/gittuf/internal/policy"
 	"github.com/gittuf/gittuf/internal/signerverifier"
@@ -21,27 +22,34 @@ func (r *Repository) InitializeRoot(ctx context.Context, rootKeyBytes []byte, si
 	if err := r.InitializeNamespaces(); err != nil {
 		return err
 	}
+	slog.Debug("Initialized gittuf namespaces")
 
 	publicKey, err := tuf.LoadKeyFromBytes(rootKeyBytes)
 	if err != nil {
 		return err
 	}
+	slog.Debug("Loaded root public key", "ID", publicKey.KeyID)
+
 	signer, err := signerverifier.NewSignerVerifierFromSecureSystemsLibFormat(rootKeyBytes)
 	if err != nil {
 		return err
 	}
+	slog.Debug("Loaded root signer")
 
 	rootMetadata := policy.InitializeRootMetadata(publicKey)
+	slog.Debug("Initialized root metadata")
 
 	env, err := dsse.CreateEnvelope(rootMetadata)
 	if err != nil {
 		return nil
 	}
+	slog.Debug("Created DSSE envelope with root metadata")
 
 	env, err = dsse.SignEnvelope(ctx, env, signer)
 	if err != nil {
 		return nil
 	}
+	slog.Debug("Signed DSSE envelope via signer verifier")
 
 	state := &policy.State{
 		RootPublicKeys: []*tuf.Key{publicKey},
@@ -50,7 +58,13 @@ func (r *Repository) InitializeRoot(ctx context.Context, rootKeyBytes []byte, si
 
 	commitMessage := "Initialize root of trust"
 
-	return state.Commit(ctx, r.r, commitMessage, signCommit)
+	err = state.Commit(ctx, r.r, commitMessage, signCommit)
+	if err != nil {
+		return err
+	}
+	slog.Debug("Committed policy state")
+
+	return nil
 }
 
 // AddRootKey is the interface for the user to add an authorized key
@@ -60,6 +74,7 @@ func (r *Repository) AddRootKey(ctx context.Context, rootKeyBytes, newrootKeyByt
 	if err != nil {
 		return err
 	}
+	slog.Debug("Loaded root signer")
 	rootKeyID, err := sv.KeyID()
 	if err != nil {
 		return err
@@ -69,6 +84,7 @@ func (r *Repository) AddRootKey(ctx context.Context, rootKeyBytes, newrootKeyByt
 	if err != nil {
 		return err
 	}
+	slog.Debug("Loaded current policy state")
 
 	rootMetadata, err := state.GetRootMetadata()
 	if err != nil {
@@ -83,6 +99,7 @@ func (r *Repository) AddRootKey(ctx context.Context, rootKeyBytes, newrootKeyByt
 	if err != nil {
 		return err
 	}
+	slog.Debug("Loaded policy key", "ID", newRootKey.KeyID)
 
 	rootMetadata = policy.AddRootKey(rootMetadata, newRootKey)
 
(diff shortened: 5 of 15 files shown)


internal/repository/root.go
```diff
@@ -60,37 +73,47 @@ func (r *Repository) AddTopLevelTargetsKey(ctx context.Context, rootKeyBytes, ta
...
 	rootMetadata, err := state.GetRootMetadata()
 	if err != nil {
 		return err
 	}
+	slog.Debug("Constructed root meta data with deserialized paylod of state's root envelope")
 
 	if !isKeyAuthorized(rootMetadata.Roles[policy.RootRoleName].KeyIDs, rootKeyID) {
 		return ErrUnauthorizedKey
 	}
+	slog.Debug("Successfully authorized root of trust with root key id")
```
spectre10, line 98: We are not authorizing the root key here. Just checking if it is already authorized or not.

internal/repository/root.go
```diff
@@ -21,27 +22,34 @@ func (r *Repository) InitializeRoot(ctx context.Context, rootKeyBytes []byte, si
 	if err := r.InitializeNamespaces(); err != nil {
 		return err
 	}
+	slog.Debug("Initialized custom git namespaces: gittuf/policy, gittuf/reference-state-log")
```
adityasaky, line 25:
```suggestion
	slog.Debug("Initialized gittuf namespaces")
```

internal/repository/root.go
```diff
@@ -21,27 +22,34 @@ func (r *Repository) InitializeRoot(ctx context.Context, rootKeyBytes []byte, si
...
 
 	publicKey, err := tuf.LoadKeyFromBytes(rootKeyBytes)
 	if err != nil {
 		return err
 	}
+	slog.Debug("Loaded public key from root key bytes")
```
adityasaky, line 31:
```suggestion
	slog.Debugf("Loaded root public key with ID '%s'", publicKey.KeyID)
```

internal/repository/root.go
```diff
@@ -21,27 +22,34 @@ func (r *Repository) InitializeRoot(ctx context.Context, rootKeyBytes []byte, si
...
+
 	signer, err := signerverifier.NewSignerVerifierFromSecureSystemsLibFormat(rootKeyBytes)
 	if err != nil {
 		return err
 	}
+	slog.Debug("Loaded new signer in custom SSL format for signing DSSE envelope")
```
adityasaky, line 37:
```suggestion
	slog.Debug("Loaded root signer")
```

internal/repository/root.go
```diff
@@ -21,27 +22,34 @@ func (r *Repository) InitializeRoot(ctx context.Context, rootKeyBytes []byte, si
...
 
 	rootMetadata := policy.InitializeRootMetadata(publicKey)
+	slog.Debug("Constructed root metadata from root's public key")
```
adityasaky, line 40:
```suggestion
	slog.Debug("Initialized root metadata")
```

internal/repository/root.go
```diff
@@ -50,7 +58,12 @@ func (r *Repository) InitializeRoot(ctx context.Context, rootKeyBytes []byte, si
 
 	commitMessage := "Initialize root of trust"
 
-	return state.Commit(ctx, r.r, commitMessage, signCommit)
+	err = state.Commit(ctx, r.r, commitMessage, signCommit)
+	if err != nil {
+		return err
+	}
+	slog.Debug("Committed the state of policy to refs/gittuf/policy")
```
adityasaky, line 65:
```suggestion
	slog.Debug("Committed policy state")
```

internal/repository/root.go
```diff
@@ -50,7 +58,12 @@ func (r *Repository) InitializeRoot(ctx context.Context, rootKeyBytes []byte, si
...
+	return nil
```
adityasaky, line 66:
```suggestion

	return nil
```

internal/repository/root.go
```diff
@@ -60,37 +73,47 @@ func (r *Repository) AddTopLevelTargetsKey(ctx context.Context, rootKeyBytes, ta
 	if err != nil {
 		return err
 	}
+	slog.Debug("Loaded new signer verifier in custom ssl format to generate key id of root of trust")
```
adityasaky, line 76:
```suggestion
	slog.Debug("Loaded root signer")
```

internal/repository/root.go
```diff
@@ -60,37 +73,47 @@ func (r *Repository) AddTopLevelTargetsKey(ctx context.Context, rootKeyBytes, ta
...
 	rootKeyID, err := sv.KeyID()
 	if err != nil {
 		return err
 	}
+	slog.Debug("Created root key id with signer verifier")
```
adityasaky, line 81:
```suggestion
```

internal/repository/root.go
```diff
@@ -60,37 +73,47 @@ func (r *Repository) AddTopLevelTargetsKey(ctx context.Context, rootKeyBytes, ta
...
 
 	state, err := policy.LoadCurrentState(ctx, r.r)
 	if err != nil {
 		return err
 	}
+	slog.Debug("Loaded current state of repository's active policy")
```
adityasaky, line 87:
```suggestion
	slog.Debug("Loaded current policy state")
```

internal/repository/root.go
```diff
@@ -60,37 +73,47 @@ func (r *Repository) AddTopLevelTargetsKey(ctx context.Context, rootKeyBytes, ta
...
 
 	rootMetadata, err := state.GetRootMetadata()
 	if err != nil {
 		return err
 	}
+	slog.Debug("Constructed root meta data with deserialized paylod of state's root envelope")
```
adityasaky, line 93:
```suggestion
```

internal/repository/root.go
adityasaky:
```suggestion
```

internal/repository/root.go
```diff
@@ -60,37 +73,47 @@ func (r *Repository) AddTopLevelTargetsKey(ctx context.Context, rootKeyBytes, ta
...
 	if !isKeyAuthorized(rootMetadata.Roles[policy.RootRoleName].KeyIDs, rootKeyID) {
 		return ErrUnauthorizedKey
 	}
+	slog.Debug("Successfully authorized root of trust with root key id")
 
 	targetsKey, err := tuf.LoadKeyFromBytes(targetsKeyBytes)
 	if err != nil {
 		return err
 	}
+	slog.Debug("Loaded pointer to the key generated from targets key bytes")
```
adityasaky, line 104:
```suggestion
	slog.Debug("Loaded policy key")
```

I'd again add in the key ID for this one as well.

General tip: the log statement does not need to contain extremely granular details about the _implementation_ (eg: pointer to the key does not matter to a gittuf user). Rather, we want to log important activities as they happen with pertinent information like the key IDs, things that users will care about.

adityasaky: I think statements can be more brief as well.

internal/repository/root.go
```diff
@@ -60,37 +73,47 @@ func (r *Repository) AddTopLevelTargetsKey(ctx context.Context, rootKeyBytes, ta
...
 
 	rootMetadata = policy.AddTargetsKey(rootMetadata, targetsKey)
+	slog.Debug("Added targets key to the root meta data as trusted public key for top level Targets role")
 
 	rootMetadata.SetVersion(rootMetadata.Version + 1)
+	slog.Debug("Updated version of root meta data")
```
adityasaky, line 110:
```suggestion
```

internal/repository/root.go
```diff
@@ -60,37 +73,47 @@ func (r *Repository) AddTopLevelTargetsKey(ctx context.Context, rootKeyBytes, ta
...
+	slog.Debug("Successfully authorized root of trust with root key id")
 
 	targetsKey, err := tuf.LoadKeyFromBytes(targetsKeyBytes)
 	if err != nil {
 		return err
 	}
+	slog.Debug("Loaded pointer to the key generated from targets key bytes")
 
 	rootMetadata = policy.AddTargetsKey(rootMetadata, targetsKey)
+	slog.Debug("Added targets key to the root meta data as trusted public key for top level Targets role")
```
adityasaky, line 107:
```suggestion
	slog.Debug("Added policy key to the gittuf root")
```

maybe again with key ID

internal/repository/root.go
adityasaky: This can't fail, no real reason to log this IMO.

internal/repository/root.go
```diff
@@ -60,37 +73,47 @@ func (r *Repository) AddTopLevelTargetsKey(ctx context.Context, rootKeyBytes, ta
...
 
 	rootMetadata.SetVersion(rootMetadata.Version + 1)
+	slog.Debug("Updated version of root meta data")
+
 	rootMetadataBytes, err := json.Marshal(rootMetadata)
 	if err != nil {
 		return err
 	}
+	slog.Debug("Constructed rootMetadataBytes with Jsonified root meta data")
```
adityasaky, line 116:
```suggestion
```

internal/repository/root.go
```diff
@@ -100,12 +123,18 @@ func (r *Repository) AddTopLevelTargetsKey(ctx context.Context, rootKeyBytes, ta
 	if err != nil {
 		return err
 	}
+	slog.Debug("Signed DSSE envelope via signer verifier")
```
adityasaky, line 126: This can and should be consistent for each function, the signing operation is the same whether it's when initializing root or updating it.

internal/repository/root.go
```diff
@@ -100,12 +123,18 @@ func (r *Repository) AddTopLevelTargetsKey(ctx context.Context, rootKeyBytes, ta
...
 	state.RootEnvelope = env
 
 	commitMessage := fmt.Sprintf("Add policy key '%s' to root", targetsKey.KeyID)
 
-	return state.Commit(ctx, r.r, commitMessage, signCommit)
+	err = state.Commit(ctx, r.r, commitMessage, signCommit)
+	if err != nil {
+		return err
+	}
+	slog.Debug("Committed the state of policy to refs/gittuf/policy")
```
adityasaky, line 136: See above.

adityasaky (changes requested):
Took a first pass at reviewing root.go in repository. I'll add more comments later. Thanks for taking a stab at this!

adityasaky:
@patzielinski could you take a look too? Thanks!

Dev-Voldemort:
Not sure if I should replace these [print](https://github.com/gittuf/gittuf/blob/c99a12d35a7dfc428b2a0dfcd9b1b0eb2b58d073/internal/cmd/rsl/remote/check/check.go#L27) statements with logs or let it be

Dev-Voldemort:
Rebased the PR and added log statements to changes made via #225.

internal/cmd/addhooks/addhooks.go
```diff
@@ -31,6 +32,9 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 	}
 
 	err = repo.UpdateHook(repository.HookPrePush, prePushScript, o.force)
+	if err == nil {
+		slog.Info("Hook updated", "type", repository.HookPrePush)
```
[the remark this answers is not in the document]
adityasaky: Info is the default level right? So this will always be printed? I wonder if that's too much noise.

internal/cmd/clone/clone.go
```diff
@@ -27,6 +28,11 @@ func (o *options) Run(cmd *cobra.Command, args []string) error {
 		dir = args[1]
 	}
 	_, err := repository.Clone(cmd.Context(), args[0], dir, o.branch)
+
+	if err == nil {
+		slog.Info("Repository cloned", "repository", args[0])
```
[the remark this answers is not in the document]
adityasaky: I'm not entirely convinced of the value of log statements in the cmd package. Not much really happens there, we've purposely kept most of the logic in the repository package.

internal/cmd/rsl/annotate/annotate.go
```diff
@@ -38,7 +40,13 @@ func (o *options) Run(_ *cobra.Command, args []string) error {
 		return err
 	}
 
-	return repo.RecordRSLAnnotation(args, o.skip, o.message, true)
+	err = repo.RecordRSLAnnotation(args, o.skip, o.message, true)
+	if err != nil {
+		return err
+	}
+	slog.Info("Added RSL annotation")
```
adityasaky, line 47: This could well be a statement in RecordRSLAnnotation and we'd need no diff here...

internal/repository/hook.go
```diff
@@ -35,11 +36,13 @@ func (r *Repository) UpdateHook(hookType HookType, content []byte, force bool) e
 	if tree == nil {
 		return fmt.Errorf("worktree is nil, can't update hooks")
 	}
+	slog.Debug("Fetched repository's worktree")
```
adityasaky, line 39: I don't know, I'm struggling to judge what information is helpful and isn't, I might be too close to the implementation? I'd like to reduce noise as much as possible so the log is valuable, and I'll admit I don't have a lot of experience with threading this needle. @wlynch could you take a look at this PR too?

internal/repository/root.go
```diff
@@ -21,27 +22,34 @@ func (r *Repository) InitializeRoot(ctx context.Context, rootKeyBytes []byte, si
...
 	if err != nil {
 		return nil
 	}
+	slog.Debug("Created DSSE envelope with root metadata")
 
 	env, err = dsse.SignEnvelope(ctx, env, signer)
 	if err != nil {
 		return nil
 	}
+	slog.Debug("Signed DSSE envelope via signer verifier")
```
adityasaky, line 52: I think this one needs an edit. "signer verifier" doesn't provide value to a user. The important bit is that the envelope was signed (and maybe logging what signer was used to sign it via the key ID). I'd go so far as to say even "DSSE" doesn't add value to the user. The user needn't care whether we as the makers of gittuf choose to store policy metadata using DSSE, COSE, etc, but just that the metadata was in fact signed. Do you see what I mean? :smile:

adityasaky (changes requested):
Added some thoughts but I'm struggling to find the balance between useful log messages and introducing details users may not care about...@wlynch may have more value on this one that I at the moment.

adityasaky:
Superseded by #302, thanks for your work on this @Dev-Voldemort!

PR #250, opened 2023-12-25
Cleaned up the policy root

naveensrinivasan:
- Cleaned up the policy root to make it handle errors
- Update the tests and updated the docs.


internal/policy/root.go
```diff
@@ -27,35 +29,49 @@ func InitializeRootMetadata(key *tuf.Key) *tuf.RootMetadata {
 	return rootMetadata
 }
 
-// AddTargetsKey adds targetsKey as a trusted public key in rootMetadata for the
+// AddTargetsKey function is used to add the 'targetsKey' as a trusted public key in 'rootMetadata' for the
 // top level Targets role.
-func AddTargetsKey(rootMetadata *tuf.RootMetadata, targetsKey *tuf.Key) *tuf.RootMetadata {
+func AddTargetsKey(rootMetadata *tuf.RootMetadata, targetsKey *tuf.Key) (*tuf.RootMetadata, error) {
+	if rootMetadata == nil {
```
adityasaky, line 35: Hm, the only consumer for this is `repository.AddTopLevelTargetsKey`, which is after the current state is loaded (during which the state is verified as well). I'm ambivalent on adding this check.

internal/policy/root.go
```diff
@@ -27,35 +29,49 @@ func InitializeRootMetadata(key *tuf.Key) *tuf.RootMetadata {
...
+		return nil, fmt.Errorf("rootMetadata is nil")
+	}
+	if targetsKey == nil {
+		return nil, fmt.Errorf("targetsKey is nil")
```
adityasaky, line 39: The same thing here, the targetsKey instance is explicitly loaded by the repository API, which aborts prior to this if the key isn't loaded correctly...

internal/policy/root.go
```diff
@@ -27,35 +29,49 @@ func InitializeRootMetadata(key *tuf.Key) *tuf.RootMetadata {
 	return rootMetadata
 }
 
-// AddTargetsKey adds targetsKey as a trusted public key in rootMetadata for the
+// AddTargetsKey function is used to add the 'targetsKey' as a trusted public key in 'rootMetadata' for the
```
adityasaky, line 32: nit: 80 cols on docstrings

internal/policy/root.go
```diff
@@ -27,35 +29,49 @@ func InitializeRootMetadata(key *tuf.Key) *tuf.RootMetadata {
...
 	rootMetadata.Roles[TargetsRoleName] = targetsRole
 
-	return rootMetadata
+	return rootMetadata, nil
 }
 
-// DeleteTargetsKey removes keyID from the list of trusted top level Targets
-// public keys in rootMetadata. It does not remove the key entry itself as it
-// does not check if other roles can be verified using the same key.
+// DeleteTargetsKey removes 'keyID' from trusted public keys for top level Targets
```
adityasaky, line 65: nit: 80 cols on docstrings

internal/policy/root.go
```diff
@@ -27,35 +29,49 @@ func InitializeRootMetadata(key *tuf.Key) *tuf.RootMetadata {
...
+// role in 'rootMetadata'. Note: It doesn't remove the key entry itself as it doesn't
+// check if other roles can use the same key.
 func DeleteTargetsKey(rootMetadata *tuf.RootMetadata, keyID string) (*tuf.RootMetadata, error) {
+	if rootMetadata == nil {
```
adityasaky, line 69: similar thoughts here. Also if we do add this check, maybe we could have named errors.

internal/policy/root.go
```diff
@@ -27,35 +29,49 @@ func InitializeRootMetadata(key *tuf.Key) *tuf.RootMetadata {
...
+		return nil, fmt.Errorf("rootMetadata is nil")
+	}
+	if keyID == "" {
```
adityasaky, line 72: Good catch here, repository just passes on the key ID string for this one.

adityasaky: But I wonder if this check actually belongs there instead. I think it's sensible to position them that way because the `cmd` package does mark this as compulsory, but if `repository` is used for an API in future, we want guardrails there.

internal/policy/root.go
naveensrinivasan: >I'm ambivalent on adding this check.

Why? This will avoid panics.

internal/policy/root.go
naveensrinivasan: The checks should be at the bottom of the stack. This avoids panics.

internal/policy/root.go
adityasaky: Agreed, it'll avoid panics on direct invocations, but my point was we don't really expect that for this package. That said, redundant checks are probably better, so let's proceed. :)

internal/policy/root.go
```diff
@@ -27,35 +29,49 @@ func InitializeRootMetadata(key *tuf.Key) *tuf.RootMetadata {
 	return rootMetadata
 }
 
-// AddTargetsKey adds targetsKey as a trusted public key in rootMetadata for the
+// AddTargetsKey function is used to add the 'targetsKey' as a trusted public key in 'rootMetadata' for the
 // top level Targets role.
-func AddTargetsKey(rootMetadata *tuf.RootMetadata, targetsKey *tuf.Key) *tuf.RootMetadata {
+func AddTargetsKey(rootMetadata *tuf.RootMetadata, targetsKey *tuf.Key) (*tuf.RootMetadata, error) {
+	if rootMetadata == nil {
+		return nil, fmt.Errorf("rootMetadata is nil")
```
adityasaky, line 36: can we use named errors instead?

internal/policy/root.go
adityasaky: same here re named errors

internal/policy/root_test.go
```diff
@@ -90,9 +99,19 @@ func TestDeleteTargetsKey(t *testing.T) {
...
 
-	rootMetadata = AddTargetsKey(rootMetadata, targetsKey1)
-	rootMetadata = AddTargetsKey(rootMetadata, targetsKey2)
+	rootMetadata, err = AddTargetsKey(rootMetadata, targetsKey1)
+	assert.Nil(t, err)
+	rootMetadata, err = AddTargetsKey(rootMetadata, targetsKey2)
+	assert.Nil(t, err)
 
+	_, err = DeleteTargetsKey(nil, targetsKey1.KeyID)
+	if err == nil {
```
adityasaky, line 108: With named errors, this and other checks can be `assert.ErrorIs`.

adityasaky (changes requested):

go.sum
```diff
@@ -1,3 +1,4 @@
+bitbucket.org/creachadair/shell v0.0.7/go.mod h1:oqtXSSvSYr4624lnnabXHaBsYW6RD80caLi2b3hJk0U=
```
adityasaky, line 1: any idea why there's a major change to go.sum here?

internal/policy/root.go
```diff
@@ -73,35 +79,49 @@ func DeleteRootKey(rootMetadata *tuf.RootMetadata, keyID string) (*tuf.RootMetad
 	return rootMetadata, nil
 }
 
-// AddTargetsKey adds targetsKey as a trusted public key in rootMetadata for the
-// top level Targets role.
-func AddTargetsKey(rootMetadata *tuf.RootMetadata, targetsKey *tuf.Key) *tuf.RootMetadata {
+// AddTargetsKey function is used to add the 'targetsKey' as a trusted public key in 'rootMetadata'
```
adityasaky, line 82: nit: 80 cols

internal/policy/root.go
adityasaky: bump, this is still over 80 cols per my editor

image

internal/policy/root.go
```diff
@@ -111,12 +131,15 @@ func DeleteTargetsKey(rootMetadata *tuf.RootMetadata, keyID string) (*tuf.RootMe
 	if len(targetsRole.KeyIDs) <= targetsRole.Threshold {
 		return nil, ErrCannotMeetThreshold
 	}
+
 	for i, k := range targetsRole.KeyIDs {
-		if k == keyID {
-			targetsRole.KeyIDs = append(targetsRole.KeyIDs[:i], targetsRole.KeyIDs[i+1:]...)
-			break
+		if k != keyID {
```
adityasaky, line 136: Could you explain the rationale here? AFAICT this isn't more efficient. Thanks!

adityasaky: Also, if we're modifying this, maybe we could do something like: https://github.com/gittuf/gittuf/pull/242/files#diff-0082f62103fcad0364437363ac14e85ec846f1f3a0bd2b3c9a379c3cc4a3fccfR131-R139. This ensures that even there are duplicate instances of a key ID, we still ensure they're removed. Maybe long term, we should implement a set for these types of fields under the hood...

internal/policy/root_test.go
```diff
@@ -89,7 +89,16 @@ func TestAddTargetsKey(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	rootMetadata = AddTargetsKey(rootMetadata, targetsKey)
+	_, err = AddTargetsKey(nil, targetsKey)
+	if err == nil {
```
adityasaky, line 93: This can be replaced by `assert.ErrorIs(t, err, ErrRootMetadataNil)`

internal/policy/root_test.go
```diff
@@ -89,7 +89,16 @@ func TestAddTargetsKey(t *testing.T) {
...
+		t.Fatal("expected error when rootMetadata is nil")
+	}
+	_, err = AddTargetsKey(rootMetadata, nil)
+	if err == nil {
+		t.Fatal("expected error when targetsKey is nil")
```
adityasaky, line 98: same here, with `ErrorIs`.

adityasaky (changes requested):
I think a final round of changes and we should be good to merge. :)

adityasaky (approved):

adityasaky:
@naveensrinivasan I took the liberty of pushing a change to your branch that addresses my outstanding comments and resolves conflicts. Thanks for working on this!

PR #251, opened 2023-12-27
Fixed the createTestStateWithDelegatedPolicies

neilnaveen:
- Fixed the createTestStateWithDelegatedPolicies, so that when it is used in conjunction with createTestRepository, and when the state is commited, it can pass the verification check.
- Currently the test will not pass, but when PR 219 is merged, it should work


adityasaky (approved):

adityasaky:
Note that tests pass because this isn't used for a verification test yet.

PR #252, opened 2024-01-02
Removed createTestRepositoryWithTargets in favor of createTestRepositoryWithPolicy

neilnaveen:
- Removed createTestRepositoryWithTargets in favor of createTestRepositoryWithPolicy, since both have similar functions.
- With the embeded key bytes in helpers_test, we do not need the return value of createTestRepositoryWithTargets
- Addresses https://github.com/gittuf/gittuf/pull/225#discussion_r1437251654

### DIFF
--- a/internal/repository/targets_test.go
+++ b/internal/repository/targets_test.go
@@ -64,6 +64,7 @@ func TestAddDelegation(t *testing.T) {
 	targetsMetadata, err := state.GetTargetsMetadata(policy.TargetsRoleName)
 	assert.Nil(t, err)
 	assert.Equal(t, 1, len(targetsMetadata.Delegations.Keys))
+	assert.Equal(t, 2, len(targetsMetadata.Delegations.Roles))
 	assert.Contains(t, targetsMetadata.Delegations.Roles, policy.AllowRule())
 
 	err = r.AddDelegation(context.Background(), targetsKeyBytes, policy.TargetsRoleName, ruleName, authorizedKeyBytes, rulePatterns, false)
--- a/internal/repository/targets_test.go
+++ b/internal/repository/targets_test.go
@@ -16,7 +16,16 @@ import (
 
 func TestInitializeTargets(t *testing.T) {
 	// The helper also runs InitializeTargets for this test
-	r := createTestRepositoryWithPolicy(t, "")
+	r, _ := createTestRepositoryWithRoot(t, "")
+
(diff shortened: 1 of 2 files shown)


internal/repository/targets_test.go
```diff
@@ -47,9 +47,15 @@ func TestAddDelegation(t *testing.T) {
...
+	gpgKey, err := gpg.LoadGPGKeyFromBytes(gpgPubKeyBytes)
+	if err != nil {
+		t.Fatal(err)
+	}
+
 	targetsMetadata, err := state.GetTargetsMetadata(policy.TargetsRoleName)
 	assert.Nil(t, err)
-	assert.Empty(t, targetsMetadata.Delegations.Keys)
+	assert.Contains(t, targetsMetadata.Delegations.Keys, gpgKey.KeyID)
+	assert.Equal(t, 1, len(targetsMetadata.Delegations.Keys))
```
adityasaky, line 58: Let's instead assert there are 2 rules, as we're adding one more in this test and we assert again it was added.

internal/repository/targets_test.go
```diff
@@ -16,7 +16,7 @@ import (
 
 func TestInitializeTargets(t *testing.T) {
 	// The helper also runs InitializeTargets for this test
-	r, _ := createTestRepositoryWithTargets(t)
```
adityasaky, line 19: This is the one place where we see a difference. when we just initialize targets, we don't want to see any rules in it except the allow rule. I suggest using createTestRepositoryWithRoot here and running r.InitializeTargets here.

adityasaky (changes requested):
Minor comments aside, this is looking good!

internal/repository/targets_test.go
neilnaveen: Fixed

internal/repository/targets_test.go
neilnaveen: I have moved the check and added an assert.Equal for the length of the keys

adityasaky: We still need to assert the len(targetsMetadata.Delegations.Roles) here so that we see it's 2, then 3 after AddDelegation.

adityasaky: You correctly do the inverse of the check for Remove below, 3 then 2.

adityasaky (changes requested):

internal/repository/targets_test.go
neilnaveen: Sorry for the late reply, I have fixed this.

adityasaky (approved):

Issue #229, opened 2023-12-18
[Bug]: Unify signing mechanisms


adityasaky:
### What happened?

gittuf currently has two different signing workflows based on what's being signed. We use the user's default git configuration to sign commits / RSL entries etc, but a custom workflow for the policy metadata (where signatures are embedded in DSSE documents). The signerverifier package along with go-securesystemslib implements this at the moment.

The issues with having these be separate are as follows:
1) metadata signing uses non standard keys right now, the custom sslib format (which is also being retired upstream)
2) rules that authorize a user by their GPG or sigstore identity (for git commit signatures) cannot trivially be converted into a delegation because those signing methods currently can't be used to sign delegated policy metadata

I propose we unify the policy signing around git signing, because we want to preserve standard git signing workflows for compatibility with existing user workflows. So, we want to get to a place where the same gpg key / ssh key / sigstore identity from a user's git config can be used to sign gittuf policy metadata. FWIW, an ssh key used with git signing will likely align with upstream securesystemslib changes that retire the custom on-disk formats (solving point 1 above). We currently don't have prior art on signing DSSE with GPG but sigstore tools do sign in-toto attestations embedded in DSSE using OIDC. I think we can plug into those for sigstore and implement GPG + DSSE ourselves with some input from other securesystemslib / DSSE users.

cc @datosh

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

JustinCappos:
This is a laudable goal.  

I'm curious what a git user will see in this case if we have multiple signatures or signatures using a key type unsupported by git.

adityasaky:
This one's come up a couple times in conversations because people are confused between the two types of keys. It also doesn't help that the metadata signing keys currently use a custom format, it's not obvious how to even create those test keys. https://github.com/secure-systems-lab/go-securesystemslib/pull/67 is a starting point upstream to fixing this. Once that's merged, that _should_ also unblock SSH git signing. Separately, we can discuss in go-securesystemslib how to sign DSSE with GPG, which should get us most of the way to unification.

PR #253, opened 2024-01-02
gitinterface: Support verifying SSH key signatures

adityasaky:
This PR adds support for verifying Git signatures issued using SSH keys (we need to evaluate the missing algorithms / key types still). This goes hand in hand with https://github.com/secure-systems-lab/go-securesystemslib/pull/67 and is related to #229.


adityasaky:
https://github.com/secure-systems-lab/go-securesystemslib/pull/67 now blocks this PR. I've added test code from that PR here that must be taken out.

I've also added ed25519 keys but there's an inconsistency in supported formats between `ssh-keygen` and securesystemslib, though I haven't dived into python-securesystemslib for this specifically yet. I've added test for commits signed using RSA and ECDSA keys generated using ssh-keygen.

adityasaky:
@wlynch I think this is ready for review. The ed25519 thing hasn't been resolved yet but if everything else looks fine, I'm okay to proceed with a ticket for that one. I think it might primarily be documentation related that talks about what formats to use / how to convert between them, with example commands.

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/go-git/go-git/v5 v5.11.0
 	github.com/in-toto/attestation v1.0.1
 	github.com/jonboulle/clockwork v0.4.0
-	github.com/secure-systems-lab/go-securesystemslib v0.8.0
+	github.com/secure-systems-lab/go-securesystemslib v0.8.1-0.20240108171218-da429971be5a
```
adityasaky, line 11: Also, this has been merged into main. dependabot will auto-update on the next release, v0.9.0.

internal/gitinterface/keys.go
```diff
@@ -257,3 +268,66 @@ func verifyGitsignSignature(ctx context.Context, key *tuf.Key, data, signature [
...
+	sshSignature, err := decodeSSHSignature(signature)
+	if err != nil {
+		return errors.Join(ErrVerifyingSSHSignature, err)
+	}
+
+	if err := publicKey.Verify(data, sshSignature); err == nil {
+		return nil
+	}
+
+	return ErrIncorrectVerificationKey
```
wlynch, line 293:
```suggestion
	if err := publicKey.Verify(data, sshSignature); err != nil {
		return errors.Join(ErrIncorrectVerificationKey, err)
	}
	return nil
```

we probably want to join the errors here too, this way we don't drop any debug information about why the verify may have failed.

wlynch (changes requested):
Looks great! One small change.

internal/gitinterface/keys.go
adityasaky: Done! Took the liberty of updating this in verifyGitsignSignature as well.

wlynch (approved):

PR #254, opened 2024-01-04
(WIP) Use attestations during verification

adityasaky:
This PR plugs in the attestations from #170 and #242 into the verification workflow.


adityasaky:
Now part of #261, closing.

PR #255, opened 2024-01-05
Sign RSL entry in evaluation mode with specified key

adityasaky:
Related: #110 

In evaluation mode, we create non-latest RSL entries. In addition, we want to also control the key used to sign the entries rather than defaulting to the key specified in Git config.

~This PR includes commits from #253.~


adityasaky:
Blocked by #257 and #258.

wlynch (approved):

PR #257, opened 2024-01-09
Add support for loading PEM encoded keys into policy

adityasaky:
Accompanies #253

### DIFF
--- a/internal/cmd/policy/addrule/addrule.go
+++ b/internal/cmd/policy/addrule/addrule.go
@@ -66,7 +66,7 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 
 	authorizedKeysBytes := [][]byte{}
 	for _, key := range o.authorizedKeys {
-		kb, err := common.ReadKeyBytes(key)
+		kb, err := common.ReadKeyBytes(key) //nolint:staticcheck
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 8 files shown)


internal/cmd/policy/addrule/addrule.go
```diff
@@ -66,7 +66,7 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 
 	authorizedKeysBytes := [][]byte{}
 	for _, key := range o.authorizedKeys {
-		kb, err := common.ReadKeyBytes(key)
+		kb, err := common.ReadKeyBytes(key) //nolint:staticcheck
```
wlynch, line 69: Should we move these to the non-deprecated func in this pr? 🤔

wlynch (approved):

internal/cmd/policy/addrule/addrule.go
adityasaky:
```
// Deprecated: With key format standardization, we can use tuf.Key for public
// keys. Use LoadKey to get a tuf.Key object directly. This deprecation can only
// be complete with API changes in repository to handle a signer object for the
// private key bits. Then, we'll have retired passing around key bytes
// altogether.
```

So, I wanted to handle passing a) tuf.Key objects, and b) signer objects to `repository.*` together, so that we simultaneously drop passing around `[]byte` representation of private / public keys. If you think it's better to proceed with this right away, I can make the change strictly for the public keys passed to `repository` here.

adityasaky: To add on, I figured if we were going to perform surgery on `repository` endpoints, we could do both in the same pass...:)

PR #258, opened 2024-01-09
Fix SSH signature verification

adityasaky:
I discovered a bug in the SSH signature verification introduced in #253. This fixes it!

~Draft until #257 is merged.~

I didn't dive into the specific issue entirely (time constraints), but my thinking is it had to do with how we were decoding SSH signature fields and passing it to crypto/ssh for verification. Switching to the sshsig library fixed it right away, and signature verification compatibility aligns with `git verify-commit` now when we specify trusted ssh keys using `gpg.ssh.allowedsigners`. I would like to dive into the specific bug in the previous implementation in the near future.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/ProtonMail/go-crypto v0.0.0-20230923063757-afb1ddc0824c
 	github.com/go-git/go-billy/v5 v5.5.0
 	github.com/go-git/go-git/v5 v5.11.0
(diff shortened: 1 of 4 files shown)


adityasaky:
I've taken #257 out as this didn't strictly depend on that. It was stacked to help test the fix locally.

wlynch (approved):
Can you add more details on what the issue was in the PR description and how this fixes it? Thanks! 🙏

PR #259, opened 2024-01-09
Detect cycles in delegations DFS, make "root" invalid policy file name

adityasaky:
Detect if a policy role was already expanded during DFS.

Also added a check to ensure a delegated role or rule isn't called "root". Most of the diff in the tests is in moving an existing test into a t.Run block to add another checking for the invalid name.


JustinCappos:
Glad to see this added!

wlynch (approved):

Issue #230, opened 2023-12-18
Roadmap: Dogfood gittuf


adityasaky:
As noted in the gittuf roadmap, we must start using gittuf to protect the gittuf source repository. To do so, we need to consider the following things:

1) Policies: what policies do we enable for the repository? Given that this is an open source project with external contributors, I suggest we start with a simple branch protection rule on `main` that requires a maintainer to have merged some change.
1) Verification points: we ought to run gittuf verification in CI using the GitHub action that provisions gittuf. We also want to distribute this verification so I propose we gittuf maintainers run verification whenever we pull in the latest changes to the repository. This will get simpler with #220 and #158.
1) RSL entry creation: we don't want to move away from GitHub's PRs for user contributions. For a single policy protecting `main` we only really need RSL entries for that branch. As a start, we could move _only merges_ off the GitHub web UI so RSL entries are signed by the maintainers, but this will not scale. Along the way, I think we'll be able to flesh out how to align RSL entry creation with the GitHub UI, perhaps using a CI job that authenticates using Sigstore or signs an entry after authenticating the action using the GitHub API.

neilnaveen:
Another basic rule could be a tag protection rule, which would stops tag teleportation attacks.

adityasaky:
Yes, protecting release tags is another check we should add.

adityasaky:
Major features we need added before we can start doing this:
* Support for policy-staging for multiple signers on root.json / targets.json
* Support for threshold verification / attestations

adityasaky:
Current blockers:
* #308 
* #261 
* #260

adityasaky:
We now have a document detailing how we envision dogfooding to go: https://github.com/gittuf/gittuf/blob/main/docs/dogfood.md.

We've started the phase 1 of dogfooding via GitHub Action workflows. I'm going to keep this issue open to track us moving to higher levels!

PR #260, opened 2024-01-09
Introduce Basic Staging Capabilities Utilizing the Policy Staging Ref

neilnaveen:
**How does the policy staging currently work** 

1. All changes to the policy are first made in the policy staging ref. These changes only impact the repo once they are merged into the policy ref, utilizing a fast-forward merge.
2. Before merging anything from the policy staging ref into the policy ref, the latest state of the policy staging ref is inspected and checked to see if it is a valid state. This guarantees that the project will always work with a good policy state since the policy ref leapfrogs from a valid state to a valid state. 
3. This is especially useful when attestations are added into gittuf, meaning the state would have to be invalid at specific points before being valid. 
4. For example, let's say that we want to have three root key holders in our root of trust from day one. We also want to set the threshold to three. Without policy staging, if one key is added, but the threshold is 3, this would be impossible to do. But if this was in the policy-staging ref, these changes could be pushed up the upstream repo, fetched by other key holders, and added to the root of trust. Finally, the changes can be merged into the main when all three keys are added.

**Possible features for policy staging** 

1. Having only one policy staging reference is limiting since they can be overwritten, so instead of having a single policy staging reference, you could potentially create multiple feature policy "branch" references.

**Problems that are created by policy staging**

1. We cannot assume that the first RSL entry is the first valid state since it most likely would be a policy-staging RSL entry, which could mean it could be invalid. That logic will have to be changed.
2. Currently, if you only have one policy staging branch, this would mean that whenever you pull the latest RSL from upstream, your WIP policy changes would be overwritten.


neilnaveen:
listr-ules is being modified only for the purpose of seeing the difference between the policy staging ref and the policy ref.

internal/policy/policy.go
```diff
@@ -715,6 +713,65 @@ func ListRules(ctx context.Context, repo *git.Repository) ([]*DelegationWithDept
...
+	}
+	policyCommit, err := gitinterface.GetCommit(repo, policyRef.Hash())
+	if err == nil {
+		// This check ensures that the policy staging branch is a direct forward progression of the policy branch,
+		// preventing any overwrites of policy history and maintaining a linear policy evolution, since a
+		// fast forward merge does not work with a non-linear history.
+
+		// This is only being checked if there are no problems finding the tip of the policy ref, since if there
+		// is no tip, then it cannot be an ancestor of the tip of the policy staging ref
+		isAncestor, err := policyCommit.IsAncestor(policyStagingCommit)
```
adityasaky, line 753: gitinterface.KnowsCommit does this for us IIRC

internal/policy/policy.go
```diff
@@ -500,9 +498,9 @@ func (s *State) Verify(ctx context.Context) error {
 	return targetsVerifier.Verify(ctx, nil, s.TargetsEnvelope)
 }
 
-// Commit verifies and writes the State to the policy namespace. It also creates
-// an RSL entry recording the new tip of the policy namespace.
-func (s *State) Commit(ctx context.Context, repo *git.Repository, commitMessage string, signCommit bool) error {
+// Commit verifies and writes the State to the specified namespace. It also creates
+// an RSL entry recording the new tip of the specified namespace.
+func (s *State) Commit(ctx context.Context, repo *git.Repository, commitMessage string, signCommit bool, targetRef string) error {
```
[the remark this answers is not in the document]
adityasaky: The question of workflow would then extend to the repository package which is where policy is actually updated.

adityasaky (commented):
Added some thoughts. This might be an issue that warrants a synchronous discussion of workflows. cc @JustinCappos @wlynch @reza-curtmola

internal/policy/policy.go

internal/policy/policy.go
```diff
@@ -715,6 +713,65 @@ func ListRules(ctx context.Context, repo *git.Repository) ([]*DelegationWithDept
...
+		return fmt.Errorf("state verification failed: %w", err)
+	}
+
+	// Get the reference for the PolicyRef
+	policyRef, err := repo.Reference(plumbing.ReferenceName(PolicyRef), true)
+	if err != nil {
+		return fmt.Errorf("failed to get policy reference %s: %w", PolicyRef, err)
+	}
+
+	// Get the reference for then PolicyStagingRef
```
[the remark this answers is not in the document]

internal/policy/policy.go
```diff
@@ -715,6 +713,65 @@ func ListRules(ctx context.Context, repo *git.Repository) ([]*DelegationWithDept
 	return allDelegations, nil
 }
 
+func Apply(ctx context.Context, repo *git.Repository, signRSLEntry bool) error {
+	state, err := LoadCurrentState(ctx, repo, PolicyStagingRef)
+	if err != nil {
+		return fmt.Errorf("failed to load current state: %w", err)
+	}
+
+	if err := state.Verify(ctx); err != nil {
```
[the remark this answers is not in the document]

internal/policy/policy.go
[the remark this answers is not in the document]
neilnaveen: I have fixed this. So state.Commit will never be used in a way to change the policy ref. Does this mean that the policy ref is almost useless, other than the fact that it's RSL entries are only on valid states that where applied?

neilnaveen:
Rebased the code with main 👀 @adityasaky

internal/policy/policy.go
```diff
@@ -855,6 +853,61 @@ func (s *State) getRootVerifier() (*Verifier, error) {
...
+	}
+
+	// Check if the PolicyStagingRef is ahead of PolicyRef (fast-forward)
+
+	policyStagingCommit, err := gitinterface.GetCommit(repo, policyStagingRef.Hash())
+	if err != nil {
+		// if there is no tip for the policy staging ref, this means that no change will be made to the policy ref
+		return fmt.Errorf("failed to get policy staging tip commit: %w", err)
+	}
+	policyCommit, err := gitinterface.GetCommit(repo, policyRef.Hash())
```
adityasaky, line 881:
```suggestion

	policyCommit, err := gitinterface.GetCommit(repo, policyRef.Hash())
```

internal/policy/policy.go
```diff
@@ -855,6 +853,61 @@ func (s *State) getRootVerifier() (*Verifier, error) {
...
+		// fast forward merge does not work with a non-linear history.
+
+		// This is only being checked if there are no problems finding the tip of the policy ref, since if there
+		// is no tip, then it cannot be an ancestor of the tip of the policy staging ref
+		isAncestor, err := gitinterface.KnowsCommit(repo, policyStagingCommit.Hash, policyCommit)
+		if err != nil {
+			return fmt.Errorf("failed to check if policy commit is ancestor of policy staging commit: %w", err)
+		}
+		if !isAncestor {
+			return fmt.Errorf("cannot apply changes since policy is not an ancestor of the policy staging ")
```
adityasaky, line 894: can we use a named error here?

internal/policy/policy_test.go
```diff
@@ -530,3 +532,45 @@ func TestStateHasFileRule(t *testing.T) {
 		assert.False(t, hasFileRule)
 	})
 }
+
+func TestApply(t *testing.T) {
```
adityasaky, line 536: I'd like to see more cases, split up with `t.Run` that also inspects the policy and policy-staging refs post apply.

internal/policy/verify.go
```diff
@@ -169,7 +169,7 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
 			slog.Debug(fmt.Sprintf("Verifying entry '%s'...", entry.ID.String()))
 
 			slog.Debug("Checking if entry is for policy reference...")
-			if entry.RefName == PolicyRef {
+			if entry.RefName == PolicyRef || entry.RefName == PolicyStagingRef {
```
adityasaky, line 172: No, we should never be loading the policy staging ref during verification.

adityasaky: staging should be special cased so we skip to the next one.

internal/repository/policy.go
```diff
@@ -24,7 +24,7 @@ var (
 // update to the RSL.
 func (r *Repository) PushPolicy(ctx context.Context, remoteName string) error {
```
adityasaky, line 25: We should think about the interface for Push so we know what ref is being pushed. Perhaps we don't want to push both everytime...

internal/repository/sync_test.go
```diff
@@ -85,7 +88,6 @@ func TestClone(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-
```
adityasaky, line 88: nit: unrelated diff

internal/repository/sync_test.go
```diff
@@ -200,7 +202,7 @@ func TestClone(t *testing.T) {
 		defer os.Chdir(currentDir) //nolint:errcheck
 
 		dirName := "myRepo"
-		if err := os.Mkdir(dirName, 0755); err != nil {
+		if err := os.Mkdir(dirName, 0o755); err != nil {
```
adityasaky, line 205: nit: unrelated diff

internal/repository/targets.go
```diff
@@ -9,6 +9,7 @@ import (
 	"log/slog"
 
 	"github.com/gittuf/gittuf/internal/policy"
+
```
adityasaky, line 12: not sure why there's a diff here, gofmt / goimports should take care of this?

internal/repository/targets_test.go
```diff
@@ -83,8 +82,7 @@ func TestAddDelegation(t *testing.T) {
 		ruleName := "test-rule"
 		authorizedKeyBytes := []*tuf.Key{targetsPubKey}
 		rulePatterns := []string{"git:branch=main"}
-
```
adityasaky, line 86: nit: newline removed

internal/repository/verify_test.go
```diff
@@ -28,6 +28,10 @@ func TestVerifyRef(t *testing.T) {
 	entryID := common.CreateTestRSLReferenceEntryCommit(t, repo.r, entry, gpgKeyBytes)
 	entry.ID = entryID
 
+	if err := policy.Apply(context.Background(), repo.r, false); err != nil {
```
adityasaky, line 31: shouldn't this happen in the createTestRepositoryWithPolicy helper?

adityasaky: I see it there as well.

adityasaky (changes requested):
This is looking pretty good. Some thoughts in addition to my inline comments:
1) could you update the description with the new workflow for updating gittuf policy? I think we need to discuss this as a group, @JustinCappos, @wlynch.
2) we need to think about how this relates to policy state self-verification, especially with the changes in #308.

internal/rsl/rsl.go
```diff
@@ -27,6 +27,12 @@ const (
 	EntryIDKey                 = "entryID"
 	SkipKey                    = "skip"
 
+	// PolicyRef defines the Git namespace used for gittuf policies.
+	PolicyRef = "refs/gittuf/policy"
```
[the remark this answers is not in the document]
neilnaveen: Fixed

internal/policy/policy.go
```diff
@@ -611,21 +609,21 @@ func (s *State) Commit(ctx context.Context, repo *git.Repository, commitMessage
 		return err
 	}
 
-	ref, err := repo.Reference(plumbing.ReferenceName(PolicyRef), true)
+	ref, err := repo.Reference(plumbing.ReferenceName(PolicyStagingRef), true)
```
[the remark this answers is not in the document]
neilnaveen: I thought that since all state changes should be routed through the policy staging namespace, there would be no need to commit changes to the policy namespace.

internal/policy/policy.go
```diff
@@ -735,7 +733,7 @@ func (s *State) loadRuleNames() error {
 // pre order traversal of the delegation tree, with the depth of each
 // delegation.
 func ListRules(ctx context.Context, repo *git.Repository) ([]*DelegationWithDepth, error) {
-	state, err := LoadCurrentState(ctx, repo)
+	state, err := LoadCurrentState(ctx, repo, PolicyRef)
```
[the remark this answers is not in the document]
neilnaveen: I have changed it to a required string parameter, not the whole ref name, but the end name, like `gittuf list rules -targetRef policy`

internal/policy/policy.go
```diff
@@ -855,6 +853,61 @@ func (s *State) getRootVerifier() (*Verifier, error) {
 	}, nil
 }
 
+func Apply(ctx context.Context, repo *git.Repository, signRSLEntry bool) error {
```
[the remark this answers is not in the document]
neilnaveen: Added

internal/policy/policy.go
```diff
@@ -855,6 +853,61 @@ func (s *State) getRootVerifier() (*Verifier, error) {
...
+	_, err := LoadCurrentState(ctx, repo, PolicyStagingRef)
```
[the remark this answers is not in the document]
neilnaveen: I was using LoadCurrentState since it also verifies the state of the repo, which means that if this fails, apply should not be run, since verification failed on the policy staging ref

internal/policy/policy.go
neilnaveen: fixed

internal/policy/policy.go

internal/policy/policy_test.go
neilnaveen: I have inspected the policy and policy-staging refs, but I do not know how to create an invalid state. How should I do so?

internal/policy/verify.go

internal/repository/policy.go
neilnaveen: Should this be a later PR, or should this be incorporated into this one as a different function in cmd/rsl

internal/repository/sync_test.go
neilnaveen: Fixed

internal/repository/sync_test.go
neilnaveen: Fixed

internal/repository/verify_test.go
neilnaveen: Removed

internal/policy/policy.go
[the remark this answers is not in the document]
adityasaky: I think it depends on the workflow we collectively decide on, so you're right. I'd like to hear from other maintainers here.

internal/policy/policy.go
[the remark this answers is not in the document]
adityasaky: I see. That's a good call, let's keep this thread open, especially in the context of #308.

internal/policy/verify.go
```diff
@@ -169,6 +169,9 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
 			slog.Debug(fmt.Sprintf("Verifying entry '%s'...", entry.ID.String()))
 
 			slog.Debug("Checking if entry is for policy reference...")
+			if entry.RefName == PolicyStagingRef {
```
adityasaky, line 172: Move this check above the debug statement and add a debug message saying checking if it's for policy-staging?

internal/repository/helpers_test.go
```diff
@@ -97,6 +101,9 @@ func createTestRepositoryWithPolicy(t *testing.T, location string) *Repository {
 	if err := r.AddDelegation(testCtx, targetsSigner, policy.TargetsRoleName, "protect-main", []*tuf.Key{gpgKey}, []string{"git:refs/heads/main"}, false); err != nil {
 		t.Fatal(err)
 	}
+	if err := policy.Apply(testCtx, r.r, false); err != nil {
```
adityasaky, line 104:
```suggestion

	if err := policy.Apply(testCtx, r.r, false); err != nil {
```

internal/repository/policy.go
adityasaky: This is another what-should-the-workflow-be question, so let's see.

internal/policy/policy.go
```diff
@@ -855,6 +854,67 @@ func (s *State) getRootVerifier() (*Verifier, error) {
...
+		return fmt.Errorf("failed to get policy reference %s: %w", PolicyRef, err)
+	}
+
+	// Get the reference for the PolicyStagingRef
+	policyStagingRef, err := repo.Reference(plumbing.ReferenceName(PolicyStagingRef), true)
+	if err != nil {
+		return fmt.Errorf("failed to get policy staging reference %s: %w", PolicyStagingRef, err)
+	}
+
+	// Check if the PolicyStagingRef is ahead of PolicyRef (fast-forward)
```
adityasaky, line 880: To fail early, we should also be performing this check when loading the staging area as we may be mutating the policy...

internal/rsl/rsl.go
```diff
@@ -27,6 +27,9 @@ const (
 	EntryIDKey                 = "entryID"
 	SkipKey                    = "skip"
 
+	// PolicyStagingRef defines the Git namespace used as a staging area when creating or updating gittuf policies.
+	PolicyStagingRef = "refs/gittuf/policy-staging"
```
adityasaky, line 31: why is this necessary in the rsl package?

neilnaveen: Removed

internal/repository/helpers_test.go
neilnaveen: Fixed

internal/policy/verify.go
neilnaveen: Fixed

internal/cmd/policy/listrules/listrules.go
```diff
@@ -10,17 +10,27 @@ import (
...
+}
 
 func (o *options) Run(cmd *cobra.Command, _ []string) error {
 	repo, err := repository.LoadRepository()
 	if err != nil {
 		return err
 	}
 
-	rules, err := repo.ListRules(cmd.Context())
+	rules, err := repo.ListRules(cmd.Context(), "refs/gittuf/"+o.targetRef)
```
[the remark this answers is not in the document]
neilnaveen: I have moved this to repository/policy.

JustinCappos:
FYI: I think we should also think about having the ability to stage and merge changes to different policy files independently.

internal/policy/policy.go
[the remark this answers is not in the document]
neilnaveen: I agree that the current version of policy staging is quite painful to use, especially with multiple contributors creating changes to the policy state. 

I was thinking about addressing these issues, and https://github.com/gittuf/gittuf/pull/260#issuecomment-1962047511,  in a follow-up PR since if we can merge this simple, bare-bones version of policy-staging, making it just work(albeit in a clunky manner), other problems can start to be fixed, like those listed here https://github.com/gittuf/gittuf/issues/230#issuecomment-1961448955.

neilnaveen:
@adityasaky @JustinCappos Does each RSL entry only change one policy file at a time, since if not, staging policy files seperatly would be impossible, since we could not stage a commit that changes both file `a` and `b` if we only want commits that change file `a`

internal/policy/policy.go
```diff
@@ -143,13 +142,116 @@ func LoadState(ctx context.Context, repo *git.Repository, entry *rsl.ReferenceEn
...
 		return nil, err
 	}
 
-	return LoadState(ctx, repo, latestEntry)
+	return LoadStateForEntry(ctx, repo, entry)
+}
+
+// LoadStateForEntry returns the State for a specified RSL reference entry for
+// the policy namespace.
+func LoadStateForEntry(ctx context.Context, repo *git.Repository, entry *rsl.ReferenceEntry) (*State, error) {
```
adityasaky, line 156: I think this is an incorrect rebase after #308? That one makes this a private function.

internal/cmd/policy/listrules/listrules.go
```diff
@@ -10,17 +10,27 @@ import (
...
 
-func (o *options) AddFlags(_ *cobra.Command) {}
+func (o *options) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(
+		&o.targetRef,
+		"targetRef",
+		"",
+		"to specify which ref should be inspected",
+	)
+	cmd.MarkFlagRequired("targetRef") //nolint:errcheck
```
adityasaky, line 24: does this need to be required? Maybe make the default the main policy ref?

internal/policy/policy.go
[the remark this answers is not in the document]
adityasaky: I think treating staging refs as any other branch is fine. The only thing I want us to think through is what happens when merging multiple policy changes. If we have a good story there, we should be fine.

internal/policy/verify.go
```diff
@@ -192,6 +192,10 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
 
 			slog.Debug(fmt.Sprintf("Verifying entry '%s'...", entry.ID.String()))
 
+			slog.Debug("Checking if entry is for policy staging reference...")
+			if entry.RefName == PolicyStagingRef {
+				continue
+			}
```
adityasaky, line 198:
```suggestion
			}

```

internal/repository/policy.go
```diff
@@ -24,7 +25,7 @@ var (
...
 	slog.Debug(fmt.Sprintf("Pushing policy and RSL references to %s...", remoteName))
-	if err := gitinterface.Push(ctx, r.r, remoteName, []string{policy.PolicyRef, rsl.Ref}); err != nil {
+	if err := gitinterface.Push(ctx, r.r, remoteName, []string{policy.PolicyRef, policy.PolicyStagingRef, rsl.Ref}); err != nil {
```
adityasaky, line 28: We should also check what the impact of PolicyRef being up to date here is. It shouldn't assume the other refs are updated. I suspect the behavior is right but double checking / testing would be good.

internal/cmd/policy/listrules/listrules.go
neilnaveen: Fixed

internal/policy/verify.go
neilnaveen: Fixed

internal/repository/policy.go
neilnaveen: I have added a test

internal/policy/policy.go

docs/cli/gittuf.md
```diff
@@ -17,6 +17,7 @@ A security layer for Git repositories, powered by TUF
 * [gittuf add-hooks](gittuf_add-hooks.md)	 - Add git hooks that automatically create and sync RSL
 * [gittuf clone](gittuf_clone.md)	 - Clone repository and its gittuf references
 * [gittuf dev](gittuf_dev.md)	 - Developer mode commands
+* [gittuf gittuf](gittuf_gittuf.md)	 - applies work in progress changes to the policy state to the current policy state
```
adityasaky, line 20: Looks like a mistake but this command is plumbed through as gittuf gittuf apply?

neilnaveen: I have fixed this, but that is what `make generate`. After these changes if I run `make generate`, the changes are reversed and the `gittuf gittuf` command is added to the docs

neilnaveen: Docs are failing, since generate creates a different set of docs with the `gittuf gittuf` command

neilnaveen:
I have rolled back the changes to the simple version of policy staging

adityasaky:
@patzielinski: I tried out the policy-staging for serial signing. Can you build off this branch and kick the tyres some more? I primarily stayed on the happy path FWIW.

@neilnaveen , @wlynch: I've added some helper commands such as the ability to just sign the root metadata as well as bump the root role's threshold. Neil, can you take a look at my commits? Billy, I think the whole PR could use a review by you.

Also, this functionality is still marked developer-mode only, which I think is fine to keep as is for now.

internal/cmd/policy/listrules/listrules.go
adityasaky: FWIW, I still see this in the latest commit. It possibly snuck back in? :)

neilnaveen:
I think the commits look good, and having the functionality under developer mode also seems to be a good idea, since should not be a regularly used feature at the moment

internal/cmd/policy/listrules/listrules.go
neilnaveen: Fixed

internal/policy/policy.go
```diff
@@ -71,11 +72,9 @@ func InitializeNamespace(repo *git.Repository) error {
 		}
 	}
 
-	// Disable PolicyStagingRef until it is actually used
-	// https://github.com/gittuf/gittuf/issues/45
-	// if err := repo.Storer.SetReference(plumbing.NewHashReference(plumbing.ReferenceName(PolicyStagingRef), plumbing.ZeroHash)); err != nil {
-	// 	return err
-	// }
+	if err := repo.Storer.SetReference(plumbing.NewHashReference(plumbing.ReferenceName(PolicyStagingRef), plumbing.ZeroHash)); err != nil {
```
wlynch, line 75: nit: you can inline this in the loop above:

```go
for _, name := range []string{PolicyRef, PolicyStagingRef} {
	refname := plumbing.ReferenceName(name)
	if ref, err := repo.Reference(refname, true); err != nil {
			return err
		}
	} else if !ref.Hash().IsZero() {
		return ErrPolicyExists
	}
	if err := repo.Storer.SetReference(plumbing.NewHashReference(refname, plumbing.ZeroHash)); err != nil {
		return err
	}
}
return nil
```

wlynch (commented):
I think this needs to be rebased - there are a bunch of new commands included here that look to be unrelated to this change.

Only looking at internal, this LGTM. I'm not 100% confident in my ability to grok all the subtleties here though, so will defer to @adityasaky for final confirmation.

adityasaky:
> there are a bunch of new commands included here that look to be unrelated to this change.

yep! I added some commands to allow users to add their extra signatures, which is one of the primary motivations to have a staging area (add root.json with all root keys, sign serially, then apply). This did muck up the PR a bit but it allowed for IRL testing of the staging bits.

adityasaky:
@wlynch could you take a look at https://github.com/gittuf/gittuf/pull/260/commits/1c6c2a5088682cc443c35aebfdde28d846e13550 and https://github.com/gittuf/gittuf/pull/260/commits/93da1fc46e185efec141cd198bb8b41903ca8809? I noticed that root threshold wasn't being verified on apply and realized we needed to handle a couple more edge cases with apply. In my playground, I'm seeing expected behavior both in success and failure cases.

wlynch (approved):

adityasaky (approved):

Issue #45, opened 2023-05-18
Enable the use of `policy-staging` in root signing


adityasaky:
#37 added basic root signing capabilities. There's currently a `policy-staging` namespace created for gittuf that is unused. Ideally, the root signing workflow (`gittuf trust init`) should allow for using this staging area for all root key holders to add their keys and then add their signatures.

neilnaveen:
I would like to work on this

PR #261, opened 2024-01-10
Add support for setting and verifying threshold for rules / delegations

adityasaky:
This PR adds support for specifying thresholds in gittuf rules. For full threshold support, we need #260 / #45 merged / resolved, so that the top level targets role can meet the threshold in the refs/gittuf/policy.

As a start, though, this can be used to specify thresholds in rules that are expected to be met using RSL entry signatures / commit signatures / attestations (#254).

### DIFF
--- a/internal/policy/root.go
+++ b/internal/policy/root.go
@@ -10,9 +10,10 @@ import (
 )
 
 var (
-	ErrCannotMeetThreshold = errors.New("removing key will drop authorized keys below threshold")
+	ErrCannotMeetThreshold = errors.New("insufficient keys to meet threshold")
 	ErrRootMetadataNil     = errors.New("rootMetadata is nil")
 	ErrRootKeyNil          = errors.New("root key not found")
+	ErrTargetsMetadataNil  = errors.New("targetsMetadata not found")
 	ErrTargetsKeyNil       = errors.New("targetsKey is nil")
 	ErrKeyIDEmpty          = errors.New("keyID is empty")
 )
@@ -141,6 +142,23 @@ func DeleteTargetsKey(rootMetadata *tuf.RootMetadata, keyID string) (*tuf.RootMe
 	}
 	targetsRole.KeyIDs = newKeyIDs
 
+	rootMetadata.Roles[TargetsRoleName] = targetsRole
+
+	return rootMetadata, nil
+}
+
+// UpdateTargetsThreshold sets the threshold for the top level Targets role.
+func UpdateTargetsThreshold(rootMetadata *tuf.RootMetadata, threshold int) (*tuf.RootMetadata, error) {
+	targetsRole, ok := rootMetadata.Roles[TargetsRoleName]
+	if !ok {
+		return nil, ErrTargetsMetadataNil
+	}
+
+	if len(targetsRole.KeyIDs) < threshold {
+		return nil, ErrCannotMeetThreshold
+	}
+
+	targetsRole.Threshold = threshold
 	rootMetadata.Roles[TargetsRoleName] = targetsRole
 
 	return rootMetadata, nil
--- a/internal/repository/root.go
+++ b/internal/repository/root.go
@@ -68,36 +68,13 @@ func (r *Repository) AddRootKey(ctx context.Context, signer sslibdsse.SignerVeri
 		return err
 	}
 
-	slog.Debug("Loading current root metadata...")
-	rootMetadata, err := state.GetRootMetadata()
+	rootMetadata, err := r.loadRootMetadata(state, rootKeyID)
 	if err != nil {
 		return err
-	}
-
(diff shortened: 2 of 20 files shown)


internal/policy/root.go
```diff
@@ -121,3 +121,21 @@ func DeleteTargetsKey(rootMetadata *tuf.RootMetadata, keyID string) (*tuf.RootMe
...
+func UpdateTargetsThreshold(rootMetadata *tuf.RootMetadata, threshold int) (*tuf.RootMetadata, error) {
+	if _, ok := rootMetadata.Roles[TargetsRoleName]; !ok {
+		return rootMetadata, nil
+	}
+
+	targetsRole := rootMetadata.Roles[TargetsRoleName]
+
+	if len(targetsRole.KeyIDs) < threshold {
+		return nil, ErrCannotMeetThreshold
+	}
```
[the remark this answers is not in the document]
adityasaky: The code here isn't aware of policy-staging. A lot of these checks will have to change as that is introduced.

internal/policy/root.go
```diff
@@ -144,3 +144,21 @@ func DeleteTargetsKey(rootMetadata *tuf.RootMetadata, keyID string) (*tuf.RootMe
 
 	return rootMetadata, nil
 }
+
+// UpdateTargetsThreshold sets the threshold for the top level Targets role.
+func UpdateTargetsThreshold(rootMetadata *tuf.RootMetadata, threshold int) (*tuf.RootMetadata, error) {
+	if _, ok := rootMetadata.Roles[TargetsRoleName]; !ok {
+		return rootMetadata, nil
```
wlynch, line 151: if the target isn't present, should we be returning error here?

internal/policy/root.go
```diff
@@ -144,3 +144,21 @@ func DeleteTargetsKey(rootMetadata *tuf.RootMetadata, keyID string) (*tuf.RootMe
...
+	}
+
+	targetsRole := rootMetadata.Roles[TargetsRoleName]
```
wlynch, line 154: nit
```suggestion
	targetsRole, ok := rootMetadata.Roles[TargetsRoleName]
	if !ok {
		return rootMetadata, nil
	}
```

internal/repository/root.go
```diff
@@ -286,3 +286,57 @@ func (r *Repository) RemoveTopLevelTargetsKey(ctx context.Context, signer sslibd
...
+	env.Signatures = []sslibdsse.Signature{}
+	env.Payload = base64.StdEncoding.EncodeToString(rootMetadataBytes)
+
+	slog.Debug(fmt.Sprintf("Signing updated root metadata using '%s'...", rootKeyID))
+	env, err = dsse.SignEnvelope(ctx, env, signer)
+	if err != nil {
+		return err
+	}
+
+	state.RootEnvelope = env
```
wlynch, line 336: Optional if you want to make changes in this PR, but an observation - we're repeating similar code in this file to load/save state, which is signal that we should look to hoist this out, even if it's as simple as:

```go
rootMetadata, err := r.LoadRootMetadata(ctx, sv)
if err != nil {
  ...
}

// make whatever change
...

return r.SaveRootMetadata(ctx, sv, metadata)
```

internal/repository/targets.go
```diff
@@ -73,7 +73,7 @@ func (r *Repository) InitializeTargets(ctx context.Context, signer sslibdsse.Sig
 
 // AddDelegation is the interface for the user to add a new rule to gittuf
 // policy.
-func (r *Repository) AddDelegation(ctx context.Context, signer sslibdsse.SignerVerifier, targetsRoleName string, ruleName string, authorizedKeys []*tuf.Key, rulePatterns []string, signCommit bool) error {
+func (r *Repository) AddDelegation(ctx context.Context, signer sslibdsse.SignerVerifier, targetsRoleName string, ruleName string, authorizedKeys []*tuf.Key, rulePatterns []string, threshold int, signCommit bool) error {
```
wlynch, line 76: This list is starting to get long 😵 

We should start thinking about when variadic options with sensible defaults may make more sense ([example](https://pkg.go.dev/github.com/sigstore/sigstore/pkg/signature#SignOption)). 🤔

wlynch (commented):
Looks good! Some minor comments.

internal/repository/root.go
adityasaky: Took a pass at this!

internal/repository/targets.go
adityasaky: This makes sense but I think we should tackle this separately.

wlynch (commented):
lgtm, but needs a rebase! 🙏

adityasaky:
I'll rebase this after #304 and #308 land as they're related to what's causing a conflict here. Thanks for the review!

adityasaky:
@wlynch I've rebased and I think this is good to go!

wlynch (approved):

PR #263, opened 2024-01-11
gitinterface: Support generic PEM header

adityasaky:
@patzielinski and I ran into this with keys created using pyca/cryptography.


wlynch (approved):

adityasaky:
Thanks!

PR #264, opened 2024-01-11
Retire use of raw bytes for private / public keys

adityasaky:
This PR retires passing around keys encoded as bytes, deduplicating code that had to handle different key encodings etc.


adityasaky:
As of now, the PR retires public key bytes. Next up is using the signer interface for the private keys used to sign gittuf metadata.

adityasaky:
I'm testing a build off this with new keys in the demo etc from the CLI, but I think this is ready for a review. FWIW, it may be easier to go commit by commit. Apologies for the mammoth PR.

PR #265, opened 2024-01-12
Generate CLI docs.

wlynch:
Adds `DisableAutoGenTag: true` to disable timestamp in generated docs.


internal/cmd/addhooks/addhooks.go
```diff
@@ -46,9 +46,10 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 func New() *cobra.Command {
 	o := &options{}
 	cmd := &cobra.Command{
-		Use:   "add-hooks",
-		Short: "Add git hooks that automatically create and sync RSL.",
-		RunE:  o.Run,
+		Use:               "add-hooks",
+		Short:             "Add git hooks that automatically create and sync RSL.",
+		RunE:              o.Run,
+		DisableAutoGenTag: true,
```
adityasaky, line 52: I couldn't find a quick way to apply this to all child commands from root, looks like there's a stale open PR for that...we'll have to keep tabs on this as we add new commands.

adityasaky (approved):
LGTM!

docs/cli/main.go
```diff
@@ -0,0 +1,34 @@
+//go:generate go run .
```
adityasaky, line 1: Also, should we add the SPDX header here?

PR #268, opened 2024-01-13
policy: Apply file rules only if they exist

adityasaky:
This is a first step at making verifyEntry more efficient. We now only check for file protection rules if the policy state has at least one in any targets role.

Also, this commit sets constants for the Git reference and file rule schemes.


adityasaky:
Blocked by #264.

wlynch (approved):

PR #269, opened 2024-01-13
Setup development mode and mark attestations dev mode only

adityasaky:
This moves eval mode functions to a separate development mode (cc @wlynch).

It also marks attestation support dev mode only until we nail down some usability flows (cc @JustinCappos and @wlynch).

Finally, the new dev subcommand is also used for the previous eval mode `gittuf rsl record` command that allowed the user to create an rsl entry for a specific ID / with a specific key (cc @patzielinski). Note that even in dev mode, the environment variable must _also_ be set to actually use any function.

This PR is basically good to go but will conflict with #264.


adityasaky:
Snuck in two commits to generalize the RSL entry creation / remove some checks that significantly slow down the creation of evaluation RSL entries. As this is only used in eval (hidden behind a separate command + environment variable), removing the checks seems fine to me.

@patzielinski, this _should_ fix the issue with creation of entries for tags.

adityasaky:
Rebased against #264 as we need all of these right now. So this is blocked by #264 and #268.

wlynch (approved):

PR #270, opened 2024-01-15
Fix bugs in root public key tracking

adityasaky:
Fixes minor bugs in root key management subcommands. Also adds gittuf policy sign to add a signature to existing policy file.

Blocked by #264, #268, and #269.

### DIFF
--- a/internal/cmd/policy/sign/sign.go
+++ b/internal/cmd/policy/sign/sign.go
@@ -1,0 +1,60 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package sign
+
+import (
+	"os"
+
+	"github.com/gittuf/gittuf/internal/cmd/common"
+	"github.com/gittuf/gittuf/internal/cmd/policy/persistent"
+	"github.com/gittuf/gittuf/internal/policy"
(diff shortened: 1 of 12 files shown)


internal/cmd/policy/sign/sign.go
```diff
@@ -0,0 +1,60 @@
...
+	p          *persistent.Options
+	policyName string
+}
+
+func (o *options) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(
+		&o.policyName,
+		"policy-name",
+		policy.TargetsRoleName,
+		"policy file to sign",
```
wlynch, line 25: (non-blocking): When I see "file" I think this should be a file path, but based on the default this is really a target name? We may want to be more intentional about naming, or better define what we mean by "policy file" 🤔

wlynch (approved):

internal/cmd/policy/sign/sign.go
adityasaky: I'll address all instances of this in a separate PR after this one, thanks for pointing it out.

PR #272, opened 2024-01-16
Added functionality for key revocation in rules, without having to remove the whole rule

neilnaveen:
- Currently to remove a key from a rule, the whole rule will have to be removed, and then recreated without that key
- The `addkey` command docs have been changed, since multiple rules can be added at once, which was not reflected in the docs

### DIFF
--- a/internal/policy/targets.go
+++ b/internal/policy/targets.go
@@ -11,7 +11,10 @@ import (
 
 const AllowRuleName = "gittuf-allow-rule"
 
-var ErrCannotManipulateAllowRule = errors.New("cannot change in-built gittuf-allow-rule")
+var (
+	ErrCannotManipulateAllowRule          = errors.New("cannot change in-built gittuf-allow-rule")
+	ErrCannotRemoveKeyUsedByMultipleRules = errors.New("cannot remove key, used by multiple rules")
+)
 
 // InitializeTargetsMetadata creates a new instance of TargetsMetadata.
 func InitializeTargetsMetadata() *tuf.TargetsMetadata {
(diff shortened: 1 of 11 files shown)


neilnaveen:
👀 @adityasaky

adityasaky:
> Currently to remove a key from a rule, the whole rule will have to be removed, and then recreated without that key

Hm, IIRC the add workflow is also the update workflow, but I may be mistaken. Either way, this is probably something we want to add soon. I'll take a look, thanks!

internal/policy/targets.go
```diff
@@ -101,6 +101,16 @@ func AddKeyToTargets(targetsMetadata *tuf.TargetsMetadata, authorizedKeys []*tuf
 	return targetsMetadata, nil
 }
 
+func RemoveKeysFromTargets(targetsMetadata *tuf.TargetsMetadata, authorizedKeys []*tuf.Key) (*tuf.TargetsMetadata, error) {
+	for _, key := range authorizedKeys {
+		if err := targetsMetadata.Delegations.RemoveKey(key.KeyID); err != nil {
```
adityasaky, line 106: The reason we don't currently do this when we remove a rule is because we need to check another rule doesn't require a key. I recommend having guard rails to stop a user from removing a key that's required by a rule.

internal/tuf/tuf.go
```diff
@@ -184,6 +183,16 @@ func (d *Delegations) AddKey(key *Key) {
 	d.Keys[key.KeyID] = key
 }
 
+func (d *Delegations) RemoveKey(keyID string) error {
+	if _, ok := d.Keys[keyID]; !ok {
+		return fmt.Errorf("key %s, is not in this rule", keyID)
```
adityasaky, line 188: Not sure this needs to error if the key isn't in the file.

internal/tuf/tuf_test.go
```diff
@@ -126,12 +126,13 @@ func TestTargetsMetadataAndDelegations(t *testing.T) {
 
 	delegations := &Delegations{}
 
-	t.Run("test AddKey", func(t *testing.T) {
+	t.Run("test AddKey and RemoveKey", func(t *testing.T) {
 		assert.Nil(t, delegations.Keys)
 		delegations.AddKey(key)
 		assert.Equal(t, key, delegations.Keys[key.KeyID])
+		err := delegations.RemoveKey(key.KeyID)
+		assert.ErrorIs(t, err, nil)
```
adityasaky, line 134: assert also that delegations.Keys doesn't contain the key?

adityasaky (commented):
Also see: https://github.com/gittuf/gittuf/issues/289

internal/tuf/tuf_test.go
neilnaveen: Fixed

internal/tuf/tuf.go
neilnaveen: Fixed

internal/policy/targets.go
neilnaveen: There is now a check to see if the key is being used by 2 or more rules, including the rules the key is being removed from. Is this what you were thinking about.

neilnaveen: How should key revocation work if a key is used in multiple rules?

neilnaveen:
Closing this PR, since the functionality already exist

PR #273, opened 2024-01-16
docs: Add logo, update year in license, add get started guide

adityasaky:
~Draft until I also add the get started guide.~

~LMK if the logo in the readme is too much.~ Now at 25% width.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright 2023 gittuf a Series of LF Projects, LLC
+Copyright gittuf a Series of LF Projects, LLC
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


adityasaky:
great, thanks!

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright 2023 gittuf a Series of LF Projects, LLC
```
wlynch, line 1: My understanding is we don't need to continuously bump this, and LF recommendation is that we don't need the year at all https://www.linuxfoundation.org/blog/blog/copyright-notices-in-open-source-software-projects

adityasaky: Interesting. I'll take the year out altogether then. cc @hythloda

adityasaky: Done.

wlynch (approved):

PR #282, opened 2024-01-17
Added gittuf rsl log to log the rsl with additional information

neilnaveen:
- Added gittuf rsl log to log the rsl with additional information
- Changed list-rules to accept a RSL entry to list rules at the RSL entries state

Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/internal/cmd/rsl/log/log.go
+++ b/internal/cmd/rsl/log/log.go
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package log
+
+import (
+	"os"
+
+	"github.com/gittuf/gittuf/internal/display"
+	"github.com/gittuf/gittuf/internal/repository"
+	"github.com/spf13/cobra"
+)
+
+type options struct {
+	page     bool
+	filePath string
+}
+
+func (o *options) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(
+		&o.page,
+		"page",
+		true,
+		"page log using system's default PAGER, only enabled if displaying to stdout",
+	)
+
+	cmd.Flags().StringVar(
+		&o.filePath,
+		"file",
+		"",
+		"write log to file at specified path",
+	)
+}
+
+func (o *options) Run(_ *cobra.Command, _ []string) error {
+	repo, err := repository.LoadRepository()
+	if err != nil {
+		return err
+	}
+
+	entries, annotationMap, err := repository.GetRSLEntryLog(repo)
+	if err != nil {
+		return err
+	}
+
+	output := os.Stdout
+	if o.filePath != "" {
+		output, err = os.Create(o.filePath)
+		if err != nil {
+			return err
+		}
+		o.page = false // override page since we're not writing to stdout
+	}
+
+	outputContents := display.PrepareRSLLogOutput(entries, annotationMap)
+	writer := display.NewDisplayWriter(output, o.page)
+
+	_, err = writer.Write([]byte(outputContents))
+	return err
+}
+
+func New() *cobra.Command {
+	o := &options{}
+	cmd := &cobra.Command{
+		Use:               "log",
+		Short:             "Display the Reference State Log",
+		RunE:              o.Run,
+		DisableAutoGenTag: true,
+	}
+	o.AddFlags(cmd)
+
+	return cmd
+}
--- a/internal/display/display.go
+++ b/internal/display/display.go
@@ -1,0 +1,83 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package display
+
+import (
+	"io"
+	"os"
+	"os/exec"
+)
+
+var pagerPrograms = []string{os.Getenv("PAGER"), "less", "more"}
+
+func NewDisplayWriter(defaultOutput io.Writer, page bool) io.WriteCloser {
+	if page {
+		for _, bin := range pagerPrograms {
+			if _, err := exec.LookPath(bin); err != nil {
+				continue
+			}
+			return pageWriter(bin, defaultOutput)
+		}
+	}
+
+	switch output := defaultOutput.(type) {
+	case io.WriteCloser:
+		return output
+	default:
+		return &noopwriter{writer: defaultOutput}
+	}
+}
+
+func pageWriter(pagerBin string, defaultOutput io.Writer) *pager {
+	cmd := exec.Command(pagerBin)
+	cmd.Stdout = defaultOutput
+	cmd.Stderr = os.Stderr
+
+	return &pager{command: cmd}
+}
+
+type pager struct {
+	command *exec.Cmd
+}
+
+func (p *pager) Write(contents []byte) (int, error) {
+	stdInWriter, err := p.command.StdinPipe()
+	if err != nil {
+		return -1, err
+	}
+
+	if err := p.command.Start(); err != nil {
+		return -1, err
+	}
+
+	written, writeErr := stdInWriter.Write(contents)
+	if writeErr != nil {
+		return -1, writeErr
+	}
+
+	if err := stdInWriter.Close(); err != nil {
+		return -1, err
+	}
+
+	if err := p.command.Wait(); err != nil {
+		return -1, err
+	}
+
+	return written, nil
+}
+
+func (p *pager) Close() error {
+	return nil
+}
+
+type noopwriter struct {
+	writer io.Writer
+}
+
+func (n *noopwriter) Write(contents []byte) (int, error) {
+	return n.writer.Write(contents)
+}
+
+func (n *noopwriter) Close() error {
+	return nil
+}
--- a/internal/display/rsl.go
+++ b/internal/display/rsl.go
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package display
+
+import (
+	"fmt"
+
+	"github.com/gittuf/gittuf/internal/rsl"
+	"github.com/go-git/go-git/v5/plumbing"
+)
+
+// PrepareRSLLogOutput takes the RSL, and returns a string representation of it,
+// with annotations attached to entries
+/* Output format:
+entry <entryID> (skipped)
+
+  Ref:    <refName>
+  Target: <targetID>
+
+    Annotation ID: <annotationID>
+    Skip:          <yes/no>
+    Message:
+      <message>
+
+    Annotation ID: <annotationID>
+    Skip:          <yes/no>
+    Message:
+      <message>
+*/
+func PrepareRSLLogOutput(entries []*rsl.ReferenceEntry, annotationMap map[plumbing.Hash][]*rsl.AnnotationEntry) string {
+	log := ""
+
+	for _, entry := range entries {
+		log += fmt.Sprintf("entry %v", entry.ID)
+
+		skipped := false
+		if annotations, ok := annotationMap[entry.ID]; ok {
+			for _, annotation := range annotations {
+				if annotation.Skip {
+					skipped = true
+					break
+				}
+			}
+		}
+
+		if skipped {
+			log += " (skipped)"
+		}
+		log += "\n"
+
+		log += fmt.Sprintf("\n  Ref:    %s", entry.RefName)
+		log += fmt.Sprintf("\n  Target: %s", entry.TargetID.String())
+
+		if annotations, ok := annotationMap[entry.ID]; ok {
+			for _, annotation := range annotations {
+				log += "\n"
+				log += fmt.Sprintf("\n    Annotation ID: %s", annotation.ID.String())
+				if annotation.Skip {
+					log += "\n    Skip:          yes"
+				} else {
+					log += "\n    Skip:          no"
+				}
+				log += fmt.Sprintf("\n    Message:\n      %s", annotation.Message)
+			}
+		}
+
+		log += "\n\n"
+	}
+
+	return log[:len(log)-1]
+}
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -681,6 +681,11 @@ func (s *State) GetTargetsMetadata(roleName string) (*tuf.TargetsMetadata, error
 		}
 		e = env
 	}
+
+	if e == nil {
+		return nil, ErrMetadataNotFound
+	}
+
 	payloadBytes, err := e.DecodeB64Payload()
 	if err != nil {
 		return nil, err
@@ -764,6 +769,10 @@ func ListRules(ctx context.Context, repo *git.Repository, targetRef string) ([]*
 	state, err := LoadCurrentState(ctx, repo, targetRef)
 	if err != nil {
 		return nil, err
+	}
+
+	if !state.HasTargetsRole(TargetsRoleName) {
+		return nil, nil
 	}
 
 	topLevelTargetsMetadata, err := state.GetTargetsMetadata(TargetsRoleName)
--- a/internal/repository/rsl.go
+++ b/internal/repository/rsl.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"slices"
 
 	"github.com/gittuf/gittuf/internal/dev"
 	"github.com/gittuf/gittuf/internal/gitinterface"
@@ -204,4 +205,26 @@ func (r *Repository) isDuplicateEntry(refName string, targetID plumbing.Hash) (b
 	}
 
 	return latestUnskippedEntry.TargetID == targetID, nil
+}
+
+// GetRSLEntryLog gives us a list of all the rsl entries, and a map with a key being
+// a reference entry, and the value being an array of all applicable annotations for that reference entry
+func GetRSLEntryLog(repo *Repository) ([]*rsl.ReferenceEntry, map[plumbing.Hash][]*rsl.AnnotationEntry, error) {
+	firstEntry, _, err := rsl.GetFirstEntry(repo.r)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	lastEntry, err := rsl.GetLatestEntry(repo.r)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	entries, annotationMap, err := rsl.GetReferenceEntriesInRange(repo.r, firstEntry.GetID(), lastEntry.GetID())
+	if err != nil {
+		return nil, nil, err
+	}
+
+	slices.Reverse(entries)
+	return entries, annotationMap, nil
 }
--- a/internal/repository/rsl_test.go
+++ b/internal/repository/rsl_test.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
 	"testing"
 
 	"github.com/gittuf/gittuf/internal/dev"
@@ -560,4 +561,35 @@ func TestPullRSL(t *testing.T) {
 		err = localRepo.PullRSL(context.Background(), remoteName)
 		assert.ErrorIs(t, err, ErrPullingRSL)
 	})
+}
+
+func TestGetRSLEntryLog(t *testing.T) {
+	r := createTestRepositoryWithPolicy(t, "")
+
+	mainRef := "refs/heads/main"
+	if err := r.r.Storer.SetReference(plumbing.NewHashReference(plumbing.ReferenceName(mainRef), plumbing.ZeroHash)); err != nil {
+		t.Fatal(err)
+	}
+
+	entries, annotationMap, err := GetRSLEntryLog(r)
+	assert.Nil(t, err)
+
+	firstEntry, _, err := rsl.GetFirstEntry(r.r)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	lastEntry, err := rsl.GetLatestEntry(r.r)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expected, _, err := rsl.GetReferenceEntriesInRange(r.r, firstEntry.GetID(), lastEntry.GetID())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	slices.Reverse(expected)
+	assert.Equal(t, expected, entries)
+	assert.Equal(t, map[plumbing.Hash][]*rsl.AnnotationEntry{}, annotationMap)
 }
--- a/docs/cli/gittuf_rsl.md
+++ b/docs/cli/gittuf_rsl.md
@@ -21,6 +21,7 @@ Tools to manage the repository's reference state log
 
 * [gittuf](gittuf.md)	 - A security layer for Git repositories, powered by TUF
 * [gittuf rsl annotate](gittuf_rsl_annotate.md)	 - Annotate prior RSL entries
+* [gittuf rsl log](gittuf_rsl_log.md)	 - Display the Reference State Log
 * [gittuf rsl record](gittuf_rsl_record.md)	 - Record latest state of a Git reference in the RSL
 * [gittuf rsl remote](gittuf_rsl_remote.md)	 - Tools for managing remote RSLs
 
--- a/docs/cli/gittuf_rsl_log.md
+++ b/docs/cli/gittuf_rsl_log.md
@@ -1,0 +1,29 @@
+## gittuf rsl log
+
+Display the Reference State Log
+
+```
+gittuf rsl log [flags]
+```
+
+### Options
+
+```
+      --file string   write log to file at specified path
+  -h, --help          help for log
+      --page          page log using system's default PAGER, only enabled if displaying to stdout (default true)
+```
+
+### Options inherited from parent commands
+
+```
+      --profile                      enable CPU and memory profiling
+      --profile-CPU-file string      file to store CPU profile (default "cpu.prof")
+      --profile-memory-file string   file to store memory profile (default "memory.prof")
+      --verbose                      enable verbose logging
+```
+
+### SEE ALSO
+
+* [gittuf rsl](gittuf_rsl.md)	 - Tools to manage the repository's reference state log
+
--- a/internal/cmd/rsl/rsl.go
+++ b/internal/cmd/rsl/rsl.go
@@ -4,6 +4,7 @@ package rsl
 
 import (
 	"github.com/gittuf/gittuf/internal/cmd/rsl/annotate"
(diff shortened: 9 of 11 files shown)


internal/cmd/rsl/log/log.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	log := ""
+
+	for index, entry := range entries {
+		log += fmt.Sprintf("Entry %v's Hash = %s:\n", index, entry.ID)
+		if annotations, ok := annotationMap[entry.ID]; ok {
+			for _, annotation := range annotations {
+				log += fmt.Sprintf("    Message:\n        %s\n", annotation.Message)
+				if annotation.Skip {
+					log += "        This RSL entry is skipped\n"
```
adityasaky, line 36: We should talk about how this is formatted some more.

internal/repository/rsl.go
```diff
@@ -187,3 +188,31 @@ func (r *Repository) isDuplicateEntry(refName string, targetID plumbing.Hash) (b
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	lastEntry, _, err := rsl.GetLatestReferenceEntryForRef(repo.r, policy.PolicyRef)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	entries, annotationMap, err := rsl.GetReferenceEntriesInRangeForRef(repo.r, firstEntry.ID, lastEntry.ID, policy.PolicyRef)
```
adityasaky, line 203: Why is this logging only the policy entries?

adityasaky, line 203: The order of entries should be reversed, the log should place the most recent entries at the top, if we're aligning with git log's behavior.

internal/repository/rsl.go
```diff
@@ -187,3 +188,31 @@ func (r *Repository) isDuplicateEntry(refName string, targetID plumbing.Hash) (b
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	rules := [][]*policy.DelegationWithDepth{}
+
+	for _, entry := range entries {
```
adityasaky, line 209: I could go either way on showing the policy state. We could instead add a parameter to `list-rules` to specify the state that is obtained from `rsl log` instead of trying to do everything everywhere. Basically, let's start simpler.

internal/cmd/rsl/log/log.go
```diff
@@ -0,0 +1,83 @@
...
+		return err
+	}
+	entries, annotationMap, rules, err := repository.LogRSLEntries(cmd.Context(), repo)
+	if err != nil {
+		return err
+	}
+	log := ""
+
+	for index, entry := range entries {
+		log += fmt.Sprintf("Entry %v's Hash = %s:\n", index, entry.ID)
```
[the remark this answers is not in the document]
neilnaveen: Index has to stay now that we are iterating through the RSL entries in reverse, so we are not using a for range loop but a three component for loop

internal/cmd/rsl/log/log.go
```diff
@@ -0,0 +1,83 @@
...
+			}
+		}
+
+		log += "    State:\n"
+		for _, rule := range rules[index] {
+			log += fmt.Sprintf("%s Rule %s\n", strings.Repeat("    ", 2+rule.Depth), rule.Delegation.Name)
+		}
+		log += "\n"
+	}
+	tmpfile, err := os.CreateTemp("", "log")
```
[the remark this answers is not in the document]
neilnaveen: I have changed it from a tempfile to just pushing in the log string into the command stdin

internal/policy/policy.go
```diff
@@ -636,6 +637,11 @@ func (s *State) GetTargetsMetadata(roleName string) (*tuf.TargetsMetadata, error
 		}
 		e = env
 	}
+
+	if e == nil {
+		return nil, ErrNoTargetMetadata
```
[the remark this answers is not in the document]

internal/repository/rsl.go
neilnaveen: I have remove outputting the current rule state for each RSL entry

neilnaveen: I can make another PR to add a flag to list-rules to list rules for all RSL entries

internal/repository/rsl.go

adityasaky:
@patzielinski can you take a look as well?

neilnaveen:
I do not know why the docs check is failing. I have tried to regenerate them, but I get no changes to the docs?

internal/cmd/rsl/log/log.go
```diff
@@ -0,0 +1,67 @@
...
+		fmt.Println("Error:", err.Error())
+	}
+	return nil
+}
+
+func New() *cobra.Command {
+	o := &options{}
+	cmd := &cobra.Command{
+		Use:               "log",
+		Short:             "Log out RSL entries",
```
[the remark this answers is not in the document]
neilnaveen: Fixed

neilnaveen:
@patzielinski I have changed it so that it looks a little more like git log, but since their is not a lot that is actually being outputted, their is not a huge difference in the code. The error you where getting was because the loop iterating through the RSL entries, started at the end of the RSL entry array, but instead of decrementing  its index, it was incrementing it like a regular loop. This is fixed.

internal/cmd/rsl/log/log.go
```diff
@@ -0,0 +1,90 @@
...
+
+		log += "\n"
+	}
+
+	cmd := exec.Command("echo", log)
+	pager := exec.Command("less")
+
+	pager.Stdin, _ = cmd.StdoutPipe()
+	pager.Stdout = os.Stdout
+	pager.Stderr = os.Stderr
```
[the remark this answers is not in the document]
adityasaky: One more thing, you're going to have to read PAGER to understand what to use (PAGER=less is how git knows to use less). But exec.Command requires a const (IIRC) for the command itself, so we'll have to think how to make it possible.

internal/cmd/rsl/log/log.go
```diff
@@ -0,0 +1,90 @@
...
+		return fmt.Errorf("error: %s", err.Error())
+	}
+	return nil
+}
+
+func New() *cobra.Command {
+	o := &options{}
+	cmd := &cobra.Command{
+		Use:               "log",
+		Short:             "Display the RSL entry log",
```
adityasaky, line 83: I think this doesn't line up with the generated docs. Also since the L in RSL is log, I wonder if more tweaking is in order.

internal/policy/policy.go
```diff
@@ -668,7 +673,9 @@ func ListRules(ctx context.Context, repo *git.Repository) ([]*DelegationWithDept
 	}
 
 	topLevelTargetsMetadata, err := state.GetTargetsMetadata(TargetsRoleName)
-	if err != nil {
+	if errors.Is(err, ErrMetadataNotFound) {
```
adityasaky, line 676: I have a slight preference for:

```
if err != nil {
    if errors.Is(...) {
        return nil, nil
    }
    return nil, err
}

File: internal/policy/policy_test.go [hunk 11]
```diff
@@ -415,7 +415,6 @@ func TestStateFindDelegationEntry(t *testing.T) {
 func TestListRules(t *testing.T) {
 	t.Run("no delegations", func(t *testing.T) {
 		repo, _ := createTestRepository(t, createTestStateWithPolicy)
-
```
adityasaky, line 418: Can you add this line back?

internal/repository/rsl.go
```diff
@@ -187,3 +187,22 @@ func (r *Repository) isDuplicateEntry(refName string, targetID plumbing.Hash) (b
 
 	return latestUnskippedEntry.TargetID == targetID, nil
 }
+
+func LogRSLEntries(repo *Repository) ([]*rsl.ReferenceEntry, map[plumbing.Hash][]*rsl.AnnotationEntry, error) {
```
adityasaky, line 191: This function name sounds like you want to log RSL entries. I'd rename to something like `GetRSLEntryLog` or so.

adityasaky: Also docstring.

internal/rsl/rsl.go
```diff
@@ -362,6 +362,40 @@ func GetLatestNonGittufReferenceEntry(repo *git.Repository) (*ReferenceEntry, []
 	return targetEntry, annotations, nil
 }
 
+// GetLatestReferenceEntry returns the first reference entry
+func GetLatestReferenceEntry(repo *git.Repository) (*ReferenceEntry, []*AnnotationEntry, error) {
```
adityasaky, line 366: Why is `GetLatestEntry` not enough?

internal/cmd/rsl/log/log.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return err
+	}
+	entries, annotationMap, err := repository.LogRSLEntries(repo)
+	if err != nil {
+		return err
+	}
+	log := ""
+
+	for index := len(entries) - 1; index >= 0; index-- {
```
adityasaky, line 29: I don't like having this much logic here, I suggest reversing the list in the repository package instead.

internal/cmd/rsl/log/log.go
```diff
@@ -0,0 +1,90 @@
...
+		if annotations, ok := annotationMap[entry.ID]; ok {
+			for _, annotation := range annotations {
+				log += fmt.Sprintf("    Message:\n        %s\n", annotation.Message)
+			}
+		}
+
+		log += "\n"
+	}
+
+	cmd := exec.Command("echo", log)
```
adityasaky, line 60: I'm still leaning towards creating some kind of `internal/display` package to handle these more involved interactions with stdout.

adityasaky (changes requested):
@neilnaveen here are my latest set of comments. I think landing this will be great for UX!

internal/cmd/rsl/log/log.go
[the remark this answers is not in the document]
neilnaveen: I have fixed this, and changed pager to be what the user has set, and defaulting to less. Is this what you meant?

internal/cmd/rsl/log/log.go
```diff
@@ -0,0 +1,90 @@
...
+	pager.Stdin, _ = cmd.StdoutPipe()
+	pager.Stdout = os.Stdout
+	pager.Stderr = os.Stderr
+
+	if err := pager.Start(); err != nil {
+		return fmt.Errorf("error: %s", err.Error())
+	}
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("error: %s", err.Error())
+	}
```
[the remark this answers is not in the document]
neilnaveen: Fixed

internal/cmd/rsl/log/log.go
neilnaveen: I have changed it to "Displays the Reference State Log"

internal/policy/policy.go

internal/policy/policy_test.go

internal/repository/rsl.go

internal/rsl/rsl.go
neilnaveen: Since GetLatestEntry does not fetch the latest RSL reference entry, just any entry, reference or annotation.

internal/cmd/rsl/log/log.go
neilnaveen: I have moved this to repository/rsl. I have also added a new function in common to reverse slices of any type

internal/cmd/rsl/log/log.go
neilnaveen: Should this be incorporated in this PR, or a later one?

internal/rsl/rsl.go
adityasaky: Does this only display reference entries?

internal/cmd/rsl/log/log.go
adityasaky: I'd probably do it here for this one. Separately, a future PR could move logic from cmd for list-rules.

neilnaveen: For this, are you saying that the logic should be in internal/display, but the calling should still be in cmd, correct?

adityasaky: yes

internal/rsl/rsl.go
neilnaveen: Currently the idea is that it only outputs reference entries, and added applicable annotations to the entry itself, kind of like this 

entry , hash

annotation(if exist)

entry, hash ...

internal/cmd/rsl/log/log.go
neilnaveen: Fixed

neilnaveen:
@patzielinski Does this need testing?

internal/display/rsl.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/gittuf/gittuf/internal/repository"
+)
+
+func ShowRSL() error {
+	repo, err := repository.LoadRepository()
+	if err != nil {
+		return err
+	}
+	entries, annotationMap, err := repository.GetRSLEntryLog(repo)
```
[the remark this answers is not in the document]
neilnaveen: Fixed

internal/display/rsl.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/gittuf/gittuf/internal/repository"
+)
+
+func ShowRSL() error {
```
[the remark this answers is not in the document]
neilnaveen: Fixed

internal/display/rsl.go
```diff
@@ -0,0 +1,68 @@
+package display
```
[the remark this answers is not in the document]
neilnaveen: Fixed

neilnaveen:
The Docs check is now passing

internal/common/common.go
```diff
@@ -303,3 +303,12 @@ func CreateTestSignedTag(t *testing.T, repo *git.Repository, tagName string, tar
 
 	return tagHash
 }
+
+// ReverseSlice reverses a slice of any type using generics.
+func ReverseSlice[T any](slice []T) []T {
```
adityasaky, line 308: can we just use https://pkg.go.dev/slices#Reverse?

internal/display/rsl.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/gittuf/gittuf/internal/repository"
+)
+
+// ShowRSL displays the RSL using the pager defined by PAGER
+func ShowRSL() error {
```
adityasaky, line 15: IMO this can be more generic, it accepts something that needs to be printed, a destination, and maybe a bool to indicate if it should be paged.

adityasaky: Note that this allows for better testing too, you can pass in some arbitrary tempfile instead of os.Stdout in the test.

internal/repository/rsl_test.go
```diff
@@ -561,3 +562,36 @@ func TestPullRSL(t *testing.T) {
...
 }
+
+func TestGetRSLEntryLog(t *testing.T) {
+	r := createTestRepositoryWithPolicy(t, "")
+
+	mainRef := "refs/heads/main"
+	if err := r.r.Storer.SetReference(plumbing.NewHashReference(plumbing.ReferenceName(mainRef), plumbing.ZeroHash)); err != nil {
+		t.Fatal(err)
+	}
+	entries, annotationMap, err := GetRSLEntryLog(r)
```
adityasaky, line 573:
```suggestion
	
	entries, annotationMap, err := GetRSLEntryLog(r)
```

internal/repository/rsl_test.go
```diff
@@ -561,3 +562,36 @@ func TestPullRSL(t *testing.T) {
...
+
```
adityasaky, line 574:
```suggestion
```

internal/rsl/rsl.go
adityasaky: Doesn't GetReferenceEntriesInRange handle that case, where lastID is an annotation? Ideally, it should still return the ref entries in that range even if the end point isn't a ref entry. I think neither firstID nor lastID is required to be that of a reference entry FWIW.

adityasaky (changes requested):

spectre10:
Interesting! I suppose the motive behind this is the fact that annotation entries are created linearly and we want to skip those and update the relevant rsl entry with the annotation message. However, I was wondering if it was considered to have annotations branch out of the annotated rsl entry. That would make the rsl history clean and we could use the builtin commands to operate directly on `gittuf/reference-state-log` instead of creating separate logic. Those annotations could be organized in separate namespace such as `gittuf/annotations/$entryID`.

with the above approach this pr would end up like this:
```go
        ref, err := r.Head()
	CheckIfError(err)

	// ... retrieves the commit history
	since := time.Date(2019, 1, 1, 0, 0, 0, 0, time.UTC)
	until := time.Date(2019, 7, 30, 0, 0, 0, 0, time.UTC)
	cIter, err := r.Log(&git.LogOptions{From: ref.Hash(), Since: &since, Until: &until})
	CheckIfError(err)

        err = cIter.ForEach(func(c *object.Commit) error {
		UpdateWithAnnotationMessage(c.Message)
                fmt.Println(c)
		return nil
	})
	CheckIfError(err)
```
No need to handle formatting.
Although the performance aspect ought to be discussed due to the need of finding children commits.

adityasaky:
> However, I was wondering if it was considered to have annotations branch out of the annotated rsl entry. 

I'm not sure I fully understand this, could you provide an example? Thanks!

spectre10:
> similar to [`git-notes`](https://git-scm.com/docs/git-notes). From the same page:

although in the example I gave, a different ref is used for each annotation, while in git-notes it records it in a single ref `refs/notes/commits`. But, the idea is to track annotations separately.

internal/cmd/rsl/log/log.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/gittuf/gittuf/internal/display"
+	"github.com/spf13/cobra"
+)
+
+type options struct{}
+
+func (o *options) AddFlags(_ *cobra.Command) {}
+
+func (o *options) Run(_ *cobra.Command, _ []string) error {
+	return display.ShowRSL()
```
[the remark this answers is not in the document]
neilnaveen: I have changed it so it accepts a file path, and cmd/rsl/log creates a temp file, uses ShowRSL, and outputs the file using exec, with whatever PAGER setting the user has specified as their env

internal/common/common.go
neilnaveen: Fixed

internal/repository/rsl_test.go
neilnaveen: Fixed

internal/rsl/rsl.go
neilnaveen: Yes it does, fixed

adityasaky:
Okay, I understand better now. This is a little off topic, so if we continue discussing this beyond message, please open an issue.

The reason we don't do this right now is so that we want a strict ordering of all entries, including annotations and reference entries. If we move annotations to a separate ref, while yes, they can refer like a git-note does to the ref entry, we lose this ordering.

internal/repository/rsl_test.go
neilnaveen: Fixed

internal/display/display.go
```diff
@@ -0,0 +1,46 @@
+package display
+
+import (
+	"errors"
+	"io"
+	"os"
+	"os/exec"
+)
+
+func Display(defaultOutput, errorOutput io.Writer, contents []byte, page bool) error {
```
adityasaky, line 10: could you add some tests for this?

adityasaky: i also want us to walk through how the default works on windows for paging. It'd be worth looking at what git itself does here.

internal/display/rsl.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"fmt"
+
+	"github.com/gittuf/gittuf/internal/rsl"
+	"github.com/go-git/go-git/v5/plumbing"
+)
+
+// ShowRSL logs out the entries and annotations into a specified file
+func ShowRSL(entries []*rsl.ReferenceEntry, annotationMap map[plumbing.Hash][]*rsl.AnnotationEntry) string {
```
adityasaky, line 13: I'd rename this function to indicate it prepares the output that ought to be displayed.

internal/policy/policy.go
```diff
@@ -669,6 +674,9 @@ func ListRules(ctx context.Context, repo *git.Repository) ([]*DelegationWithDept
 
 	topLevelTargetsMetadata, err := state.GetTargetsMetadata(TargetsRoleName)
 	if err != nil {
+		if errors.Is(err, ErrMetadataNotFound) {
```
adityasaky, line 677: could you instead add a `state.HasRole` check before this? It's easier to read / reason through what's happening.

adityasaky (changes requested):

internal/policy/policy.go

internal/display/rsl.go
neilnaveen: Fixed

internal/display/display.go
neilnaveen: I have added some tests. I work on a Mac, so I don't think I can check if paging works differently for Windows.

internal/cmd/rsl/log/log.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	return nil
+}
+
+func New() *cobra.Command {
+	o := &options{}
+	cmd := &cobra.Command{
+		Use:               "log",
+		Short:             "Displays the Reference State Log",
```
adityasaky, line 41:
```suggestion
		Short:             "Display the Reference State Log",
```

internal/repository/rsl_test.go
```diff
@@ -561,3 +562,36 @@ func TestPullRSL(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	expected, _, err := rsl.GetReferenceEntriesInRange(r.r, firstEntry.GetID(), lastEntry.GetID())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	slices.Reverse(expected)
+
```
adityasaky, line 593:
```suggestion
```

internal/repository/rsl_test.go
```diff
@@ -561,3 +562,36 @@ func TestPullRSL(t *testing.T) {
...
+	assert.Equal(t, expected, entries)
+
```
adityasaky, line 595:
```suggestion
```

internal/display/display.go
adityasaky: Gotcha. I think we might need a _linux and _windows version of this file, basically. @wlynch wdyt?

internal/cmd/rsl/log/log.go
neilnaveen: Fixed

internal/repository/rsl_test.go
neilnaveen: Fixed

internal/repository/rsl_test.go
neilnaveen: Fixed

internal/display/display.go
wlynch: You can probably get away with using [exec.LookPath](https://pkg.go.dev/os/exec#LookPath)

For windows support, it looks like `more` is a supported command so if you iterate though

1. $PAGER
2. less
3. more
4. stdout

You'll hit something that will be supported.

internal/display/display.go
```diff
@@ -0,0 +1,49 @@
...
```
wlynch, line 12: I'd recommend structuring this a little bit differently. Instead of trying to pass through everything here, I'd have this func setup everything you need for a writer and return `(io.WriteCloser, error)`

```go
NewDisplayWriter(default io.Writer, page bool) (io.WriteCloser, error) {

if page {
  for _, bin := range []string{os.GetEnv("PAGER"), "less", "more"} {
    if _, err := exec.LookPath(bin); err != nil {
      continue
    }
    return execWriter(bin)
  }
}

return default

}
```

then where this is used:

```go
out, err := NewDisplayWriter(os.Stdout, true)
if err != nil {
  ...
}
defer out.Close()

fmt.Fprintln(out, ...)
```

This way it simplifies some of the parameters (e.g. should errors go to errorOutput or error?), and makes it easier to pass the writer around if we don't want to write all of the contents at once.

wlynch (changes requested):

internal/display/display.go
```diff
@@ -0,0 +1,110 @@
...
+package display
+
+import (
+	"errors"
+	"io"
+	"os"
+	"os/exec"
+)
+
+var pagerPrograms = []string{os.Getenv("PAGER"), "less", "more"}
```
neilnaveen, line 12: Should't this include stdout?

adityasaky:
This is blocked on me fiddling with the display package. :(

https://github.com/gittuf/gittuf/pull/282/files#diff-9612ef033ee230b6dd921d82f4b9d79c63886ac80599b4275af5e1b6121b596cR14-R30

https://github.com/gittuf/gittuf/pull/282#discussion_r1546557093

internal/display/display.go
adityasaky: no, that's a list of pager programs?

neilnaveen:
I have pushed some changes, which should fix all the problems with display. pageWriter was not working because we were starting the pager.command before we set up the acutal writer we would use for the pager with StdinPipe. This does mean that when we use Write with a pager, it automaticaly closes the writer too.

internal/display/display.go
neilnaveen: I have refactored Aditya's version of this to work both in test, and in the command line.

adityasaky:
@wlynch could you take a look?

neilnaveen:
Rebased with main

wlynch (commented):
Aside from the test failures, LGTM!

neilnaveen:
@adityasaky @wlynch, I am not able to figure out why the test are failing for macos but not ubuntu? They work on my local machine, which is a mac? Could they be flakey tests?

adityasaky (approved):

adityasaky:
Thanks @neilnaveen!

PR #283, opened 2024-01-18
policy: Add caching

adityasaky:
~Pulling some early ideas together.~

This PR adds a cache for verifiers identified for some path in a specific policy state.


adityasaky:
Updated this PR to only cache the verifiers identified for a path. Originally, this was also caching loaded policy states but this is not necessary anymore since we don't have the potential to load the same policy state multiple times (which we were doing when we were identifying policy for each commit).

wlynch (approved):

PR #285, opened 2024-01-19
Add from flag to verify-ref

patzielinski:
This PR adds a `--from-entry` flag to verify a ref from a certain RSL entry onwards.

### DIFF
--- a/internal/repository/verify.go
+++ b/internal/repository/verify.go
@@ -20,7 +20,7 @@ import (
 // another is to create a new RSL entry for the current state.
 var ErrRefStateDoesNotMatchRSL = errors.New("Git reference's current state does not match latest RSL entry") //nolint:stylecheck
 
-func (r *Repository) VerifyRef(ctx context.Context, target string, latestOnly bool) error {
+func (r *Repository) VerifyRef(ctx context.Context, target string, latestOnly bool, from string) error {
 	var (
 		expectedTip plumbing.Hash
 		err         error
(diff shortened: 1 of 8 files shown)


adityasaky:
Blocked by #284

adityasaky:
@wlynch I've helped Patrick with this PR so I shouldn't be reviewing it...

internal/repository/verify.go
```diff
@@ -43,6 +44,26 @@ func (r *Repository) VerifyRef(ctx context.Context, target string, latestOnly bo
 	return r.verifyRefTip(target, expectedTip)
 }
 
+func (r *Repository) VerifyRefFromEntry(ctx context.Context, target, entryID string) error {
+	if !dev.InDevMode() {
```
wlynch, line 48: If this is a dev mode feature, we should try to make this more obvious from the flag options. e.g. we should move a variant of this command to `gittuf dev` (i.e. `gittuf dev verify-ref`), mark the flags as hidden, or worst case make sure the flag is documented.

wlynch (commented):
Looks good! Just a small thing to make sure the flag is documented.

internal/repository/verify.go
adityasaky: Good catch, I'm going to add the dev mode only message to the flag.

adityasaky: Done!

wlynch (approved):

PR #287, opened 2024-01-23
Add verify-ref --from-commit

adityasaky:
Breaks away --from-commit from #285. This needs a bit more thinking.


adityasaky:
Closing this, we've largely removed the functionality of identifying the first time a commit was seen in the RSL as it's hard to be sure without walking the full RSL.

PR #295, opened 2024-02-02
clone: display progress on Stdout

spectre10:
Currently, the clone command does not show any output which is traditionally shown by git (e.g. enumerating objects, compressing objects, progress etc.). Enable this on Stdout.


adityasaky (approved):
Thanks! @patzielinski can you pull this in and test?

adityasaky:
Thanks!

PR #296, opened 2024-02-03
Fix Apache 2.0 license file

adityasaky:
It was pointed out to me (cc @06kellyjac) that the LICENSE file must include the full text of the Apache 2.0 license. The text we previously included is meant to license specific files rather than the project as a whole.

Note that this does not change how gittuf is licensed.


JustinCappos (approved):

06kellyjac:
Awesome 👍

PR #302, opened 2024-02-09
Add debug log statements

adityasaky:
This PR takes over #249.

It's possible I'm misusing the slog format here...

### DIFF
--- a/internal/policy/verify.go
+++ b/internal/policy/verify.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"strings"
 
 	"github.com/gittuf/gittuf/internal/gitinterface"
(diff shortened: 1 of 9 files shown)


internal/policy/verify.go
```diff
@@ -74,18 +80,21 @@ func VerifyRef(ctx context.Context, repo *git.Repository, target string) (plumbi
...
 func VerifyRefFull(ctx context.Context, repo *git.Repository, target string) (plumbing.Hash, error) {
 	// 1. Trace RSL back to the start
+	slog.Debug("Identifying first RSL entry...")
 	firstEntry, _, err := rsl.GetFirstEntry(repo)
 	if err != nil {
 		return plumbing.ZeroHash, err
 	}
 
 	// 2. Find latest entry for target
+	slog.Debug(fmt.Sprintf("Identifying latest RSL entry for '%s'...", target))
```
spectre10, line 90: I wonder if some these generic variables like target ref for example, could be converted to key-val pairs and be used throughout the codebase. Might be helpful in filtering out logs, like for a specific ref.

wlynch: Not a bad idea! I've been using https://pkg.go.dev/github.com/chainguard-dev/clog to plumb key/value pairs through the context.

Not blocking for this change though.

adityasaky: @spectre10 do you want to take a stab at working on this?

spectre10: @adityasaky I'll try `chainguard-dev/clog` and see if it's worth doing.

PR #307, opened 2024-02-14
Add extra verification when verifying tags

neilnaveen:
- Adds an extra check that verifies that the target hash of the tag object is the same as its corresponding RSL entry's target hash.

### DIFF
--- a/internal/policy/verify.go
+++ b/internal/policy/verify.go
@@ -682,6 +682,15 @@ func verifyTagEntry(ctx context.Context, repo *git.Repository, policy *State, en
 		return err
 	}
 
+	entryTagRef, err := repo.Reference(plumbing.ReferenceName(entry.RefName), true)
+	if err != nil {
+		return err
+	}
+
+	if entry.TargetID != entryTagRef.Hash() && entry.TargetID != tagObj.Target {
+		return fmt.Errorf("verifying RSL entry failed, tag reference set to unexpected target")
+	}
+
 	if len(tagObj.PGPSignature) == 0 {
 		return fmt.Errorf(noSignatureMessage)
 	}
--- a/internal/policy/verify_test.go
+++ b/internal/policy/verify_test.go
@@ -744,39 +744,90 @@ func TestVerifyCommit(t *testing.T) {
 }
 
 func TestVerifyTag(t *testing.T) {
-	repo, _ := createTestRepository(t, createTestStateWithPolicy)
-	refName := "refs/heads/main"
+	t.Run("normal test", func(t *testing.T) {
+		repo, _ := createTestRepository(t, createTestStateWithPolicy)
+		refName := "refs/heads/main"
 
-	commitIDs := common.AddNTestCommitsToSpecifiedRef(t, repo, refName, 3, gpgKeyBytes)
-	entry := rsl.NewReferenceEntry(refName, commitIDs[len(commitIDs)-1])
-	entryID := common.CreateTestRSLReferenceEntryCommit(t, repo, entry, gpgKeyBytes)
-	entry.ID = entryID
+		commitIDs := common.AddNTestCommitsToSpecifiedRef(t, repo, refName, 3, gpgKeyBytes)
+		entry := rsl.NewReferenceEntry(refName, commitIDs[len(commitIDs)-1])
+		entryID := common.CreateTestRSLReferenceEntryCommit(t, repo, entry, gpgKeyBytes)
+		entry.ID = entryID
 
-	tagName := "v1"
-	tagID := common.CreateTestSignedTag(t, repo, tagName, commitIDs[len(commitIDs)-1], gpgKeyBytes)
+		tagName := "v1"
+		tagID := common.CreateTestSignedTag(t, repo, tagName, commitIDs[len(commitIDs)-1], gpgKeyBytes)
 
-	expectedStatus := map[string]string{tagID.String(): unableToFindRSLEntryMessage}
-	status := VerifyTag(context.Background(), repo, []string{tagID.String()})
-	assert.Equal(t, expectedStatus, status)
+		expectedStatus := map[string]string{tagID.String(): unableToFindRSLEntryMessage}
+		status := VerifyTag(context.Background(), repo, []string{tagID.String()})
+		assert.Equal(t, expectedStatus, status)
 
-	entry = rsl.NewReferenceEntry(string(plumbing.NewTagReferenceName(tagName)), tagID)
-	entryID = common.CreateTestRSLReferenceEntryCommit(t, repo, entry, gpgKeyBytes)
-	entry.ID = entryID
+		entry = rsl.NewReferenceEntry(string(plumbing.NewTagReferenceName(tagName)), tagID)
+		entryID = common.CreateTestRSLReferenceEntryCommit(t, repo, entry, gpgKeyBytes)
+		entry.ID = entryID
 
-	// Use tag ID
-	expectedStatus = map[string]string{tagID.String(): goodTagSignatureMessage}
-	status = VerifyTag(context.Background(), repo, []string{tagID.String()})
-	assert.Equal(t, expectedStatus, status)
+		// Use tag ID
+		expectedStatus = map[string]string{tagID.String(): goodTagSignatureMessage}
+		status = VerifyTag(context.Background(), repo, []string{tagID.String()})
+		assert.Equal(t, expectedStatus, status)
 
-	// Use tagName
-	expectedStatus = map[string]string{tagName: goodTagSignatureMessage}
-	status = VerifyTag(context.Background(), repo, []string{tagName})
-	assert.Equal(t, expectedStatus, status)
+		// Use tagName
+		expectedStatus = map[string]string{tagName: goodTagSignatureMessage}
+		status = VerifyTag(context.Background(), repo, []string{tagName})
+		assert.Equal(t, expectedStatus, status)
 
-	// Use refs path for tagName
-	expectedStatus = map[string]string{string(plumbing.NewTagReferenceName(tagName)): goodTagSignatureMessage}
-	status = VerifyTag(context.Background(), repo, []string{string(plumbing.NewTagReferenceName(tagName))})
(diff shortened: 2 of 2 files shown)


internal/policy/verify.go
```diff
@@ -410,6 +410,23 @@ func VerifyTag(ctx context.Context, repo *git.Repository, ids []string) map[stri
 			continue
 		}
 
+		tagRef, err := repo.Tag(strings.Trim(absPath, "refs/tags"))
```
adityasaky, line 413: This could just use `repo.Reference` and avoid having to trim the prefix.

adityasaky: Also, this should maybe be a part of `verifyTagEntry` where we also load the tag object.

internal/policy/verify.go
```diff
@@ -410,6 +410,23 @@ func VerifyTag(ctx context.Context, repo *git.Repository, ids []string) map[stri
...
+		if err != nil {
+			status[id] = err.Error()
+			continue
+		}
+
+		tagObj, err := gitinterface.GetTag(repo, tagRef.Hash())
+
```
adityasaky, line 420:
```suggestion
```

internal/policy/verify.go
```diff
@@ -410,6 +410,23 @@ func VerifyTag(ctx context.Context, repo *git.Repository, ids []string) map[stri
...
+		if err != nil {
+			status[id] = err.Error()
+			continue
+		}
+
+		if entry.TargetID != tagObj.Target {
```
adityasaky, line 426: is this right? the entry may point to the tag obj, not the target of the tag obj

internal/policy/verify.go
adityasaky: Also, perhaps worth double checking that the entry's ref matches tagObj.Name.

adityasaky (changes requested):
Thanks for this patch! Some comments...

internal/policy/verify.go
neilnaveen: I have moved it to verifyTagEntry, and I have added a check to double check that the entry's ref matches the tagObj.Name.

internal/policy/verify.go

internal/policy/verify.go
neilnaveen: I have changed this to check if it is either pointing to the entry.TargetID, or the tagObj itself

internal/policy/verify.go
neilnaveen: repo.Reference still outputs in the format refs/tags/tagname, so trim would still have to be done.

internal/policy/verify.go
```diff
@@ -682,6 +682,14 @@ func verifyTagEntry(ctx context.Context, repo *git.Repository, policy *State, en
 		return err
 	}
 
+	if entry.TargetID != tagObj.Target && entry.TargetID != tagObj.Hash {
+		return fmt.Errorf("verifying RSL entry failed, %s", ErrUnauthorizedSignature.Error())
+	}
+
+	if entry.RefName != "refs/tags/"+tagObj.Name {
+		return fmt.Errorf("verifying RSL entry failed, tag name has been changed")
```
adityasaky, line 690: it may have changed or it may not have, it may actually be the case that refs/tags/<x> points to tag <y>. Changed is more indicative that the same tag name has a different target. I'd rephrase this to just say they don't match.

FWIW, I think Git may have some default protections for this submitted as part of the original omitting commits work.

adityasaky (changes requested):
This largely looks good to me, just a couple quick comments.

I also wonder if we want to update the rsl record flow for tags to perform the same checks.

adityasaky (dismissed):
Looks good! Can you add some tests that hit these checks?

internal/policy/verify.go
```diff
@@ -682,6 +682,14 @@ func verifyTagEntry(ctx context.Context, repo *git.Repository, policy *State, en
 		return err
 	}
 
+	if entry.TargetID != tagObj.Hash && entry.TargetID != tagObj.Target {
+		return fmt.Errorf("verifying RSL entry failed, %s", ErrUnauthorizedSignature.Error())
+	}
+
+	if entry.RefName != plumbing.NewTagReferenceName(tagObj.Name).String() {
+		return fmt.Errorf(plumbing.NewTagReferenceName(entry.RefName).String())
```
adityasaky, line 690: error needs to be updated? Perhaps that it doesn't match. Also, you can just use entry.RefName afaict.

neilnaveen:
I have added additional test, getting coverage for 

```
if entry.RefName != plumbing.NewTagReferenceName(tagObj.Name).String() {
 	return fmt.Errorf("verifying RSL entry failed, tag name does not match expected tag name from RSL")
 }
```

I am not sure how to test the other clause, how would I go about this? For this to be done, I would need to record two RSL entries, which would error out, since tags cannot have more than one.

adityasaky:
this is with the standard rsl.New<>.Commit()? IIRC if you craft the entries directly, you should be fine, can you check?

neilnaveen:
That was not actualy the problem, the problem was that I was not checking if our entry tag from the ref's hash is not equal to the tag got fromt the entry's targetHash's hash.

internal/policy/verify.go
```diff
@@ -682,6 +682,15 @@ func verifyTagEntry(ctx context.Context, repo *git.Repository, policy *State, en
 		return err
 	}
 
+	entryTagRef, err := repo.Reference(plumbing.ReferenceName(entry.RefName), false)
```
adityasaky, line 685: If you set this to `true`, does it resolve to the target?

internal/policy/verify_test.go
```diff
@@ -744,39 +744,65 @@ func TestVerifyCommit(t *testing.T) {
...
+		entryID = common.CreateTestRSLReferenceEntryCommit(t, repo, entry, gpgKeyBytes)
+		entry.ID = entryID
+
+		common.CreateTestSignedTag(t, repo, tagName, commitIDs[len(commitIDs)-2], gpgKeyBytes)
+
+		// Use tag ID
+		expectedStatus := map[string]string{tagID.String(): fmt.Sprintf("verifying RSL entry failed, %s", ErrUnauthorizedSignature.Error())}
+		status := VerifyTag(context.Background(), repo, []string{tagID.String()})
+		assert.Equal(t, expectedStatus, status)
+	})
```
adityasaky, line 805: This doesn't include a test for duplicate checks in the verification flow, right?

internal/policy/verify.go
neilnaveen: Resolving it will either keep it the same if it is the targets hash, or if it is the tag's hash, it will change it into its target, so at the end it is just better to resolve it. Fixed.

internal/policy/verify_test.go
neilnaveen: Sorry, could you clarify what you mean?

internal/policy/verify.go
```diff
@@ -682,6 +682,15 @@ func verifyTagEntry(ctx context.Context, repo *git.Repository, policy *State, en
 		return err
 	}
 
+	entryTagRef, err := repo.Reference(plumbing.ReferenceName(entry.RefName), true)
+	if err != nil {
+		return err
+	}
+
+	if entry.TargetID != entryTagRef.Hash() && entry.TargetID != tagObj.Target {
+		return fmt.Errorf("verifying RSL entry failed, %s", ErrUnauthorizedSignature.Error())
```
adityasaky, line 691: I'd not use ErrUnauthorizedSignature here, let's use an error that reflects the mismatch in metadata?

internal/policy/verify_test.go
adityasaky: recording tag v1 in the RSL and then recording the exact same tag again in the RSL. The RSL should enforce immutability of tags it's aware of.

adityasaky:
Outstanding minor nits aside, we should be able to merge this soon. Sorry for the delay!

internal/policy/verify.go

internal/policy/verify_test.go
neilnaveen: I have added this test

internal/policy/verify.go
```diff
@@ -682,6 +682,15 @@ func verifyTagEntry(ctx context.Context, repo *git.Repository, policy *State, en
 		return err
 	}
 
+	entryTagRef, err := repo.Reference(plumbing.ReferenceName(entry.RefName), true)
+	if err != nil {
+		return err
+	}
+
+	if entry.TargetID != entryTagRef.Hash() && entry.TargetID != tagObj.Target {
+		return fmt.Errorf("verifying RSL entry failed, gittuf metadata does not match tag data")
```
adityasaky, line 691:
```suggestion
		return fmt.Errorf("verifying RSL entry failed, tag reference set to unexpected target")
```

internal/policy/verify_test.go
```diff
@@ -744,39 +744,89 @@ func TestVerifyCommit(t *testing.T) {
...
+		entry.ID = entryID
+
+		common.CreateTestSignedTag(t, repo, tagName, commitIDs[len(commitIDs)-2], gpgKeyBytes)
+
+		// Use tag ID
+		expectedStatus := map[string]string{tagID.String(): "verifying RSL entry failed, gittuf metadata does not match tag data"}
+		status := VerifyTag(context.Background(), repo, []string{tagID.String()})
+		assert.Equal(t, expectedStatus, status)
+	})
+	t.Run("tag verification with multiple RSL entries", func(t *testing.T) {
```
adityasaky, line 806:
```suggestion

	t.Run("tag verification with multiple RSL entries", func(t *testing.T) {
```

adityasaky (approved):
some nits, but LGTM!

adityasaky:
Could you address the nits + clean up commits? Thanks!

internal/policy/verify_test.go
neilnaveen: Fixed

internal/policy/verify.go

neilnaveen:
I have cleaned up the commit history and fixed the nits

adityasaky (approved):
LGTM, thanks!

PR #312, opened 2024-02-28
Add expected root keys flag for cloning 

neilnaveen:
- Added the expectedRootKeys flag to clone.go
- Added additional test in sync_test.go for expectedRootKeys

Signed-off-by: neilnaveen [42328488+neilnaveen@users.noreply.github.com](mailto:42328488+neilnaveen@users.noreply.github.com)

### DIFF
--- a/docs/cli/gittuf_clone.md
+++ b/docs/cli/gittuf_clone.md
@@ -9,8 +9,9 @@ gittuf clone [flags]
 ### Options
 
 ```
-  -b, --branch string   specify branch to check out
-  -h, --help            help for clone
+  -b, --branch string          specify branch to check out
+  -h, --help                   help for clone
+      --root-key public-keys   set of initial root of trust keys for the repository (supported values: paths to SSH keys, GPG key fingerprints, Sigstore/Fulcio identities)
 ```
 
 ### Options inherited from parent commands
--- a/internal/cmd/clone/clone.go
+++ b/internal/cmd/clone/clone.go
@@ -3,12 +3,15 @@
 package clone
 
 import (
+	"github.com/gittuf/gittuf/internal/cmd/common"
 	"github.com/gittuf/gittuf/internal/repository"
+	"github.com/gittuf/gittuf/internal/tuf"
 	"github.com/spf13/cobra"
 )
 
 type options struct {
-	branch string
+	branch           string
+	expectedRootKeys common.PublicKeys
 }
 
 func (o *options) AddFlags(cmd *cobra.Command) {
@@ -19,6 +22,11 @@ func (o *options) AddFlags(cmd *cobra.Command) {
 		"",
 		"specify branch to check out",
 	)
+	cmd.Flags().Var(
+		&o.expectedRootKeys,
+		"root-key",
+		"set of initial root of trust keys for the repository (supported values: paths to SSH keys, GPG key fingerprints, Sigstore/Fulcio identities)",
+	)
 }
 
 func (o *options) Run(cmd *cobra.Command, args []string) error {
@@ -26,7 +34,19 @@ func (o *options) Run(cmd *cobra.Command, args []string) error {
 	if len(args) > 1 {
 		dir = args[1]
 	}
-	_, err := repository.Clone(cmd.Context(), args[0], dir, o.branch)
+
+	expectedRootKeys := make([]*tuf.Key, len(o.expectedRootKeys))
+
+	for index, keyPath := range o.expectedRootKeys {
+		key, err := common.LoadPublicKey(keyPath)
+		if err != nil {
+			return err
+		}
+
+		expectedRootKeys[index] = key
+	}
+
+	_, err := repository.Clone(cmd.Context(), args[0], dir, o.branch, expectedRootKeys)
 	return err
 }
 
--- a/internal/cmd/common/common.go
+++ b/internal/cmd/common/common.go
@@ -22,6 +22,25 @@ const (
 	FulcioPrefix = "fulcio:"
 )
 
+// PublicKeys is a custom type to represent a list of paths
+type PublicKeys []string
+
+// String implements part of the pflag.Value interface.
+func (p *PublicKeys) String() string {
+	return strings.Join(*p, ", ")
+}
+
+// Set implements part of the pflag.Value interface.
+func (p *PublicKeys) Set(value string) error {
+	*p = append(*p, value)
+	return nil
+}
+
+// Type implements part of the pflag.Value interface.
+func (p *PublicKeys) Type() string {
+	return "public-keys"
+}
+
 // LoadPublicKey returns a tuf.Key object for a PGP / Sigstore Fulcio / SSH
 // (on-disk) key for use in gittuf metadata.
 func LoadPublicKey(key string) (*tuf.Key, error) {
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -141,6 +141,17 @@ func LoadCurrentState(ctx context.Context, repo *git.Repository, ref string) (*S
 	return LoadState(ctx, repo, entry)
 }
 
+// LoadFirstState returns the State corresponding to the repository's first
+// active policy. It does not verify the root of trust since it is the initial policy.
+func LoadFirstState(ctx context.Context, repo *git.Repository) (*State, error) {
+	firstEntry, _, err := rsl.GetFirstReferenceEntryForRef(repo, PolicyRef)
+	if err != nil {
+		return nil, err
+	}
+
+	return LoadState(ctx, repo, firstEntry)
+}
+
 // GetStateForCommit scans the RSL to identify the first time a commit was seen
 // in the repository. The policy preceding that RSL entry is returned as the
 // State to be used for verifying the commit's signature. If the commit hasn't
--- a/internal/policy/policy_test.go
+++ b/internal/policy/policy_test.go
@@ -270,6 +270,48 @@ func TestLoadCurrentState(t *testing.T) {
 	assert.Equal(t, state, loadedState)
 }
 
+func TestLoadFirstState(t *testing.T) {
+	repo, firstState := createTestRepository(t, createTestStateWithPolicy)
+
+	// Update policy, record in RSL
+	secondState, err := LoadCurrentState(context.Background(), repo, PolicyRef) // secondState := state will modify state as well
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	targetsMetadata, err := secondState.GetTargetsMetadata(TargetsRoleName)
+	if err != nil {
+		t.Fatal(err)
+	}
+	targetsMetadata, err = AddDelegation(targetsMetadata, "new-rule", []*tuf.Key{}, []string{"*"}, 1) // just a dummy rule
+	if err != nil {
+		t.Fatal(err)
+	}
+	signer, err := signerverifier.NewSignerVerifierFromSecureSystemsLibFormat(rootKeyBytes) //nolint:staticcheck
+	if err != nil {
+		t.Fatal(err)
+	}
+	targetsEnv, err := dsse.CreateEnvelope(targetsMetadata)
+	if err != nil {
+		t.Fatal(err)
+	}
+	targetsEnv, err = dsse.SignEnvelope(context.Background(), targetsEnv, signer)
+	if err != nil {
+		t.Fatal(err)
+	}
+	secondState.TargetsEnvelope = targetsEnv
+	if err := secondState.Commit(repo, "Second state", false); err != nil {
+		t.Fatal(err)
+	}
+
+	loadedState, err := LoadFirstState(context.Background(), repo)
+	if err != nil {
+		t.Error(err)
+	}
+
+	assert.Equal(t, firstState, loadedState)
+}
+
 func TestLoadStateForEntry(t *testing.T) {
 	repo, state := createTestRepository(t, createTestStateWithOnlyRoot)
 
--- a/internal/repository/sync.go
+++ b/internal/repository/sync.go
@@ -8,22 +8,27 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"reflect"
+	"sort"
 	"strings"
 
 	"github.com/gittuf/gittuf/internal/gitinterface"
+	"github.com/gittuf/gittuf/internal/policy"
+	"github.com/gittuf/gittuf/internal/tuf"
 	"github.com/go-git/go-git/v5/plumbing"
 )
 
 var (
-	ErrCloningRepository = errors.New("unable to clone repository")
-	ErrDirExists         = errors.New("directory exists")
+	ErrCloningRepository          = errors.New("unable to clone repository")
+	ErrDirExists                  = errors.New("directory exists")
+	ErrExpectedRootKeysDoNotMatch = errors.Join(ErrCloningRepository, errors.New("cloned root keys do not match the expected keys"))
 )
 
 // Clone wraps a typical git clone invocation, fetching gittuf refs in addition
 // to the standard refs. It performs a verification of the RSL against the
 // specified HEAD after cloning the repository.
 // TODO: resolve how root keys are trusted / bootstrapped.
-func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Repository, error) {
+func Clone(ctx context.Context, remoteURL, dir, initialBranch string, expectedRootKeys []*tuf.Key) (*Repository, error) {
 	slog.Debug(fmt.Sprintf("Cloning from '%s'...", remoteURL))
 
 	if dir == "" {
@@ -63,6 +68,35 @@ func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Reposito
 	}
 
 	repository := &Repository{r: r}
+
+	if len(expectedRootKeys) > 0 {
+		slog.Debug("Verifying if root keys are expected root keys...")
+
+		sort.Slice(expectedRootKeys, func(i, j int) bool {
+			return expectedRootKeys[i].KeyID < expectedRootKeys[j].KeyID
+		})
+
+		state, err := policy.LoadFirstState(ctx, r)
+		if err != nil {
+			return repository, errors.Join(ErrCloningRepository, err)
+		}
+		rootKeys, err := state.GetRootKeys()
+		if err != nil {
+			return repository, errors.Join(ErrCloningRepository, err)
+		}
+
+		// We sort the root keys so that we can check if the root keys array match's the expected root key array
+		sort.Slice(rootKeys, func(i, j int) bool {
+			return rootKeys[i].KeyID < rootKeys[j].KeyID
+		})
+
+		if len(rootKeys) != len(expectedRootKeys) {
+			return repository, ErrExpectedRootKeysDoNotMatch
+		}
+		if !reflect.DeepEqual(rootKeys, expectedRootKeys) {
+			return repository, ErrExpectedRootKeysDoNotMatch
+		}
+	}
 
 	slog.Debug("Verifying HEAD...")
 	return repository, repository.VerifyRef(ctx, head.Target().String(), false)
--- a/internal/repository/sync_test.go
+++ b/internal/repository/sync_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/gittuf/gittuf/internal/policy"
 	"github.com/gittuf/gittuf/internal/rsl"
 	"github.com/gittuf/gittuf/internal/signerverifier"
+	"github.com/gittuf/gittuf/internal/signerverifier/gpg"
 	"github.com/gittuf/gittuf/internal/tuf"
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
@@ -44,6 +45,9 @@ func TestClone(t *testing.T) {
 		t.Fatal(err)
 	}
 	if err := remoteRepo.AddTopLevelTargetsKey(context.Background(), rootSigner, targetsPubKey, false); err != nil {
+		t.Fatal(err)
+	}
+	if err := remoteRepo.AddRootKey(context.Background(), rootSigner, targetsPubKey, false); err != nil {
 		t.Fatal(err)
 	}
 	if err := remoteRepo.InitializeTargets(context.Background(), targetsSigner, policy.TargetsRoleName, false); err != nil {
@@ -80,15 +84,6 @@ func TestClone(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	remoteRSLRef, err := remoteRepo.r.Reference(plumbing.ReferenceName(rsl.Ref), true)
-	if err != nil {
-		t.Fatal(err)
-	}
-	remotePolicyRef, err := remoteRepo.r.Reference(plumbing.ReferenceName(policy.PolicyRef), true)
-	if err != nil {
-		t.Fatal(err)
-	}
-
 	t.Run("successful clone without specifying dir", func(t *testing.T) {
 		localTmpDir := t.TempDir()
 
@@ -97,24 +92,17 @@ func TestClone(t *testing.T) {
 		}
 		defer os.Chdir(currentDir) //nolint:errcheck
 
-		repo, err := Clone(context.Background(), remoteTmpDir, "", "")
+		repo, err := Clone(context.Background(), remoteTmpDir, "", "", nil)
 		assert.Nil(t, err)
+
 		head, err := repo.r.Head()
 		if err != nil {
 			t.Fatal(err)
 		}
 		assert.Equal(t, commitID, head.Hash())
 
-		localRSLRef, err := repo.r.Reference(plumbing.ReferenceName(rsl.Ref), true)
-		if err != nil {
-			t.Fatal(err)
-		}
-		assert.Equal(t, remoteRSLRef.Hash(), localRSLRef.Hash())
-		localPolicyRef, err := repo.r.Reference(plumbing.ReferenceName(policy.PolicyRef), true)
-		if err != nil {
-			t.Fatal(err)
-		}
-		assert.Equal(t, remotePolicyRef.Hash(), localPolicyRef.Hash())
+		assertLocalAndRemoteRefsMatch(t, repo.r, remoteRepo.r, rsl.Ref)
+		assertLocalAndRemoteRefsMatch(t, repo.r, remoteRepo.r, policy.PolicyRef)
 	})
 
 	t.Run("successful clone with dir", func(t *testing.T) {
@@ -126,8 +114,9 @@ func TestClone(t *testing.T) {
 		defer os.Chdir(currentDir) //nolint:errcheck
 
 		dirName := "myRepo"
-		repo, err := Clone(context.Background(), remoteTmpDir, dirName, "")
+		repo, err := Clone(context.Background(), remoteTmpDir, dirName, "", nil)
 		assert.Nil(t, err)
+
 		head, err := repo.r.Head()
 		if err != nil {
 			t.Fatal(err)
@@ -138,16 +127,8 @@ func TestClone(t *testing.T) {
 		assert.Nil(t, err)
 		assert.True(t, dirInfo.IsDir())
 
-		localRSLRef, err := repo.r.Reference(plumbing.ReferenceName(rsl.Ref), true)
-		if err != nil {
-			t.Fatal(err)
-		}
-		assert.Equal(t, remoteRSLRef.Hash(), localRSLRef.Hash())
-		localPolicyRef, err := repo.r.Reference(plumbing.ReferenceName(policy.PolicyRef), true)
-		if err != nil {
-			t.Fatal(err)
-		}
-		assert.Equal(t, remotePolicyRef.Hash(), localPolicyRef.Hash())
+		assertLocalAndRemoteRefsMatch(t, repo.r, remoteRepo.r, rsl.Ref)
+		assertLocalAndRemoteRefsMatch(t, repo.r, remoteRepo.r, policy.PolicyRef)
 	})
 
 	t.Run("successful clone without specifying dir, with non-HEAD initial branch", func(t *testing.T) {
@@ -158,8 +139,9 @@ func TestClone(t *testing.T) {
 		}
 		defer os.Chdir(currentDir) //nolint:errcheck
 
-		repo, err := Clone(context.Background(), remoteTmpDir, "", anotherRefName)
+		repo, err := Clone(context.Background(), remoteTmpDir, "", anotherRefName, nil)
 		assert.Nil(t, err)
+
 		head, err := repo.r.Head()
 		if err != nil {
 			t.Fatal(err)
@@ -167,16 +149,8 @@ func TestClone(t *testing.T) {
 		assert.Equal(t, commitID, head.Hash())
 		assert.Equal(t, plumbing.ReferenceName(anotherRefName), head.Name())
 
-		localRSLRef, err := repo.r.Reference(plumbing.ReferenceName(rsl.Ref), true)
-		if err != nil {
-			t.Fatal(err)
-		}
-		assert.Equal(t, remoteRSLRef.Hash(), localRSLRef.Hash())
-		localPolicyRef, err := repo.r.Reference(plumbing.ReferenceName(policy.PolicyRef), true)
-		if err != nil {
-			t.Fatal(err)
-		}
-		assert.Equal(t, remotePolicyRef.Hash(), localPolicyRef.Hash())
+		assertLocalAndRemoteRefsMatch(t, repo.r, remoteRepo.r, rsl.Ref)
+		assertLocalAndRemoteRefsMatch(t, repo.r, remoteRepo.r, policy.PolicyRef)
 	})
 
 	t.Run("unsuccessful clone when unspecified dir already exists", func(t *testing.T) {
@@ -187,10 +161,10 @@ func TestClone(t *testing.T) {
 		}
 		defer os.Chdir(currentDir) //nolint:errcheck
 
-		_, err = Clone(context.Background(), remoteTmpDir, "", "")
+		_, err = Clone(context.Background(), remoteTmpDir, "", "", nil)
 		assert.Nil(t, err)
 
-		_, err = Clone(context.Background(), remoteTmpDir, "", "")
+		_, err = Clone(context.Background(), remoteTmpDir, "", "", nil)
 		assert.ErrorIs(t, err, ErrDirExists)
 	})
 
@@ -203,10 +177,11 @@ func TestClone(t *testing.T) {
 		defer os.Chdir(currentDir) //nolint:errcheck
 
 		dirName := "myRepo"
-		if err := os.Mkdir(dirName, 0755); err != nil {
+		if err := os.Mkdir(dirName, 0o755); err != nil {
 			t.Fatal(err)
 		}
-		_, err = Clone(context.Background(), remoteTmpDir, dirName, "")
+
+		_, err = Clone(context.Background(), remoteTmpDir, dirName, "", nil)
 		assert.ErrorIs(t, err, ErrDirExists)
 	})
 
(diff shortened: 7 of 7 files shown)


docs/cli/gittuf_clone.md
```diff
@@ -9,8 +9,9 @@ gittuf clone [flags]
 ### Options
 
 ```
-  -b, --branch string   specify branch to check out
-  -h, --help            help for clone
+  -b, --branch string              specify branch to check out
+  -e, --expectedRootKeys strings   specify expected root keys in cloned repo
```
wlynch, line 13: Some name nitting to try and make it clearer what this flag does (happy to go back and forth on this)

```suggestion
  -k, --root-keys strings   path(s) of root keys to verify when cloning the repo.
```

neilnaveen: Sorry for the late reply, fixed.

internal/repository/sync.go
```diff
@@ -64,8 +68,33 @@ func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Reposito
...
+	sort.Slice(expectedRootKeys, func(i, j int) bool {
+		return expectedRootKeys[i].KeyID < expectedRootKeys[j].KeyID
+	})
+
 	repository := &Repository{r: r}
 
 	slog.Debug("Verifying HEAD...")
+
+	if len(expectedRootKeys) > 0 {
+		state, err := policy.LoadCurrentState(ctx, r)
```
adityasaky, line 80: To be clear, I think this should be the _first_ ever complete policy state. This feature is a mechanism to bootstrap the trust for the first ever* root.json. Subsequent versions of root (including their keys) can be verified using the immediate preceding version of the root. It's only the very first root keys we need to bootstrap entirely independently.

\* That said, there may be situations where there's been a compromise of a threshold of root keys, but we'll have to account for checkpointing that separately anyway.

internal/repository/sync.go
```diff
@@ -64,8 +68,33 @@ func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Reposito
...
+		if err != nil {
+			return repository, errors.Join(ErrCloningRepository, err)
+		}
+		rootKeys, err := state.GetRootKeys()
+		if err != nil {
+			return repository, errors.Join(ErrCloningRepository, err)
+		}
```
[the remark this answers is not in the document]
adityasaky: Sorry for the delay! I tend to agree, though this is actually for the very first set of root keys. I wonder if we can make that available when loading the first root's keys to go from 1.root to the currently applicable policy state.

internal/repository/sync.go
```diff
@@ -64,8 +68,33 @@ func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Reposito
...
+
+		if len(rootKeys) != len(expectedRootKeys) {
+			return repository, ErrClonedAndExpectedKeysDoNotMatch
+		}
+		slog.Debug("Verifying if root keys are expected root keys...")
```
adityasaky, line 92: I'd move this up before the length check since that's a shortcut to verifying expected root keys.

internal/repository/sync_test.go
```diff
@@ -94,7 +94,7 @@ func TestClone(t *testing.T) {
 		}
 		defer os.Chdir(currentDir) //nolint:errcheck
 
-		repo, err := Clone(context.Background(), remoteTmpDir, "", "")
+		repo, err := Clone(context.Background(), remoteTmpDir, "", "", []*tuf.Key{})
```
adityasaky, line 97: can this parameter (and others) be nil?

internal/repository/sync_test.go
```diff
@@ -234,4 +234,48 @@ func TestClone(t *testing.T) {
 		}
 		assert.Equal(t, remotePolicyRef.Hash(), localPolicyRef.Hash())
 	})
+	t.Run("successful clone without specifying dir, with expected root keys", func(t *testing.T) {
```
adityasaky, line 237:
```suggestion

	t.Run("successful clone without specifying dir, with expected root keys", func(t *testing.T) {
```

internal/repository/sync_test.go
```diff
@@ -234,4 +234,48 @@ func TestClone(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		assert.Equal(t, remoteRSLRef.Hash(), localRSLRef.Hash())
+		localPolicyRef, err := repo.r.Reference(plumbing.ReferenceName(policy.PolicyRef), true)
+		if err != nil {
+			t.Fatal(err)
+		}
+		assert.Equal(t, remotePolicyRef.Hash(), localPolicyRef.Hash())
+	})
+	t.Run("unsuccessful clone without specifying dir, with expected root keys not equaling root keys", func(t *testing.T) {
```
adityasaky, line 267:
```suggestion

	t.Run("unsuccessful clone without specifying dir, with expected root keys not equaling root keys", func(t *testing.T) {
```

adityasaky (changes requested):

internal/repository/sync_test.go
neilnaveen: Fixed

internal/repository/sync_test.go
neilnaveen: Fixed

internal/repository/sync.go
neilnaveen: I have fixed this

internal/repository/sync_test.go
adityasaky: yeah, I mean let's not pass an empty slice.

internal/repository/sync.go
```diff
@@ -64,8 +68,33 @@ func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Reposito
...
+		state, err := policy.LoadFirstState(ctx, r)
+		if err != nil {
+			return repository, errors.Join(ErrCloningRepository, err)
+		}
+		rootKeys, err := state.GetRootKeys()
+		if err != nil {
+			return repository, errors.Join(ErrCloningRepository, err)
+		}
+
+		slog.Debug("Verifying if root keys are expected root keys...")
```
adityasaky, line 87: I'd move this even further up, right at the start of the if block. Loading all policy states to find the first one may take a moment, and it'd be good to know why.

internal/repository/sync_test.go
```diff
@@ -234,4 +234,50 @@ func TestClone(t *testing.T) {
...
+	})
+
+	t.Run("unsuccessful clone without specifying dir, with expected root keys not equaling root keys", func(t *testing.T) {
+		localTmpDir := t.TempDir()
+
+		if err := os.Chdir(localTmpDir); err != nil {
+			t.Fatal(err)
+		}
+		defer os.Chdir(currentDir) //nolint:errcheck
+		expectedPublicKey, err := tuf.LoadKeyFromBytes(targetsKeyBytes)
```
adityasaky, line 276:
```suggestion

		expectedPublicKey, err := tuf.LoadKeyFromBytes(targetsKeyBytes)
```

internal/repository/sync_test.go
```diff
@@ -234,4 +234,50 @@ func TestClone(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		_, err = Clone(context.Background(), remoteTmpDir, "", "", []*tuf.Key{expectedPublicKey})
```
adityasaky, line 280:
```suggestion

		_, err = Clone(context.Background(), remoteTmpDir, "", "", []*tuf.Key{expectedPublicKey})
```

internal/repository/sync_test.go
```diff
@@ -234,4 +234,50 @@ func TestClone(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		assert.Equal(t, commitID, head.Hash())
+
+		localRSLRef, err := repo.r.Reference(plumbing.ReferenceName(rsl.Ref), true)
+		if err != nil {
+			t.Fatal(err)
+		}
+		assert.Equal(t, remoteRSLRef.Hash(), localRSLRef.Hash())
+		localPolicyRef, err := repo.r.Reference(plumbing.ReferenceName(policy.PolicyRef), true)
```
adityasaky, line 262:
```suggestion

		localPolicyRef, err := repo.r.Reference(plumbing.ReferenceName(policy.PolicyRef), true)
```

internal/repository/sync_test.go
```diff
@@ -234,4 +234,50 @@ func TestClone(t *testing.T) {
...
 	})
+
+	t.Run("successful clone without specifying dir, with expected root keys", func(t *testing.T) {
+		localTmpDir := t.TempDir()
+
+		if err := os.Chdir(localTmpDir); err != nil {
+			t.Fatal(err)
+		}
+		defer os.Chdir(currentDir) //nolint:errcheck
+		publicKey, err := tuf.LoadKeyFromBytes(rootKeyBytes)
```
adityasaky, line 245:
```suggestion

		publicKey, err := tuf.LoadKeyFromBytes(rootKeyBytes)
```

internal/repository/sync_test.go
```diff
@@ -234,4 +234,50 @@ func TestClone(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		repo, err := Clone(context.Background(), remoteTmpDir, "", "", []*tuf.Key{publicKey})
```
adityasaky, line 249:
```suggestion

		repo, err := Clone(context.Background(), remoteTmpDir, "", "", []*tuf.Key{publicKey})
```

internal/repository/sync_test.go
```diff
@@ -234,4 +234,50 @@ func TestClone(t *testing.T) {
...
+		assert.Nil(t, err)
+		head, err := repo.r.Head()
```
adityasaky, line 251:
```suggestion

		head, err := repo.r.Head()
```

internal/repository/sync.go
```diff
@@ -45,7 +49,7 @@ func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Reposito
 		return nil, errors.Join(ErrCloningRepository, err)
 	}
 
-	if err := os.Mkdir(dir, 0755); err != nil {
+	if err := os.Mkdir(dir, 0o755); err != nil {
```
adityasaky, line 52: unrelated change

internal/repository/sync.go
```diff
@@ -8,24 +8,28 @@ import (
...
+	"github.com/gittuf/gittuf/internal/tuf"
 	"github.com/go-git/go-git/v5/plumbing"
 )
 
 var (
-	ErrCloningRepository = errors.New("unable to clone repository")
-	ErrDirExists         = errors.New("directory exists")
+	ErrCloningRepository               = errors.New("unable to clone repository")
+	ErrDirExists                       = errors.New("directory exists")
+	ErrClonedAndExpectedKeysDoNotMatch = errors.New(ErrCloningRepository.Error() + ", " + "cloned root keys do not match the expected keys.")
```
adityasaky, line 25: I'd just errors.Join() when you're returning this error instead of this, but it may be a personal preference...

internal/policy/policy_test.go
```diff
@@ -94,6 +94,47 @@ func TestLoadCurrentState(t *testing.T) {
...
 
+func TestLoadFirstState(t *testing.T) {
+	repo, state := createTestRepository(t, createTestStateWithPolicy)
+
+	// Update policy, record in RSL
+	secondState, err := LoadCurrentState(context.Background(), repo) // secondState := state will modify state as well
+	if err != nil {
+		t.Fatal(err)
+	}
+	targetsMetadata, err := secondState.GetTargetsMetadata(TargetsRoleName)
```
adityasaky, line 105:
```suggestion

	targetsMetadata, err := secondState.GetTargetsMetadata(TargetsRoleName)
```

adityasaky: A rule of thumb I somewhat use is to have a blank line after an error check, though sometimes (say when I'm loading multiple keys) it feels okay to put them all together.

internal/cmd/clone/clone.go
```diff
@@ -19,14 +23,37 @@ func (o *options) AddFlags(cmd *cobra.Command) {
...
+	)
 }
 
 func (o *options) Run(cmd *cobra.Command, args []string) error {
 	var dir string
 	if len(args) > 1 {
 		dir = args[1]
 	}
-	_, err := repository.Clone(cmd.Context(), args[0], dir, o.branch)
+	expectedRootKeys := []*tuf.Key{}
```
adityasaky, line 40:
```suggestion

	expectedRootKeys := []*tuf.Key{}
```

adityasaky: This can use `make` based on the length of `o.expectedRootKeys` right?

adityasaky (changes requested):

internal/repository/sync.go
neilnaveen: Moved

internal/repository/sync_test.go
neilnaveen: Fixed

internal/repository/sync_test.go
neilnaveen: Fixed

internal/repository/sync_test.go
neilnaveen: Fixed

internal/repository/sync_test.go
neilnaveen: Fixed

internal/repository/sync.go
neilnaveen: Fixed

internal/policy/policy_test.go
neilnaveen: Fixed

internal/repository/sync.go
neilnaveen: Fixed, my gofmt/goimports keeps on formating it like that.

internal/repository/sync_test.go
neilnaveen: Fixed

internal/repository/sync_test.go
```diff
@@ -234,4 +234,56 @@ func TestClone(t *testing.T) {
...
+		}
+		defer os.Chdir(currentDir) //nolint:errcheck
+
+		expectedPublicKey, err := tuf.LoadKeyFromBytes(targetsKeyBytes)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		_, err = Clone(context.Background(), remoteTmpDir, "", "", []*tuf.Key{expectedPublicKey})
+		assert.Equal(t, err.Error(), ErrCloningRepository.Error()+", "+"cloned root keys do not match the expected keys.")
```
adityasaky, line 287: Couple things:
a) technically, the first value in assert.Equal (after `t`) is the expected value, it's backwards here
b) we should be using assert.ErrorIs and also maybe having a named error for the keys not matching (returned using errors.Join) to avoid this concat here

internal/repository/sync.go
```diff
@@ -64,8 +67,33 @@ func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Reposito
...
+		if err != nil {
+			return repository, errors.Join(ErrCloningRepository, err)
+		}
+
+		if len(rootKeys) != len(expectedRootKeys) {
+			return repository, errors.New(ErrCloningRepository.Error() + ", " + "cloned root keys do not match the expected keys.")
+		}
+		for keyid := range rootKeys {
+			if !reflect.DeepEqual(rootKeys[keyid], expectedRootKeys[keyid]) {
+				return repository, errors.New(ErrCloningRepository.Error() + ", " + "cloned root keys do not match the expected keys.")
```
adityasaky, line 92: Named error + errors.Join here and other instances?

internal/repository/sync.go
```diff
@@ -64,8 +67,33 @@ func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Reposito
...
+		}
+		rootKeys, err := state.GetRootKeys()
+		if err != nil {
+			return repository, errors.Join(ErrCloningRepository, err)
+		}
+
+		if len(rootKeys) != len(expectedRootKeys) {
+			return repository, errors.New(ErrCloningRepository.Error() + ", " + "cloned root keys do not match the expected keys.")
+		}
+		for keyid := range rootKeys {
```
adityasaky, line 90: this is the index, right? keyid has another accepted meaning, so this can be confusing. I first thought this was a map with the keyid as the key.

internal/policy/policy_test.go
adityasaky: I think there are more instances in this block that could be fixed up...

adityasaky (changes requested):
Getting pretty close. Also, we should open an issue to track repeated actions that @wlynch highlighted. Could you?

internal/policy/policy_test.go
neilnaveen: I think I have adressed them all?

internal/repository/sync_test.go
neilnaveen: Fixed, I just named the whole joined error

internal/repository/sync.go
neilnaveen: Fixed

internal/cmd/clone/clone.go
```diff
@@ -19,14 +23,37 @@ func (o *options) AddFlags(cmd *cobra.Command) {
 		"",
 		"specify branch to check out",
 	)
+	cmd.Flags().StringSliceVarP(
+		&o.expectedRootKeys,
+		"root-keys",
```
adityasaky, line 28:
```suggestion
		"root-key",
```

note that you'd pass in `--root-key <key> --root-key <key>` rather than `--root-keys <key> <key>`.

internal/cmd/clone/clone.go
```diff
@@ -19,14 +23,37 @@ func (o *options) AddFlags(cmd *cobra.Command) {
...
+	)
 }
 
 func (o *options) Run(cmd *cobra.Command, args []string) error {
 	var dir string
 	if len(args) > 1 {
 		dir = args[1]
 	}
-	_, err := repository.Clone(cmd.Context(), args[0], dir, o.branch)
+	expectedRootKeys := make([]*tuf.Key, len(o.expectedRootKeys))
```
adityasaky, line 40:
```suggestion

	expectedRootKeys := make([]*tuf.Key, len(o.expectedRootKeys))
```

internal/cmd/clone/clone.go
```diff
@@ -19,14 +23,37 @@ func (o *options) AddFlags(cmd *cobra.Command) {
...
+
+	for index, keyPath := range o.expectedRootKeys {
+		keyBytes, err := os.ReadFile(keyPath)
+		if err != nil {
+			return err
+		}
+
+		key, err := tuf.LoadKeyFromBytes(keyBytes)
```
adityasaky, line 48: can we use common.LoadPublicKey here?

internal/policy/policy.go
```diff
@@ -152,6 +152,27 @@ func LoadCurrentState(ctx context.Context, repo *git.Repository) (*State, error)
...
+	if err != nil {
+		return nil, err
+	}
+
+	firstEntry, _, err := rsl.GetFirstEntry(repo)
+	if err != nil {
+		return nil, err
+	}
+
+	allPolicyEntries, _, err := rsl.GetReferenceEntriesInRangeForRef(repo, firstEntry.ID, latestEntry.ID, PolicyRef)
```
adityasaky, line 168: we're really doing two passes at walking almost all the way back the RSL. If #344 lands first, this can just use rsl.GetFirstEntryForRef, right?

internal/repository/sync_test.go
```diff
@@ -228,10 +271,28 @@ func TestClone(t *testing.T) {
...
+		}
+		defer os.Chdir(currentDir) //nolint:errcheck
+
+		expectedPublicKey, err := tuf.LoadKeyFromBytes(targetsKeyBytes)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		_, err = Clone(context.Background(), remoteTmpDir, "", "", []*tuf.Key{expectedPublicKey})
+		assert.Equal(t, ErrExpectedRootKeysDoNotMatch, err)
```
adityasaky, line 296: -> assert.ErrorIs

internal/repository/sync_test.go
```diff
@@ -215,8 +222,44 @@ func TestClone(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		assert.Equal(t, remoteRSLRef.Hash(), localRSLRef.Hash())
+
+		localPolicyRef, err := repo.r.Reference(plumbing.ReferenceName(policy.PolicyRef), true)
+		if err != nil {
+			t.Fatal(err)
+		}
+		assert.Equal(t, remotePolicyRef.Hash(), localPolicyRef.Hash())
```
adityasaky, line 244: we should probably pull this out into a test helper...I think we've done this already, possibly in the gitinterface sync tests? we could likely copy that helper over.

adityasaky (commented):
I think this needs just another quick pass and possibly needs to land right after #344.

internal/repository/sync_test.go
neilnaveen: What part should be pulled out into a test helper?

internal/repository/sync_test.go
neilnaveen: Fixed

internal/policy/policy.go
neilnaveen: Will fix once #344 lands

internal/cmd/clone/clone.go
neilnaveen: Fixed

internal/cmd/clone/clone.go
neilnaveen: Fixed

internal/cmd/clone/clone.go
neilnaveen: Fixed

internal/repository/sync_test.go
adityasaky: the repeated assert local, remote refs match check. see: https://github.com/gittuf/gittuf/blob/main/internal/repository/common_test.go#L13-L29

neilnaveen: Fixed this, I did not have to clone the func, since they are in the same package

docs/cli/gittuf_clone.md
```diff
@@ -9,8 +9,9 @@ gittuf clone [flags]
 ### Options
 
 ```
-  -b, --branch string   specify branch to check out
-  -h, --help            help for clone
+  -b, --branch string      specify branch to check out
+  -h, --help               help for clone
+  -k, --root-key strings   path(s) of root keys to verify when cloning the repo.
```
flandweber, line 14: `-k` seems like a misleading shorthand, because in the rest of gittuf `-k` is used for a private key.

flandweber, line 14:
```suggestion
  -b, --branch string    specify branch to check out
  -h, --help             help for clone
  -k, --root-key paths   root keys to verify when cloning the repo.
```

flandweber (changes requested):
Thanks for your work, I really want to see this feature happen!

The things I marked are usibility nits as I'm not equipped to comment on the code.

I'd actually need some help on how to use this feature: 
which file is supposed to be reffered to by this flag?
I ran 
```
$ gittuf -- clone -k keys/root.pem repo repo-cloned
Error: open -----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAECmQonMqKLeOcKX4GbuHIy2zXFu+a
RWCz2Qpy/0rxe0H++iUj1YQUgEcvmjYf5rlGaGnDN4MOaaF0rCiWHPpXLA==
-----END PUBLIC KEY-----
: no such file or directory
```
and it looks like the code is looking for a file with the name of the content of the keyfile?

internal/cmd/clone/clone.go
```diff
@@ -19,14 +24,38 @@ func (o *options) AddFlags(cmd *cobra.Command) {
 		"",
 		"specify branch to check out",
 	)
+	cmd.Flags().StringSliceVarP(
+		&o.expectedRootKeys,
+		"root-key",
```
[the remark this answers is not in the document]
adityasaky: the CLI parser expects multiple invocations of the flag, so it'd be `--root-key <> --root-key <>` and so on.

docs/cli/gittuf_clone.md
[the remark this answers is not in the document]
adityasaky: yeah, this was flagged by another user elsewhere. I think updating this to `--root-key-path` is fine.

adityasaky: also a good point!

adityasaky: @neilnaveen can you check if cobra allows us to annotate the type for this?

@flandweber this is auto generated by cobra, so it uses the underlying datatype.

neilnaveen: Fixed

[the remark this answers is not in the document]
neilnaveen: fixed

neilnaveen:
@flandweber @adityasaky I was somwhat able to fix key loading, but currnetly it only works with either a gpg or fulcio key as the actual input, or a path to a PEM key, since it relies on 

```
func LoadPublicKey(key string) (*tuf.Key, error) {
	var keyObj *tuf.Key

	switch {
	case strings.HasPrefix(key, GPGKeyPrefix):
		fingerprint := strings.ToLower(strings.TrimSpace(strings.TrimPrefix(key, GPGKeyPrefix)))

		command := exec.Command("gpg", "--export", "--armor", fingerprint)
		stdOut, err := command.Output()
		if err != nil {
			return nil, err
		}

		keyObj, err = gpg.LoadGPGKeyFromBytes(stdOut)
		if err != nil {
			return nil, err
		}
	case strings.HasPrefix(key, FulcioPrefix):
		keyID := strings.TrimPrefix(key, FulcioPrefix)
		ks := strings.Split(keyID, "::")
		if len(ks) != 2 {
			return nil, fmt.Errorf("incorrect format for fulcio identity")
		}

		keyObj = &sslibsv.SSLibKey{
			KeyID:   keyID,
			KeyType: signerverifier.FulcioKeyType,
			Scheme:  signerverifier.FulcioKeyScheme,
			KeyVal: sslibsv.KeyVal{
				Identity: ks[0],
				Issuer:   ks[1],
			},
		}
	default:
		kb, err := os.ReadFile(key)
		if err != nil {
			return nil, err
		}

		keyObj, err = tuf.LoadKeyFromBytes(kb)
		if err != nil {
			return nil, err
		}
	}

	return keyObj, nil
}
```

And any key that must be loaded can only be a PEM key, since tuf.LoadKeyFromBytes can only load PEM keys

docs/cli/gittuf_clone.md
neilnaveen: I have fixed this, I did not use cobra, I just created a custom type, which made cobra display it as "paths"

adityasaky:
That's a good point, we overload the one type with both paths and scheme-based keys. I wonder if we need --root-key-path and --root-key (for gpg / fulcio). This would also apply on the signing side.

neilnaveen:
This feature would get quite confusing for the average user to use. If they have key type A, B, and C they should feed the file path in, or else they should feed the key itself. I think the easiest way to deal with this is just to go with @flandweber idea of cat'ing each key value in. Thoughts?

adityasaky:
I don't think $(cat <path>) is very user friendly. Also, for unencrypted keys, we don't want the contents to be in the shell history (if it records with this resolved).

neilnaveen:
Then would we be able to read gpg and fulcio keys from paths?

internal/cmd/clone/clone.go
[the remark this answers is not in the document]
flandweber: Makes sense, thanks!

flandweber:
> I don't think $(cat path) is very user friendly. 

I agree in some scenarios, while I see it as important in others to be able to pass a verifying command as a one-liner. Maybe complementary flags  `--root-key-path` and `--root-key` could solve this?

> Also, for unencrypted keys, we don't want the contents to be in the shell history (if it records with this resolved).

For public keys I don't see the problem

adityasaky:
> For public keys I don't see the problem

Agreed. However, I was thinking in terms of consistency for the signing flow as well.

internal/repository/sync.go
```diff
@@ -62,8 +67,33 @@ func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Reposito
...
 	}
 
+	sort.Slice(expectedRootKeys, func(i, j int) bool {
+		return expectedRootKeys[i].KeyID < expectedRootKeys[j].KeyID
+	})
+
 	repository := &Repository{r: r}
 
+	slog.Debug("Verifying if root keys are expected root keys...")
+	if len(expectedRootKeys) > 0 {
```
flandweber, line 77:
```suggestion
	if len(expectedRootKeys) > 0 {
	  slog.Debug("Verifying if root keys are expected root keys...")
```
should be switched as it's only verifying if pubkeys are passed

flandweber (commented):
> I was somwhat able to fix key loading, but currnetly it only works with either a gpg or fulcio key as the actual input, or a path to a PEM key

@neilnaveen  Thanks, loading the PEM key now doesn't throw an error.

However, the verification still seems to not be performed.
When switching the debug statement as suggested below, it does not appear when executing the following command: `gittuf --verbose clone --root-key-path keys/root.pem repo repo-cloned`, which indicates to me that the key is somehow not added to `expectedRootKeys`.

Please ping me if you're unable to reproduce so I can provide an example.

internal/cmd/clone/clone.go
```diff
@@ -3,12 +3,36 @@
...
+	"github.com/gittuf/gittuf/internal/cmd/common"
 	"github.com/gittuf/gittuf/internal/repository"
+	"github.com/gittuf/gittuf/internal/tuf"
 	"github.com/spf13/cobra"
 )
 
 type options struct {
-	branch string
+	branch           string
+	expectedRootKeys []string
```
flandweber, line 16:
```suggestion
	expectedRootKeys PathList
```

internal/cmd/clone/clone.go
```diff
@@ -19,14 +43,34 @@ func (o *options) AddFlags(cmd *cobra.Command) {
...
 		"specify branch to check out",
 	)
+	var rootKeys PathList
+	cmd.Flags().VarP(
+		&rootKeys,
+		"root-key-path",
+		"r",
+		"path(s) of root keys to verify when cloning the repo.",
+	)
+	o.expectedRootKeys = rootKeys
```
flandweber, line 53:
```suggestion
	cmd.Flags().VarP(
		&o.expectedRootKeys,
		"root-key-path",
		"r",
		"path(s) of root keys to verify when cloning the repo.",
	)
```

flandweber (commented):
found the problem. With the fixes the key loading should work.

internal/repository/sync.go
neilnaveen: Fixed

internal/cmd/clone/clone.go
neilnaveen: Fixed

internal/cmd/clone/clone.go
neilnaveen: Fixed

flandweber:
Thanks, lgtm!

internal/policy/policy.go

internal/cmd/clone/clone.go
```diff
@@ -3,12 +3,36 @@
...
 )
 
 type options struct {
-	branch string
+	branch           string
+	expectedRootKeys PathList
+}
+
+// PathList is a custom type to represent a list of paths
+type PathList []string
```
adityasaky, line 20: this applies to a bunch of commands, I think? Can we move this to cmd/common? Note: we can update the other commands to use this in a separate PR to keep PRs clean, but let's put it in cmd/common from the start.

adityasaky: Also, what does this mean for fulcio / gpg keys which may not be paths?

internal/cmd/clone/clone.go
```diff
@@ -19,14 +43,32 @@ func (o *options) AddFlags(cmd *cobra.Command) {
 		"",
 		"specify branch to check out",
 	)
+	cmd.Flags().VarP(
+		&o.expectedRootKeys,
+		"root-key-path",
```
adityasaky, line 48: I'd make this just root-key, because it may well be a gpg fingerprint or a fulcio identity.

internal/cmd/clone/clone.go
```diff
@@ -19,14 +43,32 @@ func (o *options) AddFlags(cmd *cobra.Command) {
...
+		"r",
```
adityasaky, line 49: Let's drop `VarP` to just `Var` and remove the shorthand. This is a command that directly shadows a git command, so I don't want us to use a flag shorthand that may eventually collide with git itself.

internal/cmd/clone/clone.go
```diff
@@ -19,14 +43,32 @@ func (o *options) AddFlags(cmd *cobra.Command) {
...
+		"path(s) of root keys to verify when cloning the repo.",
```
adityasaky, line 50: same q as above re path

internal/cmd/clone/clone.go
```diff
@@ -19,14 +43,32 @@ func (o *options) AddFlags(cmd *cobra.Command) {
...
+
+	expectedRootKeys := make([]*tuf.Key, len(o.expectedRootKeys))
+
+	for index, keyPath := range o.expectedRootKeys {
+		key, err := common.LoadPublicKey(keyPath)
+		if err != nil {
+			return err
+		}
+
+		expectedRootKeys[index] = key
```
adityasaky, line 68: Curious (ie this is fine), but does append() reallocate the underlying array before we actually hit the limit? Is that why you chose to use the index? I figured append would hold off on reallocating given `make` specifies the capacity, but I don't know the exact threshold it uses to reallocate. :smile:

internal/cmd/clone/clone.go
```diff
@@ -3,12 +3,36 @@
...
-	branch string
+	branch           string
+	expectedRootKeys PathList
+}
+
+// PathList is a custom type to represent a list of paths
+type PathList []string
+
+// String implements part of the pflag.Value interface.
+func (p *PathList) String() string {
```
adityasaky, line 23:
```suggestion
func (p PathList) String() string {
```

Since PathList is a slice, this doesn't need to be a pointer method to match the interface, if I'm not mistaken.

internal/cmd/clone/clone.go
```diff
@@ -3,12 +3,36 @@
...
+	return strings.Join(*p, ",")
+}
+
+// Set implements part of the pflag.Value interface.
+func (p *PathList) Set(value string) error {
+	*p = append(*p, value)
```
adityasaky, line 29: ah, i suppose we need the ** for this append to hold?

adityasaky (commented):
I think this is largely fine. My biggest comments are around the new custom flag type. I like the idea quite a bit, I'd just maybe make it more specific to keys, so its documentation can indicate it can be:
* a file path
* gpg fingerprint
* fulcio identity

adityasaky:
Also, please rebase to get #407 in.

internal/cmd/clone/clone.go
neilnaveen: We need it for the append values to propogate

internal/cmd/clone/clone.go
neilnaveen: Yes

internal/cmd/clone/clone.go
neilnaveen: Technicaly assigning by index is slightly faster, since append has a small, minute overhead of checking if the array has a high enough capacity to hold a new element, when append does not have to reallocate. Append does not reallocate before the capacity is fully used.

internal/cmd/clone/clone.go
neilnaveen: Fixed

internal/cmd/clone/clone.go
neilnaveen: Fixed

internal/cmd/clone/clone.go
neilnaveen: Fixed

internal/cmd/clone/clone.go
neilnaveen: I added a larger description in the expected root-key's description that gpg and fulcio keys should be passed in as the keys themselves.

neilnaveen:
@adityasaky Rebased over main

docs/cli/gittuf_clone.md
```diff
@@ -9,8 +9,9 @@ gittuf clone [flags]
 ### Options
 
 ```
-  -b, --branch string   specify branch to check out
-  -h, --help            help for clone
+  -b, --branch string    specify branch to check out
+  -h, --help             help for clone
+      --root-key paths   path(s) of root keys to verify when cloning the repo. For gpg or fulcio key's pass in the key itself instead of the path to the key
```
adityasaky, line 14: @lukpueh with your changes on ssh key support, we'll still be passing in a path to a key, right?

adityasaky: I think this help message needs more work.
1) we need something slightly more generic than `paths` for the type to account for gpg/fulcio
2) the "short" help msg is just too long, I wonder if renaming the type could make life easier

Perhaps the type can be something like `public-key` and the message can just focus on the root keys bit. Is it okay for the help on `public-key` itself to indicate it can be a path or a gpg fingerprint or fulcio identity?

cc @flandweber @nealmcb I believe both of you had thoughts about the original parameter type and help message, so your input is appreciated!

internal/cmd/clone/clone.go
adityasaky: I was wondering if there was an optimization for this, but I _think_ not. Admittedly I don't have any real experience writing / reading compiler code so... :smile:

internal/repository/sync_test.go
```diff
@@ -218,23 +189,59 @@ func TestClone(t *testing.T) {
...
+		defer os.Chdir(currentDir) //nolint:errcheck
+
+		publicKey, err := tuf.LoadKeyFromBytes(rootKeyBytes)
 		if err != nil {
 			t.Fatal(err)
 		}
-		assert.Equal(t, remoteRSLRef.Hash(), localRSLRef.Hash())
-		localPolicyRef, err := repo.r.Reference(plumbing.ReferenceName(policy.PolicyRef), true)
+
+		repo, err := Clone(context.Background(), remoteTmpDir, "", "", []*tuf.Key{publicKey})
```
adityasaky, line 218: could you add a test with multiple root keys?

adityasaky (changes requested):

docs/cli/gittuf_clone.md
```diff
@@ -9,8 +9,9 @@ gittuf clone [flags]
 ### Options
 
 ```
-  -b, --branch string   specify branch to check out
-  -h, --help            help for clone
+  -b, --branch string          specify branch to check out
+  -h, --help                   help for clone
+      --root-key public-keys   For ssh keys, pass in the path(s) of root keys to verify when cloning the repo
```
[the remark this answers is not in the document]
neilnaveen: Fixed

internal/cmd/common/common.go
```diff
@@ -22,6 +22,25 @@ const (
 	FulcioPrefix = "fulcio:"
 )
 
+// PathList is a custom type to represent a list of paths
```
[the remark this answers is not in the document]
neilnaveen: Fixed

internal/cmd/common/common.go
```diff
@@ -22,6 +22,25 @@ const (
...
+type PublicKeys []string
+
+// String implements part of the pflag.Value interface.
+func (p *PublicKeys) String() string {
+	return strings.Join(*p, ",")
```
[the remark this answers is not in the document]
neilnaveen: > How is this method used for other types? Such as the default ones? Should it in fact provide the value? Is this meant to be used for user facing messages? If so, I'd make the separator `, ` perhaps, with a space?
This is not actual used anywhere, it is just needed to satisfy the pflag.Value interface for cobra. I think it could potentaily be used in more complex cobra commands that I do not know about, and the default types have this implmented for them in the cobra library. 
> But IMO we need to figure this out, because a path on disk may not be very useful, unless this is primarily meant for local logging or so.
I think we should open up an issue on this, to finaly decide how the passing of keys into gittuf should work, and in what format.

internal/repository/sync.go
```diff
@@ -62,8 +67,34 @@ func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Reposito
...
+		rootKeys, err := state.GetRootKeys()
+		if err != nil {
+			return repository, errors.Join(ErrCloningRepository, err)
+		}
+
+		if len(rootKeys) != len(expectedRootKeys) {
+			return repository, ErrExpectedRootKeysDoNotMatch
+		}
+		for keyIndex := range rootKeys {
+			if !reflect.DeepEqual(rootKeys[keyIndex], expectedRootKeys[keyIndex]) {
```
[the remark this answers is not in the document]
neilnaveen: They were not, fixed.

internal/repository/sync_test.go
neilnaveen: I have replaced the single root key tests with multiple root key tests.

internal/repository/sync.go
```diff
@@ -62,8 +67,37 @@ func Clone(ctx context.Context, remoteURL, dir, initialBranch string) (*Reposito
 		return nil, errors.Join(ErrCloningRepository, err)
 	}
 
+	sort.Slice(expectedRootKeys, func(i, j int) bool {
```
adityasaky, line 70: for readability, please move this into the if len > 0 block below.

adityasaky (approved):
LGTM! Could you address the one comment below and clean up history?

internal/repository/sync.go
neilnaveen: Fixed

adityasaky (approved):

adityasaky:
Thanks @neilnaveen!

PR #313, opened 2024-02-28
tuf: Use fnmatch for delegation pattern matching

### LINKED ISSUE #309: Wildcard in rule pattern doesn't apply to subdirectories
### What happened?

gittuf currently uses glob-style patterns. This means that `file:foo/*` will apply to all files in the foo/ directory but not in the foo/bar/ directory. The glob-style patterns are based off the upstream TUF specification but they're not ideal for gittuf. Instead, I propose we switch our patterns to align with [gitignore](https://git-scm.com/docs/gitignore#_pattern_format). This should be a drop-in replacement. Note that with this change, `file:foo/*` will still not match other `/`, but we can at least support `file:foo/**`.

### Relevant log output

_No response_

### Code of Conduct

adityasaky:
Switch from glob style patterns that do not handle subdirectories to fnmatch. This commit includes tests for expected behavior with patterns so that we aren't caught unawares with changes to the fnmatch library.

Closes #309


adityasaky:
cc @JustinCappos @wlynch 

Please also suggest other cases we want to test for.

JustinCappos (commented):
Add more negative tests and those around situations where the match is in a file name / path.

For example: does *.txt match abc/foo.txtfile or allfiles.txt/foo?  I think the answer should be no in both cases

adityasaky:
> For example: does *.txt match abc/foo.txtfile or allfiles.txt/foo? I think the answer should be no in both cases

Yep, fnmatch would not match either of those.

adityasaky:
I added test cases for both of your examples (with some minor changes).

neilnaveen (commented):
LGTM, a couple more test cases

internal/tuf/tuf_test.go
```diff
@@ -145,3 +145,68 @@ func TestTargetsMetadataAndDelegations(t *testing.T) {
 		assert.Contains(t, delegations.Roles, d)
 	})
 }
+
+func TestDelegationMatches(t *testing.T) {
+	tests := map[string]struct {
+		patterns []string
+		target   string
+		expected bool
+	}{
```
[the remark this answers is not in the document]
adityasaky: thanks! I added these except for the one with whitespace. I'm actually not sure whether we should be stripping that everytime ourselves and so on to account for potential user error. The CLI parser shouldn't be passing leading/trailing whitespace but...

cc @wlynch @JustinCappos

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/sigstore/sigstore v1.8.3
 	github.com/spf13/cobra v1.8.0
 	github.com/stretchr/testify v1.9.0
+	github.com/tzvetkoff-go/fnmatch v0.0.0-20220210160758-879480b5e662
```
wlynch, line 19: why this dependency instead of similar ones that appear to be more widely used?

- https://github.com/bmatcuk/doublestar
- https://github.com/danwakefield/fnmatch

adityasaky: doublestar is _way_ nicer, switched!

adityasaky: also evaluating danwakefield/fnmatch.

adityasaky:
Latest commit tries danwakefield/fnmatch in place of doublestar. doublestar seems nice with things like being able to validate a pattern when a user adds it and so on but fnmatch is closer (possibly equivalent by my reading) to git's pathspec used in `git add <>`.

wlynch (approved):

PR #320, opened 2024-03-27
Added basic authentication attestation support

neilnaveen:
- In this current feature implementation, anyone can add an authentication attestation, with the pusher being anyone. Without evidence, as defined in the design document https://github.com/gittuf/gittuf/blob/main/docs/design-document.md#authentication-evidence-attestations, this feature is fraught with security holes.
- I do not know what format the evidence should take; I was thinking about somthing like a push event ID (for example, https://api.github.com/events) where verification could call the GitHub API, go through the push events, and try to find this ID. If found, it would check if the pusher is the pusher that the creator of the attestation says it is. The problem with this approach is that Github only allows access to the last 90 days of Push Events, so if we were to make a recording like this. If verification were run two years later, we would not be able to find that event, and the tool would mark it as failing verification, even if it were a valid attestation.

### DIFF
--- a/internal/attestations/attestations.go
+++ b/internal/attestations/attestations.go
@@ -16,6 +16,7 @@ import (
 const (
 	Ref                                  = "refs/gittuf/attestations"
 	referenceAuthorizationsTreeEntryName = "reference-authorizations"
+	authenticationEvidenceTreeEntryName  = "authentication-evidence"
 	initialCommitMessage                 = "Initial commit"
 	defaultCommitMessage                 = "Update attestations"
 )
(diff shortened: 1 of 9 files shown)


neilnaveen:
Possible Solution in Place of Evidence

- Instead of using evidence to prove that a push event happened, we can use gittuf to tell us if the pusher is the pusher or not, depending on who we trust
- Essentially, we are verifying the commit to the change to the attestation. This means we can set up a rule in gittuf that only allows certain users to create evidence, just as gittuf protects other refs. The verification is done through VerifyCommit.
- We can only allow certain people or bots to create these changes; for example, if we created a Github bot that automatically created these evidence events, we could make a rule for the attestations ref saying that only the bot can make changes. While running gittuf verify ref would fail, verifying the commits and creating authentication evidence attestations would succeed if everything went well. 
-This also circumvents the problems we encounter when using the GitHub API or other code hosts since they only maintain push logs for public use for a limited time. Creating verification for each of these tools would be painful and not as customizable as this approach; we do not even need to trust a platform; instead, a user could set themselves or a group as the only people able to create these authorization evidence attestations

neilnaveen:
This solution is not technically replacing evidence; it is simply integrating it into the gittuf's existing workflows, where the evidence is the signature of the author of the attestation.

neilnaveen:
:eyes: @adityasaky

neilnaveen:
I have fixed the faulty test, but for this to work, fetching Non-gittuf RSL entries now also fetches attestation entries as well, but does not include any others.

Issue #231, opened 2023-12-18
Roadmap: Developer teams and multiple hats


adityasaky:
gittuf currently only enumerates actors using signing keys. To improve policy usability, two enhancements are in order:

1) Support for teams: it must be possible to group actors together (team `ios-devs` is composed of `alice@example.org`, `bob@example.org`, etc.) and use the team definition in rules / delegations.
2) Support for hats: the same actor may perform different actions wearing different hats. For example, an actor who's part of the dev team and the security team may have different privileges based on the hat they're currently wearing. gittuf currently does not support this perfectly outside of asking the same individual to use different keys for different hats, but it'd be neat to improve this aspect.

adityasaky:
Related: https://github.com/theupdateframework/taps/blob/master/tap3.md

PR #326, opened 2024-04-04
[WIP] Add support for teams

patzielinski:
This PR addresses the teams portion of #231 by restructuring the underlying TUF metadata and updating the verification workflow to support teams. _This is still a work-in-progress._

**EDIT**: The teams functionality is at a prototype state. Feel free to pull this PR down and tinker with setting up teams!

### DIFF
--- a/internal/tuf/tuf.go
+++ b/internal/tuf/tuf.go
@@ -72,9 +72,18 @@ func calculateKeyID(k *Key) (string, error) {
 	return hex.EncodeToString(digest[:]), nil
 }
 
+// TEAMS
 // Role records common characteristics recorded in a role entry in Root metadata
 // and in a delegation entry.
 type Role struct {
+	Name      string   `json:"rolename"`
+	KeyIDs    []string `json:"keyids"`
+	Threshold int      `json:"threshold"`
+}
+
+// LegacyRole fulfills the same role (heh) as Role above, but aligns with the
+// old, pre-teams metadata used by gittuf.
+type LegacyRole struct {
 	KeyIDs    []string `json:"keyids"`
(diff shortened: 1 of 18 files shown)


internal/tuf/tuf.go
```diff
@@ -206,10 +208,12 @@ func (d *Delegation) Matches(target string) bool {
 // Delegation defines the schema for a single delegation entry. It differs from
 // the standard TUF schema by allowing a `custom` field to record details
 // pertaining to the delegation.
+// Updated to comply with TUF's TAP 3
```
adityasaky, line 211: For backwards compat, I suggest making this a separate struct. I want to introduce a `gittufVersion` field in metadata that we'll use to determine what the metadata looks like. Some repetition is fine for that.

adityasaky: Or actually, we introduce a _separate_ struct for "legacy" (pre-teams) metadata so we can continue verifying them. And when a legacy metadata is loaded / manipulated for whatever reason, can we automatically convert it into this new schema so that over time we get even legacy metadata moved over? We obviously can't ever drop the legacy metadata struct/schema for verification from the start scenarios, but at least it'll simplify the logic of enabling teams when it's not actually used and so on.

adityasaky: Let's make it its own PR. ATM, I think we could avoid tying it to a specific gittuf version and just have incrementing integers for the metadata schema, but maybe that introduces more complications. Can you sketch something out?

adityasaky:
Closing this. tuf metadata has evolved significantly and we have new interfaces we can leverage to define teams.

PR #335, opened 2024-04-09
gitinterface: Make user.signingkey optional

adityasaky:
Now, user.signingkey is required only when the signing format is ssh. I suspect we can loosen this too?


wlynch (approved):

adityasaky:
Removed one test artifact (git config) that wasn't needed anymore.

PR #336, opened 2024-04-10
Switch authorizations to use merge tree target

adityasaky:
This PR updates the authorization attestation to use the merge tree as the target for the attestation. This better allows for ahead of time merges. For example, if the main branch has commit A and feature branch has commit B, previously, we required the merge to be a fast forward merge in the context of the authorization attestation. Specifically, the attestation recorded commit B as the target. Now, the target is the result of `git merge-tree <commitA> <commitB>`.

### DIFF
--- a/docs/cli/gittuf_dev_authorize.md
+++ b/docs/cli/gittuf_dev_authorize.md
@@ -9,6 +9,7 @@ gittuf dev authorize [flags]
 ### Options
 
 ```
+  -f, --from-ref string      ref to authorize merging changes from
   -h, --help                 help for authorize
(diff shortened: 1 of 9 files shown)


adityasaky:
Two things to note:
a) go-git does not support merges (AIUI, the problem is a reliable three way merge library for Go), so this uses the Git binary. I haven't added a test but I have a dev branch that takes over #286 for more thorough support for the Git binary.
b) The feature ref addition also relies on the RSL. So the commit used for the merge in the feature ref is identified via the RSL.

wlynch (approved):

PR #337, opened 2024-04-10
Added verification before recording RSL entries.

neilnaveen:
- This is a simple way to address recovery mechanism M1 outlined in the design doc. If we do not allow creation of RSL if their is an invalid RSL entry, this means we do not need to cut and stitch back the commit histor, nor would we have to create a new reverting commit.
- Even if a bad actor created an RSL entry manually without gittuf, we can assume that all entries after that entry are bad, since they all would have failed the verify-ref check rsl entry creation runs.


internal/cmd/rsl/record/record.go
```diff
@@ -12,12 +14,14 @@ type options struct{}
...
 func (o *options) AddFlags(_ *cobra.Command) {}
 
-func (o *options) Run(_ *cobra.Command, args []string) error {
+func (o *options) Run(cmd *cobra.Command, args []string) error {
 	repo, err := repository.LoadRepository()
 	if err != nil {
 		return err
 	}
-
+	if err := repo.VerifyRef(cmd.Context(), args[0], false); err != nil {
```
adityasaky, line 22: verification relies on the RSL entry, no? IMO, this should create a signed RSL entry (to enable verification of current change), verify, and _remove the RSL entry if verification fails_.

neilnaveen: Makes sense, I was thinking that we should stop the creation of RSL entries, if their is an invlaid entry already in the RSL

neilnaveen: I have changed this to revoke the current RSL entry, by skipping it with an annotation entry

internal/cmd/rsl/record/record.go
```diff
@@ -12,20 +14,32 @@ type options struct{}
...
 		return err
 	}
 
-	return repo.RecordRSLEntryForReference(args[0], true)
+	if err := repo.RecordRSLEntryForReference(args[0], true); err != nil {
+		return err
+	}
+
+	if err := repo.VerifyRef(cmd.Context(), args[0], false); err != nil {
+		if err := repo.InvalidateLatestRSLEntry(args[0], true); err != nil {
```
adityasaky, line 28: Hm, I think we want to rethink this. The RSL reference entry represents the intent to _push_ and make the change to the branch (say main) available to other users. At this point, the entry's been created but no one other than this local client is aware of it. So if the entry is invalid, we should actually `gitinterface.Reset` the RSL branch and remove that entry altogether.

neilnaveen:
The test are failing for InvalidateLatestRSLEntry, because of gitinterface.ResetCommit(), and the error is that the HEAD ref is not found. How should I fix this?

internal/repository/rsl.go
```diff
@@ -90,6 +90,20 @@ func (r *Repository) RecordRSLAnnotation(rslEntryIDs []string, skip bool, messag
 	return rsl.NewAnnotationEntry(rslEntryHashes, skip, message).Commit(r.r, signCommit)
 }
 
+func (r *Repository) InvalidateLatestRSLEntry(ref string) error {
```
adityasaky, line 93: does this need ref? This is only used when invalidating the absolute latest entry, right? Also, you have to set HEAD to `main` or whatever explicitly, which is likely the issue here...

neilnaveen: I have fixed this, it turns out I cannnot use `gitinterface.ResetCommit`, I have to use `repo.r.Storer.SetReference`, since reset commit does not actualy change the tip of a ref, it just moves the HEAD pointer, so when I query the latest RSL entry, I get the invalid commit

internal/cmd/rsl/record/record.go
neilnaveen: Fixed

internal/cmd/rsl/record/record.go
neilnaveen: This was a old comment, at the moment, InvalidateRSLEntry removes the last commit

internal/repository/rsl.go
```diff
@@ -90,6 +90,23 @@ func (r *Repository) RecordRSLAnnotation(rslEntryIDs []string, skip bool, messag
...
+	if err != nil {
+		return err
+	}
+
+	v, err := rsl.GetParentForEntry(r.r, latestEntry)
+	if err != nil {
+		return err
+	}
+
+	return r.r.Storer.SetReference(plumbing.NewHashReference(rsl.Ref, v.GetID()))
```
adityasaky, line 107: gitinterface.ResetCommit?

adityasaky: / ResetDueToError

adityasaky: Hm, that may have issues because it checks out the specified ref. This might be easier with the git binary support :thinking:

internal/cmd/dev/rslrecordat/rslrecordat.go
```diff
@@ -47,7 +47,18 @@ func (o *options) Run(_ *cobra.Command, args []string) error {
...
-	return repo.RecordRSLEntryForReferenceAtTarget(args[0], o.targetID, signingKeyBytes)
+	if err := repo.RecordRSLEntryForReferenceAtTarget(args[0], o.targetID, signingKeyBytes); err != nil {
+		return err
+	}
+
+	if err := repo.VerifyRef(cmd.Context(), args[0], false); err != nil {
+		if err := repo.InvalidateLatestRSLEntry(); err != nil {
+			return err
+		}
+		return fmt.Errorf("verifcation of reference to be record failed, creation of RSL entry is not possible, err = %w", err)
```
adityasaky, line 58:
```suggestion
		return fmt.Errorf("policy verification failed for ...: %w", err)
```

The `...` could possibly include the ref and the targetID at which verification failed.

internal/cmd/rsl/record/record.go
```diff
@@ -12,20 +14,32 @@ type options struct{}
...
-	return repo.RecordRSLEntryForReference(args[0], true)
+	if err := repo.RecordRSLEntryForReference(args[0], true); err != nil {
+		return err
+	}
+
+	if err := repo.VerifyRef(cmd.Context(), args[0], false); err != nil {
+		if err := repo.InvalidateLatestRSLEntry(); err != nil {
+			return err
+		}
+		return fmt.Errorf("verifcation of reference to be record failed, creation of RSL entry is not possible, err = %w", err)
```
adityasaky, line 31: same thing here

internal/cmd/rsl/record/record.go
```diff
@@ -12,20 +14,32 @@ type options struct{}
...
+
+	return nil
 }
 
 func New() *cobra.Command {
 	o := &options{}
 	cmd := &cobra.Command{
 		Use:               "record",
 		Short:             "Record latest state of a Git reference in the RSL",
+		Long:              "Record latest state of a Git reference in the RSL. If the entry fails full ref verification on the recorded ref, the rsl entry is skipped",
```
adityasaky, line 42: this needs to be updated. Also note that RSL is always all caps in docs.

internal/cmd/dev/rslrecordat/rslrecordat.go
neilnaveen: Fixed

internal/cmd/rsl/record/record.go
neilnaveen: Fixed

internal/cmd/rsl/record/record.go
neilnaveen: Fixed

internal/repository/rsl.go
neilnaveen: I think it should currently be left like this, since we should think about the real functionality of gitinterface.ResetCommit, should we be moving the HEAD pointer of the reference pointer?

adityasaky (changes requested):
I think this is getting very close.

internal/cmd/rsl/record/record.go
```diff
@@ -12,20 +14,32 @@ type options struct{}
...
 	}
 
-	return repo.RecordRSLEntryForReference(args[0], true)
+	if err := repo.RecordRSLEntryForReference(args[0], true); err != nil {
+		return err
+	}
+
+	if err := repo.VerifyRef(cmd.Context(), args[0], false); err != nil {
+		if err := repo.InvalidateLatestRSLEntry(); err != nil {
+			return err
```
adityasaky, line 29: could you wrap this error so we know why / when this happened?

adityasaky:
Looks good on the demo :)

We have to update the demo text when this is released. Also, please rebase this PR so it's easier to test in light of policy-staging and so on.

image

adityasaky:
Another thought: we should probably have a `--no-verify` flag for rsl record that bypasses this verification. Full verification can start becoming expensive and I imagine a user might want to `git config --local gittuf.verifyrsl false` which in turn tells the transport (#411) to add the `--no-verify` flag so we don't make pushes a lot more expensive. Just adding the flag in this PR would be nice, the rest can come later as things begin to connect!

internal/cmd/rsl/record/record.go
neilnaveen: Fixed

internal/cmd/rsl/record/record.go
```diff
@@ -12,20 +14,32 @@ type options struct{}
...
+
+	return nil
 }
 
 func New() *cobra.Command {
 	o := &options{}
 	cmd := &cobra.Command{
 		Use:               "record",
 		Short:             "Record latest state of a Git reference in the RSL",
+		Long:              "Record latest state of a Git reference in the RSL. If the entry fails full ref verification on the recorded ref, the RSL entry is removed",
```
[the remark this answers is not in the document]
neilnaveen: Fixed

neilnaveen:
I have rebased the PR with main, and tests pass

neilnaveen:
> Another thought: we should probably have a `--no-verify` flag for rsl record that bypasses this verification. Full verification can start becoming expensive and I imagine a user might want to `git config --local gittuf.verifyrsl false` which in turn tells the transport (#411) to add the `--no-verify` flag so we don't make pushes a lot more expensive. Just adding the flag in this PR would be nice, the rest can come later as things begin to connect!
> 
> Also a `--from-entry` flag that gets passed on to verifyref would be nice for similar reasons when we can identify since when we want to perform local verification.

Wouldn't this defeat the purpose of verification before creating RSL entries, since one of the main motivators is that if no RSL entries by good actors are made that break verification, this would mean that we would never have an RSL history with good and bad RSL entries interspersed with each other? Instead, it would first be good RSL entries; after those, only bad RSL entries could exist. If we add these flags, the history could become corrupted, and performing "surgery" on the RSL with bad entries throughout the RSL could be hard or impossible.

adityasaky:
Verification would still happen when someone else fetches the repo _before they add their entry_, we'd offload the cost there with `--from-entry`. With that said, adding `--from-entry` here would help us get the same benefit here and we can reevaluate --no-verify later.

internal/cmd/rsl/record/record.go
```diff
@@ -3,29 +3,58 @@
...
+		err = repo.VerifyRef(cmd.Context(), args[0], false)
+	} else {
+		err = repo.VerifyRefFromEntry(cmd.Context(), args[0], o.fromEntry)
+	}
+
+	if err != nil {
+		if err := repo.InvalidateLatestRSLEntry(); err != nil {
+			return fmt.Errorf("removal of the RSL entry failed for ...: %w", err)
+		}
+		return fmt.Errorf("policy verification failed for ...: %w", err)
```
adityasaky, line 47: the `...` was to say "fill in the blanks" about the ref that failed and so on. :smile:

neilnaveen: Fixed; I was confused about what they were for!

internal/repository/rsl.go
```diff
@@ -91,6 +91,23 @@ func (r *Repository) RecordRSLAnnotation(rslEntryIDs []string, skip bool, messag
 	return rsl.NewAnnotationEntry(rslEntryHashes, skip, message).Commit(r.r, signCommit)
 }
 
+// InvalidateLatestRSLEntry is used when the latest RSL entry is invalid.
+// This does not skip the RSL entry with an annotation, instead it resets
+// the ref, rolling it back by one commit.
+func (r *Repository) InvalidateLatestRSLEntry() error {
```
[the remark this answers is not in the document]
adityasaky: yes, that's a good call!

internal/cmd/rsl/record/record.go
```diff
@@ -3,29 +3,58 @@
...
+	}
+	return nil
 }
 
 func New() *cobra.Command {
 	o := &options{}
 	cmd := &cobra.Command{
 		Use:               "record",
 		Short:             "Record latest state of a Git reference in the RSL",
+		Long:              "Record latest state of a Git reference in the RSL. Policy verification must succeed for the RSL entry to be created",
```
adityasaky, line 57:
```suggestion
		Long:              "Record latest state of a Git reference in the RSL. Policy verification must succeed for the RSL entry to be created.",
```

I'd actually make this a bit more verbose since this is the "long" message.

adityasaky:
@neilnaveen this LGTM, could you update the docs a bit more and rename the API as Patrick suggested? Thanks!

internal/repository/rsl.go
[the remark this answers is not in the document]

internal/cmd/rsl/record/record.go
neilnaveen: Fixed. I don't think there is much more to say about it. Is there somthing else you would like me to add?

internal/cmd/rsl/record/record.go
```diff
@@ -3,29 +3,58 @@
...
+	}
+	return nil
 }
 
 func New() *cobra.Command {
 	o := &options{}
 	cmd := &cobra.Command{
 		Use:               "record",
 		Short:             "Record latest state of a Git reference in the RSL",
+		Long:              "Record latest state of a Git reference in the RSL. Policy verification must succeed for the RSL entry to be created. The only ref that is verified is the ref being recorded",
```
adityasaky, line 57:
```suggestion
```

internal/cmd/rsl/record/record.go
```diff
@@ -3,29 +3,58 @@
...
+		return fmt.Errorf("policy verification failed for %s: %w", args[0], err)
+	}
+	return nil
 }
 
 func New() *cobra.Command {
 	o := &options{}
 	cmd := &cobra.Command{
 		Use:               "record",
 		Short:             "Record latest state of a Git reference in the RSL",
```
adityasaky, line 56:
```suggestion
		Short:             "Record latest state of a Git reference in the RSL after verifying against gittuf policy",
```

adityasaky: `Long` may not be needed, as you point out, there isn't much to say.

neilnaveen: Fixed

internal/cmd/rsl/record/record.go
neilnaveen: Fixed

internal/cmd/dev/rslrecordat/rslrecordat.go
```diff
@@ -34,9 +35,16 @@ func (o *options) AddFlags(cmd *cobra.Command) {
 		"path to PEM encoded SSH or GPG signing key",
 	)
 	cmd.MarkFlagRequired("signing-key") //nolint:errcheck
+
+	cmd.Flags().StringVar(
+		&o.fromEntry,
+		"from-entry",
+		"",
+		fmt.Sprintf("perform verification from specified RSL entry (developer mode only, set %s=1)", dev.DevModeKey),
```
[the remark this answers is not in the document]
neilnaveen: Fixed

adityasaky:
@neilnaveen could you `make generate` and lint? Also, clean up commits, please? Thanks!

neilnaveen:
@adityasaky Fixed the commit history, docs, and lint issues

internal/repository/rsl_test.go
```diff
@@ -239,6 +239,62 @@ func TestRecordRSLAnnotation(t *testing.T) {
 	assert.True(t, annotation.Skip)
 }
 
+func TestInvalidateLatestRSLEntry(t *testing.T) {
```
adityasaky, line 242: func name

internal/repository/rsl_test.go
```diff
@@ -239,6 +239,62 @@ func TestRecordRSLAnnotation(t *testing.T) {
...
+	latestEntry, err := rsl.GetLatestEntry(repo.r)
+	if err != nil {
+		t.Fatal(err)
+	}
+	latestRefEntry, ok := latestEntry.(*rsl.ReferenceEntry)
+	if !ok {
+		t.Fatal(fmt.Errorf("invalid entry type"))
+	}
+	assert.Equal(t, refName, latestRefEntry.RefName)
+	assert.Equal(t, firstTestHash, latestRefEntry.TargetID)
```
adityasaky, line 295: can you assert that latestRefEntry.ID is as expected? This test would theoretically pass with a different RSL entry that has the same contents.

adityasaky (commented):
Two minor things I missed before, then this is really good to go.

internal/repository/rsl_test.go
neilnaveen: Fixed

internal/repository/rsl_test.go
neilnaveen: I have added the check. I have concatenated it down into `assert.Equal(t, latestGoodEntry, latestRefEntry)`. If this is not intuitive, I can roll the changes back to check all elements of the RSL entry separately.

adityasaky:
I suspect a rebase is needed now that #372 has landed.

neilnaveen:
I have rebased with the main, but the tests are failing, with the same problem as before. I am using `r.r.ResetDueToError`, which does not set the ref tip to the given commit. What should I use to change the ref's tip now that we are not using go-git for the underlying repository?

adityasaky:
> I am using r.r.ResetDueToError, which does not set the ref tip to the given commit.

sorry, I'm a bit confused. does not?

adityasaky:
I need to double check if the docstring is fixed in one of my pending PRs but is https://github.com/gittuf/gittuf/blob/main/internal/gitinterface/utils.go#L72-L80 not an option?

neilnaveen:
I tried to use r.r.ResetDueToError, but after querying the latest RSL entry after resetting, it stayed the same

adityasaky:
weird, can you dig in a bit?

neilnaveen:
@adityasaky, the problem was not in `r.r.ResetDueToError`; I was resetting the ref to the ref the latest RSL entry was pointing to instead of the actual RSL ref. I have fixed the issue

internal/repository/rsl_test.go
```diff
@@ -211,20 +211,7 @@ func TestRollbackLatestRSLEntry(t *testing.T) {
...
-	treeWithContentsID, err := treeBuilder.WriteRootTreeFromBlobIDs(map[string]gitinterface.Hash{"README.md": blobID})
-	if err != nil {
-		t.Fatal(err)
-	}
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	if _, err := repo.r.Commit(treeWithContentsID, refName, "Second commit", false); err != nil {
+	if _, err := repo.r.Commit(emptyTreeHash, refName, "Second commit", false); err != nil {
```
adityasaky, line 214: could you explain the change in test here, sorry?

neilnaveen: I realized that I did not need to actual commit real data, so I thougth this would make it easier to understand

neilnaveen:
@adityasaky @patzielinski I have rebased with main and cleaned up the commit history

internal/cmd/dev/rslrecordat/rslrecordat.go
```diff
@@ -56,14 +65,31 @@ func (o *options) Run(_ *cobra.Command, args []string) error {
 		return err
 	}
 
-	return repo.RecordRSLEntryForReferenceAtTarget(args[0], o.targetID, signingKeyBytes, rslopts.WithOverrideRefName(o.dstRef))
+	if err := repo.RecordRSLEntryForReferenceAtTarget(args[0], o.targetID, signingKeyBytes, rslopts.WithOverrideRefName(o.dstRef)); err != nil {
+		return err
+	}
+
+	if o.fromEntry == "" {
```
adityasaky, line 72: fromEntry should be marked dev only like its verify-ref counterpart.

adityasaky: oh nvm, this entire command is dev only.

internal/cmd/rsl/record/record.go
```diff
@@ -20,22 +24,44 @@ func (o *options) AddFlags(cmd *cobra.Command) {
...
 	if err != nil {
 		return err
 	}
 
-	return repo.RecordRSLEntryForReference(args[0], true, rslopts.WithOverrideRefName(o.dstRef))
+	if err := repo.RecordRSLEntryForReference(args[0], true, rslopts.WithOverrideRefName(o.dstRef)); err != nil {
+		return err
+	}
+
+	if o.fromEntry == "" {
```
adityasaky, line 45: Here, though, we want to flag this as dev only

internal/repository/rsl.go
```diff
@@ -139,6 +139,34 @@ func (r *Repository) RecordRSLAnnotation(rslEntryIDs []string, skip bool, messag
...
 
+// RollbackLatestRSLEntry is used when the latest RSL entry is invalid.
+// This does not skip the RSL entry with an annotation, instead it resets
+// the ref, rolling it back by one commit.
+func (r *Repository) RollbackLatestRSLEntry() error {
+	latestEntry, err := rsl.GetLatestEntry(r.r)
+	if err != nil {
+		return err
+	}
+	_, ok := latestEntry.(*rsl.ReferenceEntry)
```
adityasaky, line 150:
```suggestion
	if _, isReferenceEntry := latestEntry.(*rsl.ReferenceEntry); !isReferenceEntry {
```
and so on

internal/repository/rsl.go
```diff
@@ -139,6 +139,34 @@ func (r *Repository) RecordRSLAnnotation(rslEntryIDs []string, skip bool, messag
...
+		return err
+	}
+
+	errMustRollback := fmt.Errorf("the current RSL entry fails verification, rolling back changes")
+
+	if err := r.r.ResetDueToError(errMustRollback, rsl.Ref, v.GetID()); !errors.Is(err, errMustRollback) {
+		return err
+	}
+
+	return nil
```
adityasaky, line 167: isn't this unreachable? I think r.r.ResetDueToError always returns an error, if nothing else, the `cause` we pass in.

internal/repository/rsl.go
```diff
@@ -139,6 +139,34 @@ func (r *Repository) RecordRSLAnnotation(rslEntryIDs []string, skip bool, messag
...
+	if !ok {
+		return fmt.Errorf("cannot rollback when the latest RSL entry is an annnoation entry")
+	}
+
+	v, err := rsl.GetParentForEntry(r.r, latestEntry)
+	if err != nil {
+		return err
+	}
+
+	errMustRollback := fmt.Errorf("the current RSL entry fails verification, rolling back changes")
```
adityasaky, line 161: let's use a named error with errors.New for this one

adityasaky:
Looking at my last review, I think this isn't too far off. @neilnaveen could you take a pass and we can land this? I think it'd go really well with the rsl recording in the gittuf transport too.

adityasaky:
@neilnaveen nudge! Can any of us step in and finish implementing this?

PR #341, opened 2024-04-13
Store GitHub PR data as an attestation

adityasaky:
In developer mode, this records a GitHub PR's information as an attestation (using the GitHub API).

Note: I want to examine and incorporate work by @neilnaveen from #320 as well.

### DIFF
--- a/internal/attestations/attestations.go
+++ b/internal/attestations/attestations.go
@@ -14,10 +14,11 @@ import (
 )
 
 const (
-	Ref                                  = "refs/gittuf/attestations"
-	referenceAuthorizationsTreeEntryName = "reference-authorizations"
-	initialCommitMessage                 = "Initial commit"
-	defaultCommitMessage                 = "Update attestations"
+	Ref                                        = "refs/gittuf/attestations"
+	referenceAuthorizationsTreeEntryName       = "reference-authorizations"
+	githubPullRequestAttestationsTreeEntryName = "github-pull-requests"
+	initialCommitMessage                       = "Initial commit"
+	defaultCommitMessage                       = "Update attestations"
 )
 
 var ErrAttestationsExist = errors.New("cannot initialize attestations namespace as it exists already")
@@ -50,9 +51,16 @@ type Attestations struct {
 	// `<ref-path>/<from-id>-<to-id>`, where `ref-path` is the absolute ref path
 	// such as `refs/heads/main` and `from-id` and `to-id` determine how the ref
 	// in question moved. For example, the key
-	// `refs/heads/main/<commit-A>-<commit-B>` indicates the authorization is
-	// for the action of moving `refs/heads/main` from `commit-A` to `commit-B`.
+	// `refs/heads/main/<commit-A>-<tree-B>` indicates the authorization is
+	// for the action of moving `refs/heads/main` from `commit-A` to a commit
+	// with `tree-B`.
 	referenceAuthorizations map[string]plumbing.Hash
+
+	// githubPullRequestAttestations maps information about the GitHub pull
+	// request for a commit and branch. The key is a path of the form
+	// `<ref-path>/<commit-id>`, where `ref-path` is the absolute ref path, and
+	// `commit-id` is the ID of the merged commit.
+	githubPullRequestAttestations map[string]plumbing.Hash
 }
 
 // LoadCurrentAttestations inspects the repository's attestations namespace and
@@ -94,10 +102,16 @@ func LoadAttestationsForEntry(repo *git.Repository, entry *rsl.ReferenceEntry) (
 		return &Attestations{}, nil
 	}
 
-	var authorizationsTreeID plumbing.Hash
+	var (
+		authorizationsTreeID     plumbing.Hash
+		githubPullRequestsTreeID plumbing.Hash
+	)
+
 	for _, e := range attestationsRootTree.Entries {
 		if e.Name == referenceAuthorizationsTreeEntryName {
 			authorizationsTreeID = e.Hash
+		} else if e.Name == githubPullRequestAttestationsTreeEntryName {
+			githubPullRequestsTreeID = e.Hash
 		}
 	}
 
@@ -106,9 +120,22 @@ func LoadAttestationsForEntry(repo *git.Repository, entry *rsl.ReferenceEntry) (
 		return nil, err
 	}
 
-	attestations := &Attestations{referenceAuthorizations: map[string]plumbing.Hash{}}
+	githubPullRequestsTree, err := gitinterface.GetTree(repo, githubPullRequestsTreeID)
+	if err != nil {
+		return nil, err
+	}
+
+	attestations := &Attestations{
+		referenceAuthorizations:       map[string]plumbing.Hash{},
+		githubPullRequestAttestations: map[string]plumbing.Hash{},
+	}
 
 	attestations.referenceAuthorizations, err = gitinterface.GetAllFilesInTree(authorizationsTree)
+	if err != nil {
+		return nil, err
+	}
+
+	attestations.githubPullRequestAttestations, err = gitinterface.GetAllFilesInTree(githubPullRequestsTree)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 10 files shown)


internal/attestations/attestations.go
```diff
@@ -50,9 +51,16 @@ type Attestations struct {
...
 	// in question moved. For example, the key
-	// `refs/heads/main/<commit-A>-<commit-B>` indicates the authorization is
-	// for the action of moving `refs/heads/main` from `commit-A` to `commit-B`.
+	// `refs/heads/main/<commit-A>-<tree-B>` indicates the authorization is
+	// for the action of moving `refs/heads/main` from `commit-A` to a commit
+	// with `tree-B`.
 	referenceAuthorizations map[string]plumbing.Hash
+
+	// githubPullRequestAttestations maps information about each GitHub pull
+	// request. The key is a path of the form `<ref-path>/<from-id>-<to-id>`,
```
wlynch, line 60: I don't know if this makes sense for pull requests 🤔 

`<ref-path>`: Is this the base ref or the PR ref? The changes are being made on the PR ref, but what if this ref doesn't exist in the base repo?

`<from-id>`: How do you see the from-id being used? I'm a bit worried that this might mirror the "Require branches to be up to date before merging" behavior for branch protections, which many repos end up turning off. Would it make more sense to key attestation off the PR commit and perform any base commit checks from metadata stored in the attestation?

Also idk how often this happens in practice, but will this be problematic if there is a collision? I think GitHub does try to get you to reuse PRs if there's already one open, but you can have multiple closed pull requests that point to the same refs/commits (e.g. https://github.com/wlynch/test/pull/29, https://github.com/wlynch/test/pull/30). An attestation for the closed PR could bleed into the open one.

internal/github/github.go
```diff
@@ -0,0 +1,95 @@
...
+func getClient() *http.Client {
+	if client == nil {
+		client = &http.Client{Timeout: timeout}
+	}
+
+	return client
+}
+
+func GetPullRequest(ctx context.Context, repository, pullRequestID string) ([]byte, error) {
+	endpoint := fmt.Sprintf("%s/%s/pulls/%s", baseEndpoint, repository, pullRequestID)
```
wlynch, line 30: I've used https://pkg.go.dev/github.com/google/go-github/v61 for a bunch of projects - will probably be better than doing type casting

internal/cmd/dev/attestgithub/attestgithub.go
```diff
@@ -0,0 +1,76 @@
...
+		&o.repository,
+		"repository",
+		"",
+		"path to GitHub repository, of form {owner}/{repo}",
+	)
+	cmd.MarkFlagRequired("repository") //nolint:errcheck
+
+	cmd.Flags().StringVar(
+		&o.pullRequestID,
+		"pull-request-ID",
```
wlynch, line 41: I think this should be pull request number? Pull requests also have ID's, but they're probably not what you want users interfacing with:


```sh
$ gh pr view --repo gittuf/gittuf 341 --json id,number
{
  "id": "PR_kwDOIVE5P85skkkv",
  "number": 341
}
```

internal/cmd/dev/attestgithub/attestgithub.go
```diff
@@ -0,0 +1,76 @@
...
+		"",
+		"signing key to use for signing attestation",
+	)
+	cmd.MarkFlagRequired("signing-key") //nolint:errcheck
+
+	cmd.Flags().StringVar(
+		&o.repository,
+		"repository",
+		"",
+		"path to GitHub repository, of form {owner}/{repo}",
```
wlynch, line 35: Make this more obvious this is the base repository.

internal/github/github.go
adityasaky: I was trying to avoid the dependency if we were only hitting a couple endpoints but given that this is almost certainly less brittle / better tested and isn't a completely random set of developers, I'll make the switch!

internal/cmd/dev/attestgithub/attestgithub.go
adityasaky: Yes, will fix.

internal/github/github.go
wlynch: go-github is intentionally pretty dependency light. https://github.com/google/go-github/blob/master/go.mod

I wouldn't worry about it too much.

internal/attestations/attestations.go
adityasaky: I've updated to store as follows.

a) It's still in refs/gittuf/attestations
b) There's still a github-pull-request dir in that tree
c) The path to the attestation is <base_branch>/<merge_sha>

If I check out that ref, I see `github-pull-requests/refs/heads/main/b69a55e72103a1560b5dbe2e4c961b488290a4ac` for example. The subject of that attestation is also now using the `gitCommit` digest instead of the `gitTree` digest.

adityasaky:
`GITTUF_DEV=1 gittuf dev attest-github --signing-key ../keys/alice --repository gittuf/gittuf --commit b69a55e72103a1560b5dbe2e4c961b488290a4ac --base-branch main`

`GITTUF_DEV=1 gittuf dev attest-github --signing-key ../keys/alice --repository gittuf/gittuf --pull-request-number 353`

These are equivalent, btw. --commit and --pull-request-number are mutually exclusive. If using --commit, it MUST be with --base-branch to ensure the right PR is picked and recorded. I think the former trivially allows us to use this in a github action on push to the main branch for the commit sha that github passes in.

internal/repository/attestations.go
```diff
@@ -195,3 +199,107 @@ func (r *Repository) RemoveReferenceAuthorization(ctx context.Context, signer ss
 	slog.Debug("Committing attestations...")
 	return allAttestations.Commit(r.r, commitMessage, signCommit)
 }
+
+// AddGitHubPullRequestAttestationForCommit identifies the pull request for a
+// specified commit ID and triggers AddGitHubPullRequestAttestationForNumber for
+// that pull request. Currently, the authentication token for the GitHub API is
+// read from the GITHUB_TOKEN environment variable.
+func (r *Repository) AddGitHubPullRequestAttestationForCommit(ctx context.Context, signer sslibdsse.SignerVerifier, owner, repository, commitID, baseBranch string, signCommit bool) error {
```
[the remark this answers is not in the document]
adityasaky: We control the identity / key used to sign the attestation. The commit adding the attestation to refs/gittuf/attestation would also be signed using gitsign.

internal/attestations/attestations.go
wlynch: > With PR commit, do you mean basically store this keyed by the PR merge / squash commit?

I was thinking the PR head commit so that PR attestations any time, similar to status checks. Using merge commits are tricky because they'll change constantly.

If the short term goal is to store attestations about PRs at merge time, then could this be stored in the existing ref attestation type with a different predicate?

Maybe we need 2 predicates? One to store the PR data on the PR commit and one to link the merge to the PR?

adityasaky: I don't know that we need different predicate types as the contents and really the intent to is basically the same. I think I can tweak this so it can be configured to be recorded on merge as well as on push in a pull request.

Here's what I suggest. If `merged_at` is not null, meaning the PR is merged, we store against the `merge_sha`. Otherwise, we store against the head of the feature branch. That should open this up to being recorded on push to `main` as well as on pull_request.

adityasaky: I've updated this now. If the PR is merged, the attestation is stored against the merge_sha and the base branch. If the PR is not yet merged, it's stored against the HEAD of the feature branch. In both cases, I also started recording the username + immutable ID, we need this for feature branches to know where they came from.

internal/repository/attestations.go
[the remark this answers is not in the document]
neilnaveen: But how do we trust the authority who controls the key?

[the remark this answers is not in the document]
adityasaky: We control the key because we're setting up the automation that adds the PR attestation.

wlynch (approved):

adityasaky:
cleaning up commits

adityasaky:
Thanks @wlynch!

PR #342, opened 2024-04-15
Fixes / debug statements in gittuf dev authorize

adityasaky:
Various things I'm running into as I'm kicking the tyres of #336.

### DIFF
--- a/internal/repository/attestations.go
+++ b/internal/repository/attestations.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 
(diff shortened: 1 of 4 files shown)


internal/repository/attestations.go
```diff
@@ -31,17 +32,17 @@ func (r *Repository) AddReferenceAuthorization(ctx context.Context, signer sslib
 
 	var err error
 
-	targetRef, err = gitinterface.AbsoluteReference(r.r, targetRef)
+	keyID, err := signer.KeyID()
```
wlynch, line 35: move this down closer to where it's used.

wlynch (approved):

PR #344, opened 2024-04-15
Changed LoadState to not assume that the first entry is for the policy ref

neilnaveen:
- Created a helper function that is almost the same as GetFirstEntry, but with an expected targetRef

### DIFF
--- a/internal/policy/policy_test.go
+++ b/internal/policy/policy_test.go
@@ -68,19 +68,196 @@ func TestInitializeNamespace(t *testing.T) {
 }
 
 func TestLoadState(t *testing.T) {
-	repo, state := createTestRepository(t, createTestStateWithOnlyRoot)
+	t.Run("loading while verifying multiple states", func(t *testing.T) {
+		repo, state := createTestRepository(t, createTestStateWithPolicy)
 
-	entry, err := rsl.GetLatestEntry(repo)
-	if err != nil {
-		t.Fatal(err)
-	}
+		entry, err := rsl.GetLatestEntry(repo)
+		if err != nil {
+			t.Fatal(err)
+		}
 
-	loadedState, err := LoadState(context.Background(), repo, entry.(*rsl.ReferenceEntry))
-	if err != nil {
-		t.Error(err)
-	}
+		loadedState, err := LoadState(context.Background(), repo, entry.(*rsl.ReferenceEntry))
+		if err != nil {
+			t.Error(err)
+		}
 
-	assert.Equal(t, state, loadedState)
+		assert.Equal(t, state, loadedState)
+
+		targetsMetadata, err := state.GetTargetsMetadata(TargetsRoleName)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		targetsMetadata, err = AddDelegation(targetsMetadata, "test-rule-1", []*tuf.Key{}, []string{""}, 1)
+		if err != nil {
+			t.Fatal(err)
+		}
+		state.ruleNames.Add("test-rule-1")
+
+		env, err := dsse.CreateEnvelope(targetsMetadata)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		signer, err := signerverifier.NewSignerVerifierFromSecureSystemsLibFormat(rootKeyBytes) //nolint:staticcheck
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		env, err = dsse.SignEnvelope(context.Background(), env, signer)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		state.TargetsEnvelope = env
+
+		if err := state.Commit(repo, "", false); err != nil {
+			t.Fatal(err)
+		}
+
+		if err := Apply(context.Background(), repo, false); err != nil {
+			t.Fatal(err)
+		}
+
+		targetsMetadata, err = AddDelegation(targetsMetadata, "test-rule-2", []*tuf.Key{}, []string{""}, 1)
+		if err != nil {
+			t.Fatal(err)
+		}
+		state.ruleNames.Add("test-rule-2")
+
+		env, err = dsse.CreateEnvelope(targetsMetadata)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		env, err = dsse.SignEnvelope(context.Background(), env, signer)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		state.TargetsEnvelope = env
+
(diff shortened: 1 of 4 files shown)


internal/rsl/rsl.go
```diff
@@ -506,6 +506,51 @@ func GetFirstEntry(repo *git.Repository) (*ReferenceEntry, []*AnnotationEntry, e
 	return firstEntry, annotations, nil
 }
 
+// GetFirstEntryForRef returns the very first entry in the RSL for a ref. It is expected to be
+// a reference entry as the first entry in the RSL cannot be an annotation.
+func GetFirstEntryForRef(repo *git.Repository, targetRef string) (*ReferenceEntry, []*AnnotationEntry, error) {
```
adityasaky, line 511: I think there's a lot of repetition with `GetFirstEntry` except for the single targetRef check. Can you update this so `GetFirstEntry` calls this one with `targetRef=""`, and only do the `targetRef == entry.RefName` check `if targetRef != ""`?

adityasaky (changes requested):
Thanks! I think this is good to go once we have this wired up to `GetFirstEntry`.

internal/rsl/rsl.go
neilnaveen: Fixed

internal/rsl/rsl.go
```diff
@@ -487,8 +493,10 @@ func GetFirstEntry(repo *git.Repository) (*ReferenceEntry, []*AnnotationEntry, e
 					// The first entry cannot be an annotation
 					return nil, nil, ErrInvalidRSLEntry
 				}
-				firstEntry = entry
-				break
+				if entry.RefName == targetRef || targetRef == "" {
```
adityasaky, line 496: I might have got it wrong last night. Does this still work when the first entry for a ref is actually the 17th entry overall? We hit this only when there aren't any parents, right? Do we want to continually track firstEntry outside this err block?

neilnaveen: I have fixed this, and changed the logic, the last check at the end is just if we never encounter a rsl entry

internal/policy/policy_test.go
```diff
@@ -68,93 +68,195 @@ func TestInitializeNamespace(t *testing.T) {
...
-	}
+		loadedState, err = LoadState(context.Background(), repo, entry.(*rsl.ReferenceEntry))
+		if err != nil {
+			t.Error(err)
+		}
 
-	assert.Equal(t, state, loadedState)
+		assert.Equal(t, state, loadedState)
+	})
+	t.Run("fail loading while verifying multiple states, bad sig", func(t *testing.T) {
```
adityasaky, line 160:
```suggestion

	t.Run("fail loading while verifying multiple states, bad sig", func(t *testing.T) {
```

internal/policy/policy_test.go
```diff
@@ -68,93 +68,195 @@ func TestInitializeNamespace(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		signer, err := signerverifier.NewSignerVerifierFromSecureSystemsLibFormat(rootKeyBytes) //nolint:staticcheck
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		badSigner, err := signerverifier.NewSignerVerifierFromSecureSystemsLibFormat(targets1KeyBytes) //nolint:staticcheck
```
adityasaky, line 196: move this down to when it's first used, please!

internal/policy/policy_test.go
```diff
@@ -68,93 +68,195 @@ func TestInitializeNamespace(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		state.TargetsEnvelope = env
+
+		if err := state.Commit(repo, "", false); err != nil {
+			t.Fatal(err)
+		}
+		policyStagingRef, err := repo.Reference(plumbing.ReferenceName(PolicyStagingRef), true)
```
adityasaky, line 237:
```suggestion

		policyStagingRef, err := repo.Reference(plumbing.ReferenceName(PolicyStagingRef), true)
```

internal/policy/policy_test.go
```diff
@@ -68,93 +68,195 @@ func TestInitializeNamespace(t *testing.T) {
...
+		if err := rsl.NewReferenceEntry(PolicyRef, policyStagingRef.Hash()).Commit(repo, false); err != nil {
+			t.Fatal(err)
+		}
+
+		entry, err = rsl.GetLatestEntry(repo)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		loadedState, err = LoadState(context.Background(), repo, entry.(*rsl.ReferenceEntry))
```
adityasaky: _, err = ...

internal/policy/policy_test.go
```diff
@@ -68,93 +68,195 @@ func TestInitializeNamespace(t *testing.T) {
...
+
```
adityasaky, line 257:
```suggestion
```

internal/policy/policy_test.go
```diff
@@ -68,93 +68,195 @@ func TestInitializeNamespace(t *testing.T) {
...
+		assert.Equal(t, "requested state has invalidly signed metadata: verifier's key and threshold constraints not met", err.Error())
```
adityasaky, line 258: I know the text is from the `fmt.Errorf` but I think the embedded error is named. Can we just do assert.ErrorIs on that one? Probably some thing about verifier constraints not being met or so

adityasaky (changes requested):

internal/policy/policy_test.go
neilnaveen: how would I be able to compare that single part of the error?

internal/policy/policy_test.go
neilnaveen: fixed

internal/policy/policy_test.go
neilnaveen: fixed

internal/policy/policy_test.go
neilnaveen: fixed

internal/policy/policy_test.go
neilnaveen: fixed

internal/policy/policy_test.go
neilnaveen: fixed

internal/policy/policy_test.go
adityasaky: Does assert.ErrorIs not match nested errors?

adityasaky (approved):

Issue #237, opened 2023-12-19
gittuf <-> gitsign attestation compatibility


wlynch:
### Add a description

Context: https://github.com/gittuf/gittuf/pull/170

We should look into gittuf + gitsign attestation compatibility.

.   | gittuf | gitsign
--|--|--
ref | refs/gittuf/attestations | refs/attestations/commits<br>refs/attestations/trees<br>...
attestation path | refs/heads/main/a-b | a/sbom.spdx.sig

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adityasaky:
cc @patzielinski

adityasaky:
Attestation compatibility would be great! Here are some questions I have towards getting there.

**1. Ref Path**

You raise a good point, `refs/attestations` is nice and generic, suitable for non gittuf tools to store their attestations as well. I think it makes sense to support this.

**2. Producer <-> Consumer**

How do we establish the assumptions around attestations? With gittuf, we'll define some specific predicate types that a gittuf user would use (the reference authorization for example) and other gittuf users would consume. This is easier to handle in terms of the types of policies gittuf allows too. What do we do for more generic attestation types wrt policies?

PR #347, opened 2024-04-17
repository: Clone attestations

adityasaky:
gittuf clone does not currently clone the attestations ref. This PR fixes that.


adityasaky:
Should this just be `refs/gittuf/*`? This fails, for example, if a repository has never had attestations set.

neilnaveen:
I would think that `refs/gittuf/*` would be better, since if more namespaces are added, this will have to constantly be updated, but if we are going to use a custom ref outside of the `refs/gittuf`, like for attestations, as mentioned in https://github.com/gittuf/gittuf/issues/237

wlynch (approved):

PR #356, opened 2024-04-24
Update docs

adityasaky:
This PR updates the roadmap and also adds a draft of a dogfood plan for gittuf.

### DIFF
--- a/docs/dogfood.md
+++ b/docs/dogfood.md
@@ -1,0 +1,37 @@
+# Dogfooding gittuf
+
+Last Modified: April 24, 2024
+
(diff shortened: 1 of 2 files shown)


JustinCappos (approved):
LGTM

adityasaky:
At the minimum, this has to land after #341. I'm fine merging this building gittuf off main, we can update the workflow after we cut v0.4.0 (which is very soon).

neilnaveen:
Why are we only recording attestations for the main branch, and not other branches too?

wlynch (approved):

adityasaky:
> Why are we only recording attestations for the main branch, and not other branches too?

I'm going to update the workflow on `pull_request` later this week.

PR #361, opened 2024-04-27
Refactoring and testing policy/verify.go

neilnaveen:
- VerifyRelativeForRef
- [x] Refactor into helper functions
- [x] Test with attestations 
- [x] Test when failing `currentPolicy.VerifyNewState`
- [x] Test with key revocation
- [x] Move VerifyTag's functionalty into VerifyRelativeForRef
- verifyEntry
- [x] Refactor into helper functions
- [x] Test with attestations 
- [x] Test when failing both types of ref rules 
- [x] Test with key revocation

In this PR, I am trying to clean up the core functionality of policy/verify.go which primarily runs through VerifyRelativeForRef and verifyEntry. I am also adding test to complete code coverage of these two functions.

In addition I am trying to figure out a clean way to merge the functionality of VerifyTag with VerifyRelativeForRef. This should be possible since at the end tags are refs themselves, just with different ways to verify them.

### DIFF
--- a/internal/policy/helpers_test.go
+++ b/internal/policy/helpers_test.go
@@ -92,6 +92,36 @@ func createTestStateWithOnlyRoot(t *testing.T) *State {
 	}
 }
 
+func createTestStateWithOnlyBadRoot(t *testing.T) *State {
+	t.Helper()
+
+	signer, err := signerverifier.NewSignerVerifierFromSecureSystemsLibFormat(targets1KeyBytes) //nolint:staticcheck
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	key, err := tuf.LoadKeyFromBytes(targets1PubKeyBytes)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rootMetadata := InitializeRootMetadata(key)
+
+	rootEnv, err := dsse.CreateEnvelope(rootMetadata)
+	if err != nil {
+		t.Fatal(err)
+	}
+	rootEnv, err = dsse.SignEnvelope(context.Background(), rootEnv, signer)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return &State{
+		RootPublicKeys: []*tuf.Key{key},
+		RootEnvelope:   rootEnv,
+	}
+}
+
 func createTestStateWithPolicy(t *testing.T) *State {
 	t.Helper()
 
--- a/internal/policy/helpers_test.go
+++ b/internal/policy/helpers_test.go
@@ -92,36 +92,6 @@ func createTestStateWithOnlyRoot(t *testing.T) *State {
 	}
 }
 
-func createTestStateWithOnlyBadRoot(t *testing.T) *State {
-	t.Helper()
-
-	signer, err := signerverifier.NewSignerVerifierFromSecureSystemsLibFormat(targets1KeyBytes) //nolint:staticcheck
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	key, err := tuf.LoadKeyFromBytes(targets1PubKeyBytes)
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	rootMetadata := InitializeRootMetadata(key)
-
-	rootEnv, err := dsse.CreateEnvelope(rootMetadata)
-	if err != nil {
-		t.Fatal(err)
-	}
-	rootEnv, err = dsse.SignEnvelope(context.Background(), rootEnv, signer)
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	return &State{
-		RootPublicKeys: []*tuf.Key{key},
-		RootEnvelope:   rootEnv,
-	}
-}
-
 func createTestStateWithPolicy(t *testing.T) *State {
 	t.Helper()
 
--- a/internal/policy/verify.go
+++ b/internal/policy/verify.go
@@ -196,7 +196,6 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
 			switch entry.RefName {
 			case PolicyStagingRef:
 				slog.Debug("Entry is for the policy staging reference...")
-				continue
 			case PolicyRef:
 				slog.Debug("Entry is for the policy reference...")
 				// TODO: this is repetition if the firstEntry is for policy
--- a/internal/policy/verify.go
+++ b/internal/policy/verify.go
@@ -191,13 +191,14 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
 			entries = entries[1:]
 
 			slog.Debug(fmt.Sprintf("Verifying entry '%s'...", entry.ID.String()))
+			slog.Debug("Checking reference of entry...")
 
-			slog.Debug("Checking if entry is for policy staging reference...")
-			if entry.RefName == PolicyStagingRef {
+			switch entry.RefName {
+			case PolicyStagingRef:
+				slog.Debug("Entry is for the policy staging reference...")
 				continue
-			}
-			slog.Debug("Checking if entry is for policy reference...")
-			if entry.RefName == PolicyRef {
+			case PolicyRef:
+				slog.Debug("Entry is for the policy reference...")
 				// TODO: this is repetition if the firstEntry is for policy
 				newPolicy, err := loadStateForEntry(repo, entry)
 				if err != nil {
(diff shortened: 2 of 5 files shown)


lukpueh (commented):
The refactor looks fine to me. Though, I can't tell if the added tests are correct, just that they pass (independently of the refactor).

I suggest someone who is more familiar with gittuf and the code takes a look at those. Or I take another look once I am that someone. :)

Btw. a tiny bit more info about the tests would be extremely helpful. e.g. in the PR description, in a commit message, or in a code comment. Cheers!

internal/policy/verify.go
```diff
@@ -191,13 +191,14 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
 
 			slog.Debug(fmt.Sprintf("Verifying entry '%s'...", entry.ID.String()))
+			slog.Debug("Checking reference of entry...")
 
-			slog.Debug("Checking if entry is for policy staging reference...")
-			if entry.RefName == PolicyStagingRef {
+			switch entry.RefName {
+			case PolicyStagingRef:
+				slog.Debug("Entry is for the policy staging reference...")
 				continue
```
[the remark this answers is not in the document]
neilnaveen: I removed the continue

internal/policy/helpers_test.go
```diff
@@ -92,6 +92,36 @@ func createTestStateWithOnlyRoot(t *testing.T) *State {
 	}
 }
 
+func createTestStateWithOnlyBadRoot(t *testing.T) *State {
```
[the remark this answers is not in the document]
neilnaveen: Looking back, this does not make much sense since, technically, it is not a bad state; it is just "bad" in context with the test that is using it. I have removed this function and moved it into the test that uses it, with some comments to better explain why it is a bad state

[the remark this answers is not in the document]
lukpueh: Thanks!

lukpueh:
Thanks for addressing my comments, @neilnaveen! I saw you also added another refactoring commit yesterday.  Is the PR ready for another round of comments or would you like to make more changes? I'm fine either way.

neilnaveen:
@lukpueh, I've made progress on the problem. I believe I've found a solution that should work, as it passes the test and seems sound in theory. I've made changes to remove VerifyTag and have all tag verification running through VerifyRelativeForRef. Currently, the actual verification of each tag entry is still being handled by verifyTagEntry, but in the next commits, I'll attempt to transfer those functionalities to verifyEntry. I think this is ready for another round of comments. This part is completed, but more will be added later.

lukpueh:
Thanks for the quick reply, @neilnaveen. I'll give it another read then. And maybe you can push future changes to a new branch and submit a separate PR. Does that work for you?

neilnaveen:
Thanks for all the feedback @lukpueh. @adityasaky @wlynch @patzielinski  be able to review this PR, since while I think it works, this is a big change in verificaiton. 

On the note of addressing the merger of verifyEntry and verifyTagEntry, I think all of that should go into this PR, since this isn't actualy changing a lot of code, just moving a few small specific checks that happen for tag verification into verifyEntry and VerifyRelativeForRef.

internal/repository/verify.go
```diff
@@ -88,9 +88,14 @@ func (r *Repository) VerifyCommit(ctx context.Context, ids ...string) map[string
 	return policy.VerifyCommit(ctx, r.r, ids...)
 }
 
-func (r *Repository) VerifyTag(ctx context.Context, ids []string) map[string]string {
+func (r *Repository) VerifyTag(ctx context.Context, tags []string) map[string]string {
```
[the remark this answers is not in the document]
neilnaveen: I also thought about this, and the same with VerifyCommit, but since VerifyTag already had this functionality, I just left it as it was

internal/policy/verify.go
```diff
@@ -211,146 +224,149 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
-				continue
-			}
+	return nil
+}
 
-			newEntryCommit, err := gitinterface.GetCommit(repo, newEntry.TargetID)
-			if err != nil {
-				return err
-			}
+// findFixEntry is only reached when we have an invalid state.
```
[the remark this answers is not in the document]
neilnaveen: fixed

internal/policy/verify.go
```diff
@@ -184,20 +184,33 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
 
-			slog.Debug("Checking if entry is for policy staging reference...")
-			if entry.RefName == PolicyStagingRef {
-				continue
-			}
-			slog.Debug("Checking if entry is for policy reference...")
-			if entry.RefName == PolicyRef {
+			switch entry.RefName {
+			case PolicyStagingRef:
+				slog.Debug("Entry is for the policy staging reference...")
```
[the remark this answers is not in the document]
neilnaveen: fixed

internal/policy/verify.go
```diff
@@ -184,20 +184,33 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
+	targetRefEntiresCount := 0
+
+	for _, entry := range entries {
+		if entry.RefName == target {
+			targetRefEntiresCount++
+		}
+	}
+
+	if strings.HasPrefix(target, gitinterface.TagRefPrefix) && targetRefEntiresCount > 1 {
+		return ErrMultipleTagRSLEntriesFoundMessage
```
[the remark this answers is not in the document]

internal/policy/verify.go
```diff
@@ -36,16 +36,16 @@ const (
...
-	ErrUnknownObjectType       = errors.New("unknown object type passed to verify signature")
-	ErrInvalidVerifier         = errors.New("verifier has invalid parameters (is threshold 0?)")
-	ErrVerifierConditionsUnmet = errors.New("verifier's key and threshold constraints not met")
+	ErrUnauthorizedSignature             = errors.New("unauthorized signature")
+	ErrInvalidEntryNotSkipped            = errors.New("invalid entry found not marked as skipped")
+	ErrLastGoodEntryIsSkipped            = errors.New("entry expected to be unskipped is marked as skipped")
+	ErrUnknownObjectType                 = errors.New("unknown object type passed to verify signature")
+	ErrInvalidVerifier                   = errors.New("verifier has invalid parameters (is threshold 0?)")
+	ErrVerifierConditionsUnmet           = errors.New("verifier's key and threshold constraints not met")
+	ErrMultipleTagRSLEntriesFoundMessage = errors.New("multiple RSL entries found for tag")
```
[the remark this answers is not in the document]

internal/policy/verify.go
```diff
@@ -184,20 +184,34 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
 	// Verify each entry, looking for a fix when an invalid entry is encountered
 	var invalidEntry *rsl.ReferenceEntry
 	var verificationErr error
+
+	targetRefEntiresCount := 0
+
+	for _, entry := range entries {
```
adityasaky, line 190: This check doesn't belong here, IMO. This would only check in the specified range, so there could well be additional entries for the tag not in the specified range. The duplication check for tags needs to happen separately.

internal/policy/verify.go
```diff
@@ -184,20 +184,34 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
 			slog.Debug(fmt.Sprintf("Verifying entry '%s'...", entry.ID.String()))
-
-			slog.Debug("Checking if entry is for policy staging reference...")
-			if entry.RefName == PolicyStagingRef {
-				continue
-			}
-			slog.Debug("Checking if entry is for policy reference...")
-			if entry.RefName == PolicyRef {
+			slog.Debug("Checking reference of entry...")
+
```
adityasaky, line 208:
```suggestion
```

internal/policy/verify.go
```diff
@@ -184,20 +184,34 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
 
 			slog.Debug(fmt.Sprintf("Verifying entry '%s'...", entry.ID.String()))
-
-			slog.Debug("Checking if entry is for policy staging reference...")
-			if entry.RefName == PolicyStagingRef {
-				continue
-			}
-			slog.Debug("Checking if entry is for policy reference...")
-			if entry.RefName == PolicyRef {
+			slog.Debug("Checking reference of entry...")
```
adityasaky, line 207:
```suggestion

			slog.Debug("Checking reference of entry...")
```

internal/policy/verify.go
```diff
@@ -184,20 +184,34 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
+
+			switch entry.RefName {
+			case PolicyStagingRef:
+				// If we encounter a policy-staging entry, we alert the user, but ignore it for gittuf verification
```
adityasaky, line 211: nit: wrap cols

internal/policy/verify.go
```diff
@@ -184,20 +184,34 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
+				slog.Debug("Entry is for the policy staging reference...")
```
adityasaky, line 212: nit: we use the `...` trailer when something is about to happen, such as "Checking X...", but to indicate a result, we don't have punctuation at the end (unless it's something like "Verification successful!").

internal/policy/verify.go
```diff
@@ -683,105 +608,59 @@ func verifyEntry(ctx context.Context, repo *git.Repository, policy *State, attes
 		return fmt.Errorf("verifying file namespace policies failed, %w", ErrUnauthorizedSignature)
 	}
 
-	return nil
-}
-
-func verifyTagEntry(ctx context.Context, repo *git.Repository, policy *State, entry *rsl.ReferenceEntry) error {
```
adityasaky, line 689: I think this is going to overload the verification of a branch entry and a tag entry. I suggest rolling this change back. IMO, verifyEntry and verifyTagEntry ought to be separate.

We need to have a different conversation about the `verify-tag` (and `verify-commit`) user facing endpoints, but here, let's keep them separate so we can better reason about it.

adityasaky: Also, doesn't this break with `getCommits` for a tag? When we load the commits modified in the case of this entry to find modified files?

internal/policy/verify_test.go
```diff
@@ -728,65 +728,220 @@ func TestVerifyRelativeForRef(t *testing.T) {
...
+		}
+
+		if err := repo.Storer.SetReference(plumbing.NewHashReference(plumbing.ReferenceName(refName), plumbing.ZeroHash)); err != nil {
+			t.Fatal(err)
+		}
+
+		initialPolicyEntry, _, err := rsl.GetLatestReferenceEntryForRef(repo, PolicyRef)
+
+		repo, _ = createTestRepository(t, createTestStateWithPolicy)
+
```
adityasaky, line 884: two patterns:

1. `if err := ...; err !=nil` then there may or may not be a newline above, depending on what it's coming after.
2. OTOH, `..., err := ...\n if err != nil`, where the err check is purely an err check, no extra newline above.

adityasaky:
```suggestion
```

adityasaky (changes requested):
I think the improvements to verifyEntry are great! IMO, verifyTagEntry needs to be separate though.

adityasaky:
Also, could you rebase over main?

internal/policy/verify.go
neilnaveen: I have moved it into verifyTagEntry

internal/policy/verify.go
neilnaveen: Fixed

internal/policy/verify.go

internal/policy/verify.go

internal/policy/verify.go
neilnaveen: I have rolled back the last commit

internal/policy/verify_test.go
neilnaveen: Fixed

neilnaveen:
@adityasaky Rebased over main

internal/policy/verify.go

internal/policy/verify.go
```diff
@@ -192,12 +193,14 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
-			}
-			slog.Debug("Checking if entry is for policy reference...")
-			if entry.RefName == PolicyRef {
+			slog.Debug("Checking reference of entry...")
+			switch entry.RefName {
+			case PolicyStagingRef:
+				// If we encounter a policy-staging entry, we alert the user, but ignore it
+				// for gittuf verification
+				slog.Debug("Entry is for the policy staging reference")
+			case PolicyRef:
```
adityasaky, line 202:
```suggestion

			case PolicyRef:
```

internal/policy/verify.go
```diff
@@ -192,12 +193,14 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
+				slog.Debug("Entry is for the policy reference...")
```
adityasaky, line 203: trailing ...

internal/policy/verify.go
```diff
@@ -192,12 +193,14 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
-			slog.Debug("Checking if entry is for policy staging reference...")
-			if entry.RefName == PolicyStagingRef {
-				continue
-			}
-			slog.Debug("Checking if entry is for policy reference...")
-			if entry.RefName == PolicyRef {
+			slog.Debug("Checking reference of entry...")
+			switch entry.RefName {
+			case PolicyStagingRef:
+				// If we encounter a policy-staging entry, we alert the user, but ignore it
```
adityasaky, line 199: line wrap

internal/policy/verify.go
```diff
@@ -211,146 +214,149 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
 
 				slog.Debug("Updating current policy...")
 				currentPolicy = newPolicy
-				continue
-			}
-
-			slog.Debug("Checking if entry is for attestations reference...")
-			if entry.RefName == attestations.Ref {
+			case attestations.Ref:
```
adityasaky, line 217:
```suggestion

			case attestations.Ref:
```

internal/policy/verify.go
```diff
@@ -211,146 +214,149 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
+				slog.Debug("Entry is for the attestations reference...")
```
adityasaky, line 218: trailing ...

adityasaky: Also, the log message for this block should match the policy block above for consistency; my mistake originally. Could you add some similar statements?

internal/policy/verify.go
```diff
@@ -211,146 +214,149 @@ func VerifyRelativeForRef(ctx context.Context, repo *git.Repository, initialPoli
...
+// queue. Entries that are for the same ref but not the fix are
+// considered invalid. The workflow enters a valid state again when a)
+// the fix entry (which hasn't also been revoked) is found, and b) all
+// entries for the ref in the invalid range are marked as skipped by an
+// annotation. If these conditions don't both hold, the workflow returns
+// an error. After the fix is found, all remaining entries in the
+// original queue are also added to the new queue. The new queue then
+// takes the place of the original queue. This ensures that all entries
+// are processed even when an invalid state is reached.
+func findFixEntry(repo *git.Repository, invalidEntry *rsl.ReferenceEntry, entries []*rsl.ReferenceEntry, annotations map[plumbing.Hash][]*rsl.AnnotationEntry, verificationErr error) ([]*rsl.ReferenceEntry, error) {
```
adityasaky, line 280: Could you move this to be after `verifyEntry`?

internal/policy/verify.go
```diff
@@ -445,70 +451,6 @@ func VerifyCommit(ctx context.Context, repo *git.Repository, ids ...string) map[
 	return status
 }
 
-// VerifyTag verifies the signature on the RSL entries for the specified tags.
```
adityasaky, line 448: Let's scope PRs. Can we address VerifyCommit/VerifyTag in separate PRs? This file is possibly the single most important in gittuf, so let's have small, easily reviewable changes. :smile:

internal/policy/verify.go
```diff
@@ -526,31 +468,23 @@ func (s *State) VerifyNewState(ctx context.Context, newPolicy *State) error {
...
 	if strings.HasPrefix(entry.RefName, gitinterface.TagRefPrefix) {
-		return verifyTagEntry(ctx, repo, policy, entry)
+		return verifyTagEntry(ctx, repo, policy, entry, target)
 	}
 
-	var (
-		gitNamespaceVerified  = false
-		pathNamespaceVerified = true // Assume paths are verified until we find out otherwise
-	)
+	pathNamespaceVerified := true // Assume paths are verified until we find out otherwise
```
adityasaky, line 480: I personally find it more readable to have both bools defined here. The helper below should return a bool that we use temporarily, as `if verified, _, err := <>; verified { gitNamespaceVerified = true }`. Having the declarations in one place make it easier to track for me at least.

internal/policy/verify.go
```diff
@@ -686,7 +604,30 @@ func verifyEntry(ctx context.Context, repo *git.Repository, policy *State, attes
 	return nil
 }
 
-func verifyTagEntry(ctx context.Context, repo *git.Repository, policy *State, entry *rsl.ReferenceEntry) error {
+func verifyAttestation(ctx context.Context, verifiers []*Verifier, commitObj *object.Commit, authorizationAttestation *sslibdsse.Envelope) (bool, string, error) {
+	// No verifiers => no restrictions for the git namespace
+	if len(verifiers) == 0 {
```
adityasaky, line 609: Let's handle this separately please, we want fewer "valid" return states. Currently, both `true, ""` and `true, verifierName` are valid. Let's make it so we only have the latter. This helper shouldn't be invoked if len(verifiers) == 0. This could also mean that we don't need a boolean return at all; we have an error returned if verification fails, or error==nil and the name of the verifier if successful.

internal/policy/verify.go
```diff
@@ -686,7 +604,30 @@ func verifyEntry(ctx context.Context, repo *git.Repository, policy *State, attes
...
+			// Unexpected error
+			return false, "", err
+		}
+		// Haven't found a valid verifier, continue with next
+	}
+	return false, "", nil
+}
+
+func verifyTagEntry(ctx context.Context, repo *git.Repository, policy *State, entry *rsl.ReferenceEntry, target string) error {
+	if _, _, err := rsl.GetLatestReferenceEntryForRefBefore(repo, target, entry.GetID()); err == nil {
```
adityasaky, line 627: Depending on when this is invoked, this could still let things fall through. What if we're verifying the first tag entry and the duplicate is after this one? Also, this change may belong in a separate PR to keep things scoped. For this PR, let's keep the focus on fixing up VerifyRelativeForRef and verifyEntry.

internal/policy/verify_test.go
```diff
@@ -728,65 +728,222 @@ func TestVerifyRelativeForRef(t *testing.T) {
...
-}
 
-func TestVerifyCommit(t *testing.T) {
-	repo, _ := createTestRepository(t, createTestStateWithPolicy)
-	refName := "refs/heads/main"
-	gpgKey, err := gpg.LoadGPGKeyFromBytes(gpgPubKeyBytes)
-	if err != nil {
-		t.Fatal(err)
-	}
+	t.Run("no recovery, with attestations", func(t *testing.T) {
```
adityasaky, line 732: I haven't looked through the tests yet, I think it can wait until you've had a chance to look through my comments above.

neilnaveen:
Closing this, since most of this PR was about refactoring out VerifyTag, which has already been done

PR #364, opened 2024-05-01
ci: Run gittuf workflows only on gittuf/gittuf

adityasaky:
This avoids failures of gittuf workflows for forks that may not have the gittuf refs.


adityasaky:
Possibly, we need to recognize too that the RSL entries synced here pertain to pushes to the main repo. So it's also possible someone wants a separate RSL for their mirror...

JustinCappos (approved):

PR #367, opened 2024-05-02
Add Git binary support to gitinterface

adityasaky:
This is an early draft to using the git binary for gitinterface operations in gittuf.


internal/gitinterface/hash.go
```diff
@@ -0,0 +1,37 @@
...
+package gitinterface
+
+import (
+	"encoding/hex"
+	"errors"
+)
+
+const (
+	zeroSHA1HashString   = "0000000000000000000000000000000000000000"
+	zeroSHA256HashString = "0000000000000000000000000000000000000000000000000000000000000000"
```
neilnaveen, line 12: Is this preperation for hash agility? If so, will this only be implmented with the git binary?

adityasaky: yep!

adityasaky:
Closing in favour of #431 as last in series.

Issue #368, opened 2024-05-02
Changing RSL record to work better with rebases/breaking ref changes


neilnaveen:
### What happened?

If we rebase, or change any ref using any comand, e.g `git reset`, the refs changes and some rsl entries might have previously pointed to these commits. If we run verify, these RSL entries will be invalid, since the commits they point to will not exist. 

To fix this, in RSL record, after recording we can iterate through each RSL entry in the repo, and if they point to a commit that does not exist, we can skip them, with an annotation. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

neilnaveen:
I will work on this

PR #369, opened 2024-05-02
Automaticaly Skip RSL Reference Entries That Point to Non-Existent Commits

neilnaveen:
- Fixes https://github.com/gittuf/gittuf/issues/368


internal/repository/rsl.go
```diff
@@ -23,7 +23,9 @@ var (
 )
 
 // RecordRSLEntryForReference is the interface for the user to add an RSL entry
-// for the specified Git reference.
+// for the specified Git reference. If rsl entries exist that point to commits
+// that do not exist, e.g, a rebase which removes some commits,
+// RecordRSLEntryForReference creates an annotation to skip all of these entries
```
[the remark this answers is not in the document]

internal/rsl/rsl.go
```diff
@@ -513,6 +514,46 @@ func GetFirstReferenceEntryForRef(repo *git.Repository, targetRef string) (*Refe
 	return firstEntry, annotations, nil
 }
 
+// SkipAllInvalidReferenceEntries skips all reference entries that could have previously been
+// normal reference entry's, but after a force change a ref that the rsl entries point to, some of the
+// entries may point to non-existent commits. This function adds an annotation that skips all entries
+// that are pointing to non-existent commits
```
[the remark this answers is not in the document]
neilnaveen: Fixed

internal/rsl/rsl.go
```diff
@@ -513,6 +514,46 @@ func GetFirstReferenceEntryForRef(repo *git.Repository, targetRef string) (*Refe
...
+	if err != nil {
+		return err
+	}
+
+	entriesToSkip := []plumbing.Hash{}
+
+	for {
+		if entry, ok := iterator.(*ReferenceEntry); ok {
+			if _, err := gitinterface.GetCommit(repo, entry.TargetID); err != nil {
+				if err.Error() == "object not found" {
```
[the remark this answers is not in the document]
neilnaveen: Fixed. I forgot about accessing public errors from go-git

internal/rsl/rsl.go
```diff
@@ -513,6 +514,46 @@ func GetFirstReferenceEntryForRef(repo *git.Repository, targetRef string) (*Refe
...
+					entriesToSkip = append(entriesToSkip, entry.ID)
```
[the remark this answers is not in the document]
neilnaveen: Fixed

neilnaveen:
@patzielinski 

I think that not skipping an entry pointing to a non-existent commit does not make sense since this RSL entry would automatically fail verification for somthing not in the repo. If we alert them to each entry removed with logging, the user can decide if some malicious action was performed. 

Still, since this happens on `rsl record`, everyone should record their changes, and rebases/breaking actions that create these situations. This means that we should never reach a scenario where somebody else's RSL entries are removed, only your own.

neilnaveen:
We also could have a -f flag which would skip everything automaticaly, or else it asks if the user wants to skip it.

neilnaveen:
Ok, that makes sense. On the note of logging at a higher level than Debug, this should be something we should look more into. For example `verify-ref` with `--verbose` throws a lot at the user, but we may want to see only when a fix entry is being found, when somthing fails verification, and at which RSL entry. It's not a Debug mode, but somthing like `git push`'s output. It is basic enough to understand, but if anything fails, it is easy to understand what and where somthing went wrong for simple problems.

adityasaky:
I view `gittuf rsl record` etc as lower level commands, possibly one that the user won't interact with. With that model, IMO we should keep such low level commands simple and to the point. Thoughts on making the recovery a separate command? `gittuf rsl recover` or so would handle this, following which a gittuf rsl record could be the "roll back" command.

This also makes it easier to perform recovery only when needed, ie when verify-ref complains. We don't want to introduce latency into the RSL entry creation command itself.

adityasaky:
>Should this be always automated without user interaction, or are there cases where someone may want to not skip these entries? (i.e. by way of a prompt like "Entry pointing to non-existent commit detected. Do you want to mark this entry to be skipped? [Y/n]" or something to that effect)

IMO, we build out the low level functionality for this now and decide the higher order workflows later, depending on what invokes this plumbing command.

internal/rsl/rsl.go
```diff
@@ -513,6 +514,47 @@ func GetFirstReferenceEntryForRef(repo *git.Repository, targetRef string) (*Refe
...
+	if err != nil {
+		return err
+	}
+
+	entriesToSkip := []plumbing.Hash{}
+
+	for {
+		if entry, ok := iterator.(*ReferenceEntry); ok {
+			if _, err := gitinterface.GetCommit(repo, entry.TargetID); err != nil {
+				if errors.Is(err, plumbing.ErrObjectNotFound) {
```
adityasaky, line 534: This isn't right when invoked on the same copy of the repository where the rebase is performed; a rebase doesn't mean the commit isn't in the object store anymore. The commit sticks around until the garbage collector cleans it up.

Instead, we want to use `KnowsCommit` from the current commit. By making this _per ref_, we walk back until we find a reference entry for the ref whose target is known by the current commit. Take the following example:

A <- B <- C <- D <- E | This is the original commit history for a branch, we have RSL entries at A, B, D, and E
A <- B <- F                 | This is the new commit history for the branch, C, D, and E are squashed to create F

 We walk back until we find the RSL entry for B which is _known_ by F. Clearly, the entry for D and E are invalid now. Thus, the function:
 a) creates an annotation entry skipping the RSL entries at D and E
b) creates a reference entry at F

Does this make sense? :smile:

adityasaky: cc @wlynch and @justincappos for if this workflow makes sense.

internal/rsl/rsl_test.go
```diff
@@ -709,6 +712,109 @@ func TestGetFirstReferenceEntryForRef(t *testing.T) {
...
+		repo, err := git.Init(memory.NewStorage(), memfs.New())
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if err := InitializeNamespace(repo); err != nil {
+			t.Fatal(err)
+		}
+
+		commitID, err := gitinterface.CommitUsingSpecificKey(repo, plumbing.ZeroHash, "refs/heads/main", "", gpgKeyBytes)
```
adityasaky, line 776: do we need a signature? The conditions we check at this package should not worry about the signature at all, no?

adityasaky (changes requested):
Thanks for working on this, this is a very important feature! I've left some comments below.

internal/repository/rsl.go
```diff
@@ -50,7 +52,11 @@ func (r *Repository) RecordRSLEntryForReference(refName string, signCommit bool)
 	// signCommit must be verified for the refName in the delegation tree.
 
 	slog.Debug("Creating RSL reference entry...")
-	return rsl.NewReferenceEntry(absRefName, ref.Hash()).Commit(r.r, signCommit)
+	if err := rsl.NewReferenceEntry(absRefName, ref.Hash()).Commit(r.r, signCommit); err != nil {
+		return err
+	}
+
+	return rsl.SkipAllInvalidReferenceEntries(r.r, signCommit)
```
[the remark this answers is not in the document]
adityasaky: but this also ties back to whether we want this to happen as part of a separate `gittuf rsl` command invoked explicitly by a higher level user facing workflow.

internal/repository/rsl.go
```diff
@@ -23,7 +23,9 @@ var (
 )
 
 // RecordRSLEntryForReference is the interface for the user to add an RSL entry
-// for the specified Git reference.
+// for the specified Git reference. If RSL entries exist that point to commits
+// that do not exist, e.g, a rebase which removes some commits,
+// RecordRSLEntryForReference creates an annotation to skip all of those entries
```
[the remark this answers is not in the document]
neilnaveen: I have removed all that text, so this is fixed

internal/rsl/rsl.go
```diff
@@ -513,6 +514,47 @@ func GetFirstReferenceEntryForRef(repo *git.Repository, targetRef string) (*Refe
 	return firstEntry, annotations, nil
 }
 
+// SkipAllInvalidReferenceEntries skips all reference entries that could have previously been
```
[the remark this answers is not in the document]
neilnaveen: Fixed

internal/repository/rsl.go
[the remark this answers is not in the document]
neilnaveen: Moved to `gittuf rsl skip-rewritten`

internal/rsl/rsl.go
```diff
@@ -513,6 +514,47 @@ func GetFirstReferenceEntryForRef(repo *git.Repository, targetRef string) (*Refe
...
+// normal reference entries, but after a force change a ref that the rsl entries point to, some of the
+// entries may point to non-existent commits. This function adds an annotation that skips all entries
+// that are pointing to non-existent commits
+func SkipAllInvalidReferenceEntries(repo *git.Repository, signCommit bool) error {
+	slog.Debug("Checking if RSL entries point to non-existent commits...")
```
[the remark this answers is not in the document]
neilnaveen: Fixed

internal/rsl/rsl.go
```diff
@@ -513,6 +514,47 @@ func GetFirstReferenceEntryForRef(repo *git.Repository, targetRef string) (*Refe
 	return firstEntry, annotations, nil
 }
 
+// SkipAllInvalidReferenceEntries skips all reference entries that could have previously been
+// normal reference entries, but after a force change a ref that the rsl entries point to, some of the
+// entries may point to non-existent commits. This function adds an annotation that skips all entries
+// that are pointing to non-existent commits
+func SkipAllInvalidReferenceEntries(repo *git.Repository, signCommit bool) error {
```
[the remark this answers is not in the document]
neilnaveen: Fixed

internal/rsl/rsl.go
neilnaveen: I have changed the logic to reflect your idea. Was this what you meant?

internal/rsl/rsl_test.go
neilnaveen: I have removed the commit using specific key. Thanks

internal/cmd/rsl/skiprewritten/skiprewritten.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	return repo.SkipAllInvalidReferenceEntriesForRef(args[0], true)
+}
+
+func New() *cobra.Command {
+	o := &options{}
+	cmd := &cobra.Command{
+		Use:               "skip-rewritten",
+		Short:             "creates an RSL annotation to skip all rsl entries that point to commits that do not exist in the specified ref",
```
adityasaky, line 28:
```suggestion
		Short:             "Creates an RSL annotation to skip RSL reference entries that point to commits that do not exist in the specified ref",
```

internal/rsl/rsl.go
```diff
@@ -513,6 +514,64 @@ func GetFirstReferenceEntryForRef(repo *git.Repository, targetRef string) (*Refe
...
+		return err
+	}
+
+	entriesToSkip := []plumbing.Hash{}
+
+	for {
+		if entry, ok := iterator.(*ReferenceEntry); ok {
+			latestEntryCommit, err := gitinterface.GetCommit(repo, latestEntry.TargetID)
+
+			var isAncestor bool
```
adityasaky, line 537: move this above the GetCommit statement: we don't want anything between assigning `err` and checking it.

internal/rsl/rsl.go
```diff
@@ -513,6 +514,64 @@ func GetFirstReferenceEntryForRef(repo *git.Repository, targetRef string) (*Refe
...
+
+			switch {
```
adityasaky, line 539: I see what you're doing, but I personally find this less readable than regular if else, even if there's a nested if in there. I think there's a more readable sequence to "okay we have an error, is it object not found?" or "we don't have an error, let's check whether there's ancestry" and so on.

adityasaky: if you want to keep the switch, even just reordering err == nil and errors.Is(err, obj not found) would go a long way as at least the err != nil cases would be grouped together (with the default).

internal/rsl/rsl.go
```diff
@@ -513,6 +514,64 @@ func GetFirstReferenceEntryForRef(repo *git.Repository, targetRef string) (*Refe
...
+					return err
+				}
+			default:
+				return err
+			}
+			if !isAncestor {
+				slog.Debug(fmt.Sprintf("Found RSL entry '%s' pointing to a commit, '%s', that does not exist in the target ref.", entry.ID, entry.TargetID))
+				entriesToSkip = append(entriesToSkip, entry.ID)
+			} else {
+				slog.Debug(fmt.Sprintf("Found RSL entry '%s' pointing to a commit, '%s', that does exist in the target ref. No more commits to skip.", entry.ID, entry.TargetID))
```
adityasaky, line 555: /s/that does exist/that exists

maybe you could also log the target ref?

internal/rsl/rsl_test.go
```diff
@@ -710,6 +710,110 @@ func TestGetFirstReferenceEntryForRef(t *testing.T) {
...
+
+		annotationEntry, ok := nextEntry.(*AnnotationEntry)
+
+		if !ok {
+			t.Fatal("invalid entry type")
+		}
+
+		assert.Equal(t, annotationEntry.RSLEntryIDs, []plumbing.Hash{entry.GetID()})
+	})
+	t.Run("everything is valid, nothing should change", func(t *testing.T) {
```
adityasaky, line 764:
```suggestion

	t.Run("everything is valid, nothing should change", func(t *testing.T) {
```

internal/rsl/rsl_test.go
```diff
@@ -710,6 +710,110 @@ func TestGetFirstReferenceEntryForRef(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		entry, err := GetLatestEntry(repo)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		referenceEntry, ok := entry.(*ReferenceEntry)
+
```
adityasaky, line 740:
```suggestion
```

internal/rsl/rsl_test.go
```diff
@@ -710,6 +710,110 @@ func TestGetFirstReferenceEntryForRef(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		nextEntry, err := GetLatestEntry(repo)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		annotationEntry, ok := nextEntry.(*AnnotationEntry)
+
```
adityasaky, line 757:
```suggestion
```

internal/rsl/rsl.go
neilnaveen: The linter hit me here, telling me to switch to a switch case. If you are ok with it I can add a linter ignore.

adityasaky: okay, let's reorder the cases then so err==nil is handled first? also `inAncestor` could be moved up.

internal/rsl/rsl_test.go
neilnaveen: Fixed

internal/rsl/rsl_test.go
neilnaveen: Fixed

internal/rsl/rsl.go
neilnaveen: I have added the target ref in front of both slog.Debug statements.

internal/rsl/rsl.go
neilnaveen: I have flipped the statements

internal/cmd/rsl/skiprewritten/skiprewritten.go
neilnaveen: Fixed

internal/rsl/rsl.go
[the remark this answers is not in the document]
adityasaky: I still see non-existent in the debug msg, could you update that?

internal/rsl/rsl.go
```diff
@@ -513,6 +514,64 @@ func GetFirstReferenceEntryForRef(repo *git.Repository, targetRef string) (*Refe
...
+				isAncestor = false
+			default:
+				return err
+			}
+
+			if !isAncestor {
+				slog.Debug(fmt.Sprintf("For target ref %s, found RSL entry '%s' pointing to a commit, '%s', that does not exist in the target ref.", targetRef, entry.ID, entry.TargetID))
+				entriesToSkip = append(entriesToSkip, entry.ID)
+			} else {
+				slog.Debug(fmt.Sprintf("For target ref %s, found RSL entry '%s' pointing to a commit, '%s', that does exist in the target ref. No more commits to skip.", targetRef, entry.ID, entry.TargetID))
```
adityasaky, line 555:
```suggestion
				slog.Debug(fmt.Sprintf("For target ref %s, found RSL entry '%s' pointing to a commit, '%s', that exists in the target ref. No more commits to skip.", targetRef, entry.ID, entry.TargetID))
```

adityasaky (approved):
could you address the two log message updates I've flagged?

internal/rsl/rsl.go
neilnaveen: Fixed

neilnaveen:
@adityasaky fixed the first `slog.Debug` in `SkipAllInvalidReferenceEntries` so as not to use the word invalid.

adityasaky:
Great! Could you clean up commits?

neilnaveen:
@adityasaky rebased

adityasaky (approved):
Thanks!

PR #370, opened 2024-05-03
Rework policy signing-key flag check

patzielinski:
Invoking gittuf's policy commands that require a signing key without the `-k` flag to specify said signing key does not raise an error.

Currently, this check is performed by setting an attribute on the flag via cobra. Unfortunately, it appears that disabling this requirement for subcommands that do not need a signing key (i.e. `list-rules`) _also_ disables the flag for subcommands that _do_ need it.

 This PR recreates the required flag check in the `CheckIfSigningViable` function.

A minor note: As this is done in PreRunE, the signing key flag check will run before other flag checks. Run `gittuf policy add-rule` without any flags before and after this PR change to see what I mean. Unsure if we care about this.

### DIFF
--- a/internal/cmd/common/common.go
+++ b/internal/cmd/common/common.go
@@ -84,6 +84,21 @@ func LoadSigner(keyBytes []byte) (sslibdsse.SignerVerifier, error) {
 	return signerverifier.NewSignerVerifierFromSecureSystemsLibFormat(keyBytes) //nolint:staticcheck
 }
 
+// CheckIfSigningViableWithFlag checks if a signing key was specified via the
+// "signing-key" flag, and then calls CheckIfSigningViable
+func CheckIfSigningViableWithFlag(cmd *cobra.Command, _ []string) error {
+	signingKeyFlag := cmd.Flags().Lookup("signing-key")
+
+	// Check if a signing key was specified via the "signing-key" flag
+	if signingKeyFlag.Value.String() == "" {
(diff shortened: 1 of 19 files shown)


internal/cmd/common/common.go
```diff
@@ -84,7 +84,14 @@ func LoadSigner(keyBytes []byte) (sslibdsse.SignerVerifier, error) {
 	return signerverifier.NewSignerVerifierFromSecureSystemsLibFormat(keyBytes) //nolint:staticcheck
 }
 
-func CheckIfSigningViable(_ *cobra.Command, _ []string) error {
+func CheckIfSigningViable(c *cobra.Command, _ []string) error {
```
adityasaky, line 87:
```suggestion
func CheckIfSigningViable(cmd *cobra.Command, _ []string) error {
```

for consistency

adityasaky (commented):
LGTM, minor nit. Have you confirmed the overlap in subcommands that has `CheckIfSigningViable` set as `preRun` in terms of whether they all also need `--signing-key`?

adityasaky (approved):
Thanks!

adityasaky:
Were you able to build off this branch and test?

neilnaveen:
trust, rsl, and policy worked for me, and gittuf policy add-key, requires a key, while list-rules passes without needing one

adityasaky:
thanks for checking!

PR #372, opened 2024-05-04
Use rewritten gitinterface

adityasaky:
complementary to #367

### DIFF
--- a/internal/gitinterface/commit.go
+++ b/internal/gitinterface/commit.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"testing"
 	"time"
 
 	"github.com/gittuf/gittuf/internal/signerverifier"
@@ -208,6 +209,37 @@ func (r *Repository) CommitUsingSpecificKey(treeID Hash, targetRef, message stri
 	}
 
 	return commitIDHash, r.CheckAndSetReference(targetRef, commitIDHash, refTip)
+}
+
+// commitWithParents creates a new commit in the repo but does not update any
+// references. It is only meant to be used for tests, and therefore accepts
+// specific parent commit IDs.
+func (r *Repository) commitWithParents(t *testing.T, treeID Hash, parentIDs []Hash, message string, sign bool) Hash { //nolint:unparam
+	args := []string{"commit-tree", "-m", message}
+
+	for _, commitID := range parentIDs {
+		args = append(args, "-p", commitID.String())
+	}
+
+	if sign {
+		args = append(args, "-S")
+	}
+
+	args = append(args, treeID.String())
+
+	now := r.clock.Now().Format(time.RFC3339)
+	env := []string{fmt.Sprintf("%s=%s", committerTimeKey, now), fmt.Sprintf("%s=%s", authorTimeKey, now)}
+
+	stdOut, err := r.executor(args...).withEnv(env...).executeString()
+	if err != nil {
+		t.Fatal(fmt.Errorf("unable to create commit: %w", err))
+	}
+	commitID, err := NewHash(stdOut)
+	if err != nil {
+		t.Fatal(fmt.Errorf("received invalid commit ID: %w", err))
+	}
+
+	return commitID
 }
 
 // ApplyCommit writes a commit object in the repository and updates the
--- a/internal/gitinterface/hash.go
+++ b/internal/gitinterface/hash.go
@@ -27,13 +27,18 @@ type Hash []byte
 
 // String returns the hex encoded hash.
 func (h Hash) String() string {
-	return hex.EncodeToString(h)
+	return hex.EncodeToString(h[:])
 }
 
 // IsZero compares the hash to see if it's the zero hash for either SHA-1 or
 // SHA-256.
 func (h Hash) IsZero() bool {
-	return bytes.Equal(h, zeroSHA1HashBytes[:]) || bytes.Equal(h, zeroSHA256HashBytes[:])
+	return bytes.Equal(h[:], zeroSHA1HashBytes[:]) || bytes.Equal(h[:], zeroSHA256HashBytes[:])
+}
+
+// Equal compares the hash to another provided Hash to see if they're equal.
+func (h Hash) Equal(other Hash) bool {
+	return bytes.Equal(h[:], other[:])
 }
 
 // ZeroHash represents an empty Hash.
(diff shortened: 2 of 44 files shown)


adityasaky:
work pending on some fixes as the other gitinterface PRs changed. However, this is close to ready, I think. Note: there's a pretty gnarly rise in time required for tests in policy, repository, rsl packages. 😬

adityasaky:
Interestingly, the linter didn't complain about the err message strings in #445. I've got that in here.

internal/cmd/rsl/remote/check/check.go
```diff
@@ -12,13 +12,13 @@ import (
...
 
-func (o *options) Run(cmd *cobra.Command, args []string) error {
+func (o *options) Run(_ *cobra.Command, args []string) error {
 	repo, err := repository.LoadRepository()
 	if err != nil {
 		return err
 	}
 
-	hasUpdates, hasDiverged, err := repo.CheckRemoteRSLForUpdates(cmd.Context(), args[0])
+	hasUpdates, hasDiverged, err := repo.CheckRemoteRSLForUpdates(args[0])
```
wlynch, line 21: fine if you want to punt on this for another PR, but we should look to plumb the context through so we can respect cancelled contexts / timeouts.

internal/gitinterface/commit.go
```diff
@@ -210,6 +211,37 @@ func (r *Repository) CommitUsingSpecificKey(treeID Hash, targetRef, message stri
 	return commitIDHash, r.CheckAndSetReference(targetRef, commitIDHash, refTip)
 }
 
+// CommitWithParents creates a new commit in the repo but does not update any
+// references. It is only meant to be used for tests, and therefore accepts
+// specific parent commit IDs.
+func (r *Repository) CommitWithParents(t *testing.T, treeID Hash, parentIDs []Hash, message string, sign bool) Hash {
```
wlynch, line 217: can we unexport this as well? I think it's only being used in the package.

internal/policy/verify.go
```diff
@@ -530,105 +502,83 @@ func verifyEntry(ctx context.Context, repo *git.Repository, policy *State, attes
...
 
-	if len(trustedKeys) == 0 {
-		allKeys, err := policy.PublicKeys()
+	if len(verifiers) == 0 {
+		return nil
+	}
+
+	var authorizationAttestation *sslibdsse.Envelope
+	if attestationsState != nil {
+		authorizationAttestation, err = getAuthorizationAttestation(repo, attestationsState, entry)
```
wlynch, line 518: It looks like `attestationsState` is only used here. Can you add some more context on the difference between these two? My best guess without digging deeper is `attestationsState` is the set of all attestations for the point in time, and `authorizationAttestation` is the attestations that only applies to the current entry.

1. If the above description is correct, could there be multiple attestations for the entry that we might need to handle?
2. We should move this nil-check to getAuthorizationAttestation if it's something that we expect to be done each time.

internal/policy/verify.go
```diff
@@ -530,105 +502,83 @@ func verifyEntry(ctx context.Context, repo *git.Repository, policy *State, attes
...
-			continue
-		}
-		if !errors.Is(err, gitinterface.ErrIncorrectVerificationKey) {
+		} else if !errors.Is(err, ErrVerifierConditionsUnmet) {
 			// Unexpected error
 			return err
 		}
-		// Haven't found a valid key, continue with next key
+		// Haven't found a valid verifier, continue with next key
 	}
```
wlynch, line 537: No action needed for this PR, but something to think about: the entry verify flow is repeated in very similar ways across the different verify funcs. We should see if there are ways to consolidate these so they don't diverge in subtle ways.

internal/policy/verify.go
```diff
@@ -530,105 +502,83 @@ func verifyEntry(ctx context.Context, repo *git.Repository, policy *State, attes
...
+	}
+
+	tagTargetID, err := repo.GetTagTarget(entry.TargetID)
+	if err != nil {
+		return err
+	}
+
+	if entry.TargetID != entryTagRef && entry.TargetID != tagTargetID {
+		return fmt.Errorf("verifying RSL entry failed, tag reference set to unexpected target")
+	}
```
wlynch, line 576: nit: Would it make sense to check these first before potentially more expensive crypto operations? Or do we need to verify the signatures first before we can trust these values?

internal/cmd/rsl/remote/check/check.go
adityasaky: Got it, I'll add it back in and just `_` for now.

internal/gitinterface/commit.go
adityasaky: I originally had it in _test but then needed it in the rsl package. https://github.com/gittuf/gittuf/pull/372/files#diff-a6428d90ac83e051ab984398f3c2d809cab07eaeb1e7a0378a10806065851b9aR669

adityasaky: I could create a commit in a new branch in the rsl package then SetReference() the main branch to that no-parent commit, lmk if that makes sense.

internal/policy/verify.go
adityasaky: Yes agreed, we consolidated some of this in verifier.Verify but we could probably do a consolidation of looping over verifiers as well.

adityasaky: #454

internal/gitinterface/hash.go
```diff
@@ -23,17 +23,17 @@ var (
 // Hash represents a Git object hash. It is a lightweight wrapper around the
 // standard hex encoded representation of a SHA-1 or SHA-256 hash used by Git.
 
-type Hash []byte
```
neilnaveen, line 26: Why are we setting the size of the hash to sha1.Size? Wouldn't this go against hash agility, as specified in the docstring?

wlynch: Good catch! This will truncate the hash down to the sha1 size. The tests don't catch this, since the assertions are only comparing the truncated values.

We should revert this, plus this to the test to make sure we can roundtrip the original hash back to the string we gave it.

```go
assert.Equal(t, test.hash, hash.String())
```

wlynch (changes requested):

internal/gitinterface/hash.go
adityasaky: good catch, this was to simplify Hash comparison elsewhere as well as to be able to use the hash obj as a map key. For now, i'm using the string representation of a hash for the map key and I've added a .Equal() on hash for the comparisons.

adityasaky: @wlynch added the roundtrip check too.

wlynch (approved):

PR #379, opened 2024-05-07
gitinterface: Add representation of a repository and hashes

adityasaky:
This PR adds a lightweight wrapper around a Git repository as well as hashes used in the Git object store. It includes support for both SHA-1 and SHA-256 hashes.


adityasaky:
Blocked by #376

adityasaky:
Linter is unhappy because of the execute with stdin helper, which I will be using in a subsequent PR. I can introduce those helpers then or we can override the linter for now as this is just a part of the bigger PR. cc @wlynch

internal/gitinterface/hash.go
```diff
@@ -0,0 +1,54 @@
...
+
+// IsZero compares the hash to see if it's the zero hash for either SHA-1 or
+// SHA-256.
+func (h Hash) IsZero() bool {
+	return h.hash == zeroSHA1HashString || h.hash == zeroSHA256HashString
+}
+
+// ZeroHash represents an empty Hash.
+// TODO: use SHA-256 zero hash for repositories that have that as the default.
+var ZeroHash = Hash{hash: zeroSHA1HashString}
```
wlynch, line 39: If you use []byte as the base storage, I think this becomes much simpler, since zero sha1 and zero sha256 should effectively both be the same, or differ only by the len of 0 bytes.

wlynch: you'll also save a ton of space if you expect a lot of hashes because of string encoding.

internal/gitinterface/hash.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+var (
+	ErrInvalidHashEncoding = errors.New("hash string is not hex encoded")
+	ErrInvalidHashLength   = errors.New("hash string is wrong length")
+)
+
+// Hash represents a Git object hash. It is a lightweight wrapper around the
+// standard hex encoded representation of a SHA-1 or SHA-256 hash used by Git.
+type Hash struct {
```
wlynch, line 22: Is this used anywhere yet? We may want to hold off on this change for the time being.

internal/gitinterface/repository.go
```diff
@@ -0,0 +1,143 @@
...
+// LoadRepository returns a Repository instance using the current working
+// directory. It also inspects the PATH to ensure Git is installed.
+func LoadRepository() (*Repository, error) {
+	_, err := exec.LookPath(binary)
+	if err != nil {
+		return nil, fmt.Errorf("unable to find Git binary, is Git installed?")
+	}
+
+	repo := &Repository{clock: clockwork.NewRealClock()}
+	envVar := os.Getenv("GIT_DIR")
```
wlynch, line 49: Is this a git convention or our own thing?

internal/gitinterface/repository.go
```diff
@@ -0,0 +1,143 @@
...
+	var (
+		stdOut bytes.Buffer
+		stdErr bytes.Buffer
+	)
+
+	cmd.Stdout = &stdOut
+	cmd.Stderr = &stdErr
+
+	err := cmd.Run()
+	stdOutString := stdOut.String() // sometimes we want the trailing new line (say when `cat-file -p` a blob, leaving it to the caller)
```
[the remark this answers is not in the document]
wlynch: If you want, you can add wrappers that do the string reading for you, e.g.

```go
// executeString executes a command and returns its result as a string.
// This should not be used for operations that may return large blob responses
// (instead use executeCommand).
func (r *Repository) executeString(args ...string) (string, error) {
  r, err := r.executeCommand(args...)
  if err != nil {
    ...
  }

  out, err := io.ReadAll(r)
  if err != nil {
    ...
  }
  return strings.TrimSpace(out), nil
}
```

internal/gitinterface/repository.go
adityasaky: This is sometimes set by Git: https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables

internal/gitinterface/repository.go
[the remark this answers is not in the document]
adityasaky: > You probably want to return the io.ReadClosers from cmd.StdoutPipe/StderrPipe.

I had an issue with these closing before they could be read by the invoking function. I'm now returning the bytes.Buffer and reading from that, but I think we have the problem of storing all of it in memory still?

internal/gitinterface/hash.go
adityasaky: It is used in the next PR onwards for all git IDs, just to sanity check that anything we're passing to a git command (say cat-file or whatever) is a valid SHA.

adityasaky: I'm using this PR to set up the foundations for the object specific APIs introduced in the next PRs in this series

internal/gitinterface/hash.go
adityasaky: Done!

adityasaky:
I've taken out three unused execute functions that I'll introduce with the next one.

internal/gitinterface/repository.go
```diff
@@ -0,0 +1,143 @@
...
+			stdErrString = "error running `git " + strings.Join(args, " ") + "`"
+		}
+	}
+	return stdOutString, stdErrString, err
+}
+
+// executeGitCommandWithStdIn is a helper to execute the specified command in
+// the repository with `stdInContents` passed into the process stdin. It
+// automatically adds the explicit `--git-dir` parameter.
+func (r *Repository) executeGitCommandWithStdIn(stdInContents []byte, args ...string) (string, string, error) {
```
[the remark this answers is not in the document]
adityasaky: I've taken this out for now, but I do have a question of piping the Buffer into stdIn, whether through the provided pipe or directly.

internal/gitinterface/repository.go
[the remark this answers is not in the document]
wlynch: bleh you're right. I guess this is fine for now, but we might want to expose more of the command interface to callers in the future.

Did you want to keep the full stderr, or use the default 32KB error behavior?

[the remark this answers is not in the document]
adityasaky: I think the default is fine for the git errors we'll run into for the commands we're actually using. In my experience, Git's errors aren't too verbose, especially for plumbing tools.

adityasaky:
BTW, I think we should proceed with this series of PRs with an issue to address the windows tmpdir clean up.

wlynch (approved):

PR #383, opened 2024-05-07
gitinterface: Add support for trees, commits, tags, and references

adityasaky:
These are a little intertwined, so I'm submitting them together.

### DIFF
--- a/internal/gitinterface/commit.go
+++ b/internal/gitinterface/commit.go
@@ -5,7 +5,11 @@ package gitinterface
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
+	"os"
+	"strings"
+	"time"
 
 	"github.com/gittuf/gittuf/internal/signerverifier"
 	"github.com/gittuf/gittuf/internal/tuf"
@@ -56,6 +60,50 @@ func Commit(repo *git.Repository, treeHash plumbing.Hash, targetRef string, mess
 	return ApplyCommit(repo, commit, curRef)
 }
 
+// Commit creates a new commit in the repo and sets targetRef's to the commit.
+// This function is meant only for gittuf references, and therefore it does not
+// mutate repository worktrees.
+func (r *Repository) Commit(treeID Hash, targetRef, message string, sign bool) (Hash, error) {
+	currentGitID, err := r.GetReference(targetRef)
+	if err != nil {
+		if !errors.Is(err, ErrReferenceNotFound) {
+			return ZeroHash, err
+		}
+	}
+
+	if err := os.Setenv(committerTimeKey, r.clock.Now().Format(time.RFC3339)); err != nil {
+		return ZeroHash, fmt.Errorf("unable to set committer time: %w", err)
+	}
+	defer os.Unsetenv(committerTimeKey) //nolint:errcheck
+	if err := os.Setenv(authorTimeKey, r.clock.Now().Format(time.RFC3339)); err != nil {
+		return ZeroHash, fmt.Errorf("unable to set author time: %w", err)
+	}
+	defer os.Unsetenv(authorTimeKey) //nolint:errcheck
+
+	args := []string{"commit-tree", "-m", message}
+
+	if !currentGitID.IsZero() {
+		args = append(args, "-p", currentGitID.String())
+	}
+
+	if sign {
+		args = append(args, "-S")
+	}
+
+	args = append(args, treeID.String())
+
+	stdOut, err := r.executeGitCommandString(args...)
+	if err != nil {
+		return ZeroHash, fmt.Errorf("unable to create commit: %w", err)
+	}
+	commitID, err := NewHash(stdOut)
+	if err != nil {
+		return ZeroHash, fmt.Errorf("received invalid commit ID: %w", err)
+	}
+
+	return commitID, r.CheckAndSetReference(targetRef, commitID, currentGitID)
+}
+
 // CommitUsingSpecificKey creates a new commit in the repository for the
 // specified parameters. The commit is signed using the PEM encoded SSH or GPG
 // private key. This function is expected for use in tests and gittuf's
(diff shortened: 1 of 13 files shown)


adityasaky:
Blocked by #376, #379, #380.

neilnaveen (commented):
I was looking at the code, and I just had a nit and a question about reference resolution

internal/gitinterface/references.go
```diff
@@ -0,0 +1,164 @@
...
+	tagName := TagReferenceName(target)
+	_, err = r.GetReference(tagName)
+	if err == nil {
+		return tagName, nil
+	}
+	if !errors.Is(err, ErrReferenceNotFound) {
+		return "", err
+	}
+
+	return "", ErrReferenceNotFound
```
neilnaveen, line 103: How would gittuf be able to handle custom refs?

internal/gitinterface/tag.go
```diff
@@ -65,6 +66,84 @@ func Tag(repo *git.Repository, target plumbing.Hash, name, message string, sign
...
+	if err != nil {
+		return ZeroHash, err
+	}
+
+	goGitRepo, err := r.GetGoGitRepository()
+	if err != nil {
+		return ZeroHash, err
+	}
+
+	targetObj, err := goGitRepo.Object(plumbing.AnyObject, plumbing.NewHash(target.String()))
```
neilnaveen, line 84: Could we use gittuf's NewHash function?

neilnaveen:
Also we will have to continuously support go-git if we want to expose gittuf as an API, if people want to work on git repo's that are not local, and will never exist in a directory on the machine running the code.

internal/gitinterface/references.go
adityasaky: We can't really determine the absolute reference for a shorthand of a custom ref. This is only really useful for when users pass in `main` for `refs/heads/main` and `v1.0.0` for `refs/tags/v1.0.0`.

adityasaky:
(I've marked the windows test as continue-on-error, it doesn't actually pass yet)

internal/gitinterface/references.go
wlynch: https://git-scm.com/docs/gitrevisions#Documentation/gitrevisions.txt-emltrefnamegtemegemmasterememheadsmasterememrefsheadsmasterem documents the behavior of ref resolution - it'd be good if we can match this as close as possible. 

Technically general refs do take precedence over branches. (e.g. given `notes/foo`, `refs/notes/foo` should take priority over `refs/heads/notes/foo`)

internal/gitinterface/tree.go
```diff
@@ -197,3 +278,130 @@ func (t *TreeBuilder) writeTrees(parent string, tree *object.Tree) (plumbing.Has
 
 	return WriteTree(t.r, tree.Entries)
 }
+
+// ReplacementTreeBuilder is used to create multi-level trees in a repository.
+// Based on `buildTreeHelper` in go-git.
+// This will replace TreeBuilder above, and will be renamed at that point.
+type ReplacementTreeBuilder struct {
```
wlynch, line 285: bikeshed: How does this different from the existing TreeBuilder? If it's the multi-levels, would `MultilevelTreeBuilder` be a better name?

wlynch (commented):
Looks good for the most part! Few comments

internal/gitinterface/tree.go
adityasaky: This replaces TreeBuilder; it's largely the same but doesn't use go-git. The plan is:
a) add git binary based APIs to the package
b) update other packages to use the new API (#372)
c) remove old, now unused APIs, also rename some things (ReplacementTreeBuilder -> TreeBuilder, TestRepositoryCommit -> TestCommit, etc.)

internal/gitinterface/commit.go
```diff
@@ -56,6 +60,50 @@ func Commit(repo *git.Repository, treeHash plumbing.Hash, targetRef string, mess
...
+	}
+
+	if err := os.Setenv(committerTimeKey, r.clock.Now().Format(time.RFC3339)); err != nil {
+		return ZeroHash, fmt.Errorf("unable to set committer time: %w", err)
+	}
+	defer os.Unsetenv(committerTimeKey) //nolint:errcheck
+	if err := os.Setenv(authorTimeKey, r.clock.Now().Format(time.RFC3339)); err != nil {
+		return ZeroHash, fmt.Errorf("unable to set author time: %w", err)
+	}
+	defer os.Unsetenv(authorTimeKey) //nolint:errcheck
```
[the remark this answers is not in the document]
adityasaky: yeah, that's a good point. I'll add an r.execute... API to pass in environment vars as well.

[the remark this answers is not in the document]
adityasaky: https://github.com/gittuf/gittuf/pull/383/commits/c9f00e1615c7e56b66cbc5be4d5241e5c120df88

internal/gitinterface/references.go
adityasaky: TIL, very interesting. I've updated this so it checks for just `refs/<name>` first.

internal/gitinterface/repository.go
```diff
@@ -104,7 +104,47 @@ func (r *Repository) executeGitCommandDirectString(args ...string) (string, erro
...
+	return r.executeGitCommandDirectWithEnv(env, args...)
+}
+
+// executeGitCommandDirectWithEnv is a helper to execute the specified command
+// in the repository after setting the provided environment variables. Note that
+// the command inherits os.Environ() first. This helper executes in the current
+// directory without specifying the GIT_DIR explicitly.
+func (r *Repository) executeGitCommandDirectWithEnv(env []string, args ...string) (io.Reader, io.Reader, error) {
 	cmd := exec.Command(binary, args...)
+	cmd.Env = append(os.Environ(), env...)
```
wlynch, line 147: Up to you if you want to do this in this PR or not, but having different funcs for different modes of operation is going to be painful to maintain in the long run (e.g. look how many places we set `--git-dir`). It might be better if we follow the approach of exec.Command and break this into it's own type that can be modified.

e.g.

```go
cmd := r.NewCommand("commit", ...)
cmd.Env = map[string]string{...}
stdout, stderr, err := cmd.Do()
```

or if we want to get fancier (depending on how much validation we need to do)

```go
stdout, stderr, err := r.NewCommand("commit", ...).WithEnv(...).Do()
```

wlynch (approved):

internal/gitinterface/repository.go
adityasaky: Agreed, this was already getting to be a pain to manage / track all the different combinations. I'll update this separately, good suggestion!

PR #397, opened 2024-05-09
Remove an extra gittuf namespace check

neilnaveen:
- Remove an extra gittuf namespace check

### DIFF
--- a/internal/rsl/rsl.go
+++ b/internal/rsl/rsl.go
@@ -768,9 +768,5 @@ func isRelevantGittufRef(refName string) bool {
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


internal/rsl/rsl.go
```diff
@@ -768,9 +768,5 @@ func isRelevantGittufRef(refName string) bool {
 		return false
 	}
 
-	if refName == gittufPolicyStagingRef {
```
adityasaky, line 771: This is necessary, we return gittuf refs except for policy-staging, by design. We might want to rethink how this is filtered, but we can't just take this out right now.

Issue #394, opened 2024-05-09
Auto-test snippets from `get-started.md`


lukpueh:
### Add a description

Following #360 I suggest to add basic testing for the code snippets in `get-started.md`. Maybe Golang has good tooling for this, otherwise, we could take inspiration from [this script](https://github.com/in-toto/demo/blob/927ba96bc398c2170898853b25250bbf46204e63/run_demo_md.py). 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #405, opened 2024-05-15
cmd, ci, docs: Add test script for get-started.md

patzielinski:
This PR adds a test script to test the commands in `docs/get-started.md`, as suggested in #394. (Thanks @lukpueh!)

At the moment, this checks that the return codes of the commands tested are all `0`'s, and raises an error if a non-zero code is returned. I can rework this to match parts of the output if we think that's better/needed (parts since output is variable in places).


neilnaveen:
Don't we already have somthing like this in the gittuf demo repo?

adityasaky:
cc @lukpueh

lukpueh (commented):
This also works, but requires manual syncing with the markdown file. May I ask, why you decided against scraping and running the actual snippets from the markdown as suggested in #394?

lukpueh:
> any change in commands or command output requires updating the script with the expected text anyways.

Sure, but CI will fail if the actual output doesn't match the expected output anymore.

lukpueh:
> Maybe we can add an expected output text file that the script loads and compares to the terminal output so we can just update that file instead and have the script extract commands from the markdown file.

Yes, that's what the script referenced in the issue does.

adityasaky:
timestamps in the debug messages may be a small blocker for that? Perhaps this is a good time to pretty that up as well, though I think it's low priority...

lukpueh:
> This is what I meant about variable output, as the timestamps prevent checking output equivalence. I can just disable the verbose output and match on the small amount of output we have anyways...

Ah right, it's harder if timestamps are in the output. There are several ways to go about this:
- change gittuf to not print timestamps
- run the test with something like `faketime`
- strip timestamps from the output in the tests


> A question though, is it better to have the output embedded in the script or as a separate file? The script referenced appears to do the former.

Whatever you prefer. If it's short enough, like in the in-toto demo, I think it's fine to just add it to the script. If it's a lot longer a separate file is probably better for readability of the script.

adityasaky:
Can we make hashes deterministic by controlling the author/committer details (including the time) and signing with an ed25519 key? You can pin those details using env vars.

lukpueh:
> > Can we make hashes deterministic by controlling the author/committer details (including the time) and signing with an ed25519 key? You can pin those details using env vars.
> 
> I can try that, but do we want to update the commands in the markdown to match? This should be to test the commands as the user would run them with only minor changes imo...

Setting up custom author/committer details and a fake time shouldn't affect the testable snippets. Key type maybe, but I guess that's okay. 

Btw. after #414, rsa keys will produce reproducible signatures too, because we'll no longer sign ourselves (with PSS), but ask `ssh-keygen` to sign, which uses (pkcsv1.5).

docs/testing/tester-expected-macos.txt
```diff
@@ -0,0 +1,29 @@
...
+Signing data on standard input
+Signing data on standard input
++ gittuf policy add-rule -k ../keys/policy --rule-name protect-main --rule-pattern git:refs/heads/main --authorize-key ../keys/developer.pem
+Signing data on standard input
+Signing data on standard input
++ gittuf policy apply
+Signing data on standard input
++ echo 'Hello, world!'
++ git add .
++ git commit -S -m 'Initial commit'
```
lukpueh, line 23: I think the single quotes being added on set -x is a bash thing. Maybe, if you explicitly run the script with bash, you could use the same expected output un ubuntu and macOS.

.github/workflows/get-started-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29
+    - name: Install Go
+      uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7
+      with:
+        go-version: '1.22'
+        cache: true
+    - name: Install Python
+      uses: actions/setup-python@v5
```
adityasaky, line 25: please pin hash, thanks

docs/testing/test-get-started-md.py
```diff
@@ -0,0 +1,95 @@
...
+                snippets[i] = snippet.replace("$ ", "")
+            script = "\nset -x\n{}".format("\n".join(snippets))
+            script += "\ngittuf verify-ref main" # Workaround for non-deterministic hashes
+
+            # Set some environment variables to make git hashes deterministic
+            cmd_env = os.environ.copy()
+
+            cmd_env["GIT_AUTHOR_NAME"] = "Example E"
+            cmd_env["GIT_AUTHOR_EMAIL"] = "example@example.org"
+            cmd_env["GIT_AUTHOR_DATE"] = "Mon, 3 Jun 2024 10:00:00 -0400"
```
adityasaky, line 70: can this be ISO 8601 / rfc 3339?

docs/testing/test-get-started-md.py
```diff
@@ -0,0 +1,95 @@
...
+            cmd_env["GIT_COMMITTER_NAME"] = "Example E"
+            cmd_env["GIT_COMMITTER_EMAIL"] = "example@example.org"
```
adityasaky, line 72: https://github.com/gittuf/gittuf/blob/main/internal/gitinterface/common.go#L17-L20 for consistency, let's use these values? :D

adityasaky:
@lukpueh could you take a look and approve if this is good to go? thanks!

docs/testing/test-get-started-md.py
```diff
@@ -0,0 +1,95 @@
...
+    testing_path = os.path.join(docs_path + "/testing")
+    keys_path = os.path.join(testing_path + "/keys")
+    os.chdir(testing_path)
+
+    # Prepare testing directory
+    match platform.system():
+        case "Linux":
+            expected_output_file = os.path.realpath(os.path.join(testing_path, EXPECTED_OUTPUT_FILENAME))
+        case "Darwin":
+            expected_output_file = os.path.realpath(os.path.join(testing_path, EXPECTED_OUTPUT_FILENAME))
```
lukpueh, line 37: minor nit:

```suggestion
        case "Linux" | "Darwin":
            expected_output_file = os.path.realpath(os.path.join(testing_path, EXPECTED_OUTPUT_FILENAME))
```

lukpueh (approved):
Nice! look great

docs/testing/test-get-started-md.py
```diff
@@ -0,0 +1,93 @@
...
+            raise SystemExit("Windows is not supported at this time.")
+        case _:
+            raise SystemExit("Unknown platform.")
+    
+    get_started_file = os.path.realpath(os.path.join(docs_path, GET_STARTED_FILENAME))
+    tmp_dir = os.path.realpath(tempfile.mkdtemp())
+    os.chdir(tmp_dir)
+
+    # Copy keys used to make git hashes deterministic    
+    repo_path = os.path.join(tmp_dir + "/repo")
```
adityasaky, line 46: os.path.join(tmp_dir, "repo")?

adityasaky (approved):

adityasaky:
Thanks @patzielinski!

PR #406, opened 2024-05-16
refactor verify.go

flandweber:
Just a small refactoring with things I noticed while looking through the verification control flow.

Feel free to close if you disagree with my observations.


flandweber:
Just noticed https://github.com/gittuf/gittuf/pull/361 which goes seems to improve the verification a lot more.

Issue #407, opened 2024-05-16
CI: Windows tmpdir cleanup


patzielinski:
### What happened?

CI fails on Windows due to an issue with the temporary directory not being accessible due to a process keeping it open. This looks like it can be fixed either in the appropriate tests or possibly even the CI configuration.

Raised in #379.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #408, opened 2024-05-17
gitinterface: Fix test failing on Windows

patzielinski:
This PR fixes CI on Windows failing due to a test in `gitinterface` by manually working with the temporary directory.

Addresses #407.

### DIFF
--- a/internal/gitinterface/common.go
+++ b/internal/gitinterface/common.go
@@ -44,11 +44,7 @@ func CreateTestGitRepository(t *testing.T, dir string) *Repository {
 
 	setupSigningKeys(t, keysDir)
 
-	if err := os.Chdir(dir); err != nil {
-		t.Fatal(err)
-	}
-
-	cmd := exec.Command(binary, "init", "-b", "main")
+	cmd := exec.Command(binary, "init", "-b", "main", dir)
 	if err := cmd.Run(); err != nil {
 		t.Fatal(err)
 	}


internal/gitinterface/config_test.go
```diff
@@ -3,18 +3,26 @@
...
+
 	repo := CreateTestGitRepository(t, tmpDir)
 
 	// CreateTestGitRepository sets our test config
 	config, err := repo.GetGitConfig()
 	assert.Nil(t, err)
 	assert.Equal(t, testName, config["user.name"])
 	assert.Equal(t, testEmail, config["user.email"])
+
+	os.Remove(tmpDir)
```
neilnaveen, line 27: Could this be deferred, so that if anything fails, the temp dir is still cleaned up?

neilnaveen (commented):
LGTM, small nit. Thanks

internal/gitinterface/config_test.go
adityasaky: I wonder if we can move the MkdirTemp call into CreateTestGitRepository and have that return the path for when we need the location of the repo. That way, we don't need to remove multiple.

+1 on deferring, also think it might be worth doing RemoveAll to avoid any timing issues with trying to remove a tmp dir that gets separately cleaned up for whatever reason.

neilnaveen: Yes, we should be using RemoveAll, since currently if we wrap the `os.Remove` with an error check, it error's out "remove /var/folders/5f/q255tppn5098m696g3fd1b_m0000gn/T/911169112: directory not empty", which means that nothing is being cleared by `os.Remove` currently.

adityasaky:
https://github.com/gittuf/gittuf/blob/main/internal/gitinterface/common.go#L47-L49

is it because we never reverse this then? Because we can replace that chdir with `git init -b main <dir>` no?

adityasaky (approved):
good catch!

Issue #245, opened 2023-12-22
Track verified states locally


adityasaky:
### Add a description

gittuf does not currently track prior instances of successful verification for branches. For example, if the main branch is verified up to commit A and commits B and C are added, gittuf verification verifies up to commit A again. We should explore how to record _locally only_ that the gittuf client verified up to commit A successfully to indicate that the next instance of the verification workflow can start from that point onwards. In doing so, we must consider how this list of "verified states" could be manipulated by an attacker depending on where they're stored.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

JustinCappos:
Right.  I agree having a local cache of such information is sensible.  

In my view, so long as we end up with a solution that does not introduce any vulnerabilities / issues regardless of the repository / other user behavior, this is a positive.

One other thing we may want to consider is that a user may want to receive this information from another device for some reason.  I think a sensible reason for doing so is when a user moves to a new laptop, they may not want to redo all of the checking for each of their repos.  While this may happen naturally if they copy all file system state over, it may not happen in other cases.  It could also (in theory) be the case that for a large repo like the Linux kernel, one decides to go from some later verification point which has been validated by a very large number of people you trust.  (I'm not arguing this is a good model in all cases, I'm just saying that people will want this to happen, I think.)

adityasaky:
Addressed by #716

PR #411, opened 2024-05-21
(WIP) Add gittuf remote handling protocol

adityasaky:


adityasaky:
I think this is ready for a first round of feedback!

@wlynch @justincappos @lukpueh @patzielinski @neilnaveen

adityasaky:
Note: as of [13e4a00](https://github.com/gittuf/gittuf/pull/411/commits/13e4a004ffa2de083858586d61bd51c1d7c51887), the transport doesn't work. I'm working on making it more resilient to different local conditions.

adityasaky:
Things work again (and are more robust than before). I also tested out some trivial force pushes / fetches and they work as well, though we will need #369 in some form to mark rsl entries for rebased comments as skipped.

neilnaveen:
A feature that would work well with the transport, since it controls all pushes and pulls, is caching the latest verified RSL entry in a specific reference, allowing only the user's signing key to commit to it. This approach would significantly speed up verification. After the initial pull from the main repository, gittuf's impact on the user experience would be minimal, even with large repositories, as long as users pull from their remote regularly.

adityasaky:
Yeah, we definitely need that (#245). This and a couple other things make me wonder if we want to carve out `refs/gittuf-local/` so that it doesn't get synced accidentally.

adityasaky:
the code is still a mess and i want to work on cleaning things up, but the functionality seems fairly stable. This hasn't switched on gittuf verify-ref yet. Seeking feedback and opinions on the best way to structure the curl and ssh helpers!

lukpueh:
Impressive work, Aditya! Your code walkthrough the other day was very helpful, but I must admit, I still find it hard to navigate by myself. 

Maybe you could try to better separate (1) main flow, (2) transport, and (3) gittuf logic. 

IIUC (1) is controlled by the messages received from and sent to the git parent process, where stdin/stdout are the read and write message queues.

By (2) I mean the messages sent to and received from the http transport (curl) or ssh client (ssh) child processes spawned by us, with similar queues.
 
And finally, (3) would be the interposed gittuf logic this feature is actually about, i.e. fetching, verifying, creating and pushing additional refs.

It would be extremely helpful to see at one glance the expected messages in both directions and on both ends, and maybe even their typical sequence. If this is not possible in code, it could be done in documentation.

What could be done in code, would be to reduce the amount of nesting. In that regard, I have one specific question, is the internal `currentState` variable necessary, or would the flow of messages be enough state, e.g. could the main loop have  a single dispatch level based on the message content?

adityasaky:
@lukpueh and I have discussed how to improve this package and make it more maintainable. I propose we do this iteratively, separately, and for now go ahead and merge this as an alpha implementation of the transport.

@wlynch @patzielinski @neilnaveen

internal/git-remote-gittuf/main.go
```diff
@@ -0,0 +1,248 @@
...
+		log("Using ssh helper")
+		handler = handleSSH
+	}
+
+	// var rslTip string
+	// cmd := exec.Command("git", "--git-dir", gitDir, "rev-parse", rsl.Ref) //nolint:gosec
+	// rslTipB, err := cmd.Output()
+	// if err == nil {
+	// 	rslTip = string(bytes.TrimSpace(rslTipB))
+	// }
```
neilnaveen, line 145: Is this supposed to still be here, or is this somthing that should be uncommented later on?

neilnaveen (approved):
This looks good. I haven't actually run it, so I can't vouch for anything; I have just one suggestion.

internal/git-remote-gittuf/main.go
adityasaky: the latter

adityasaky:
Suggestions on the best place to put the docs? We don't get nice auto-gen in docs/cli like with the main cmd package.

PR #412, opened 2024-05-22
Replace path.Join -> filepath.Join

wlynch:
path.Join only uses `/`.
filepath.Join will use `/` or `\` depending on the OS

Might be messing with Windows compatibility. (though not sure)

### DIFF
--- a/internal/repository/hook.go
+++ b/internal/repository/hook.go
@@ -8,7 +8,7 @@ import (
 	"io/fs"
 	"log/slog"
 	"os"
-	"path"
+	"path/filepath"
 )
 
 type ErrHookExists struct {
@@ -42,12 +42,12 @@ func (r *Repository) UpdateHook(hookType HookType, content []byte, force bool) e
 	}
 
 	repoRoot := tree.Filesystem.Root()
-	hookFolder := path.Join(repoRoot, ".git", "hooks")
+	hookFolder := filepath.Join(repoRoot, ".git", "hooks")
 	if err := os.MkdirAll(hookFolder, 0o750); err != nil {
 		return fmt.Errorf("making sure folder exist: %w", err)
 	}
 
-	hookFile := path.Join(hookFolder, string(hookType))
+	hookFile := filepath.Join(hookFolder, string(hookType))
 	hookExists, err := doesFileExist(hookFile)
 	if err != nil {
 		return fmt.Errorf("checking if hookFile '%s' exists: %w", hookFile, err)
--- a/internal/repository/hook_test.go
+++ b/internal/repository/hook_test.go
@@ -4,7 +4,7 @@ package repository
 
 import (
 	"os"
-	"path"
+	"path/filepath"
 	"testing"
 
 	"github.com/go-git/go-git/v5"
@@ -23,7 +23,7 @@ func TestUpdatePrePushHook(t *testing.T) {
 		err = r.UpdateHook(HookPrePush, []byte("some content"), false)
 		require.NoError(t, err)
 
-		hookFile := path.Join(tmpDir, ".git", "hooks", "pre-push")
+		hookFile := filepath.Join(tmpDir, ".git", "hooks", "pre-push")
 		prepushScript, err := os.ReadFile(hookFile)
 		require.NoError(t, err)
 		assert.Equal(t, []byte("some content"), prepushScript)
@@ -36,8 +36,8 @@ func TestUpdatePrePushHook(t *testing.T) {
 		require.NoError(t, err)
 		r := &Repository{r: repo}
 
-		hookDir := path.Join(tmpDir, ".git", "hooks")
-		hookFile := path.Join(hookDir, "pre-push")
+		hookDir := filepath.Join(tmpDir, ".git", "hooks")
+		hookFile := filepath.Join(hookDir, "pre-push")
 		err = os.Mkdir(hookDir, 0o750)
 		require.NoError(t, err)
 		err = os.WriteFile(hookFile, []byte("existing hook script"), 0o700) // nolint:gosec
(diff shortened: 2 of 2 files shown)


internal/attestations/authorization.go
```diff
@@ -143,7 +143,7 @@ func (a *Attestations) GetReferenceAuthorizationFor(repo *git.Repository, refNam
 // ReferenceAuthorizationPath constructs the expected path on-disk for the
 // reference authorization attestation.
 func ReferenceAuthorizationPath(refName, fromID, toID string) string {
-	return path.Join(refName, fmt.Sprintf("%s-%s", fromID, toID))
+	return filepath.Join(refName, fmt.Sprintf("%s-%s", fromID, toID))
```
adityasaky, line 146: This path is used within Git to set up subtrees and eventually a blob. So, path.Join is fine.

internal/attestations/github.go
```diff
@@ -72,5 +72,5 @@ func (a *Attestations) SetGitHubPullRequestAuthorization(repo *git.Repository, e
 // GitHubPullRequestAttestationPath constructs the expected path on-disk for the
 // GitHub pull request attestation.
 func GitHubPullRequestAttestationPath(refName, commitID string) string {
-	return path.Join(refName, commitID)
+	return filepath.Join(refName, commitID)
```
adityasaky, line 75: This path is used within Git to set up subtrees and eventually a blob. So, path.Join is fine.

internal/gitinterface/common.go
```diff
@@ -49,7 +48,7 @@ func CreateTestGitRepository(t *testing.T, dir string) *Repository {
 		t.Fatal(err)
 	}
 
-	repo := &Repository{gitDirPath: path.Join(dir, ".git"), clock: testClock}
+	repo := &Repository{gitDirPath: filepath.Join(dir, ".git"), clock: testClock}
```
adityasaky, line 51: I was wrong on slack. The mix of `\\` and `/` in the test is happening here, so this change is good!

internal/gitinterface/tree.go
```diff
@@ -146,7 +146,7 @@ func (t *TreeBuilder) buildIntermediates(name string, blobID plumbing.Hash) {
 	var fullPath string
 	for _, part := range parts {
 		parent := fullPath
-		fullPath = path.Join(fullPath, part)
+		fullPath = filepath.Join(fullPath, part)
```
adityasaky, line 149: This is constructing Git trees, so I'd roll this back to `path.Join`.

internal/gitinterface/utils.go
```diff
@@ -163,14 +163,14 @@ func RemoteRef(refName, remoteName string) string {
 	case strings.HasPrefix(refName, BranchRefPrefix):
 		// refs/heads/<path> -> refs/remotes/<remote>/<path>
 		rest := strings.TrimPrefix(refName, BranchRefPrefix)
-		remotePath = path.Join(RemoteRefPrefix, remoteName, rest)
+		remotePath = filepath.Join(RemoteRefPrefix, remoteName, rest)
```
adityasaky, line 166: This is a ref, IMO we should just use `/`.

internal/gitinterface/utils.go
```diff
@@ -163,14 +163,14 @@ func RemoteRef(refName, remoteName string) string {
...
 	case strings.HasPrefix(refName, TagRefPrefix):
 		// refs/tags/<path> -> refs/tags/<path>
 		remotePath = refName
 	default:
 		// refs/<path> -> refs/remotes/<remote>/<path>
 		rest := strings.TrimPrefix(refName, RefPrefix)
-		remotePath = path.Join(RemoteRefPrefix, remoteName, rest)
+		remotePath = filepath.Join(RemoteRefPrefix, remoteName, rest)
```
adityasaky, line 173: Same here.

internal/repository/hook.go
```diff
@@ -42,12 +42,12 @@ func (r *Repository) UpdateHook(hookType HookType, content []byte, force bool) e
 	}
 
 	repoRoot := tree.Filesystem.Root()
-	hookFolder := path.Join(repoRoot, ".git", "hooks")
+	hookFolder := filepath.Join(repoRoot, ".git", "hooks")
```
adityasaky, line 45: :+1:

adityasaky: Side note: once the gitinterface PRs land, I'll update this to use the `repo` context to derive git_dir correctly.

internal/repository/hook_test.go
```diff
@@ -23,7 +23,7 @@ func TestUpdatePrePushHook(t *testing.T) {
 		err = r.UpdateHook(HookPrePush, []byte("some content"), false)
 		require.NoError(t, err)
 
-		hookFile := path.Join(tmpDir, ".git", "hooks", "pre-push")
+		hookFile := filepath.Join(tmpDir, ".git", "hooks", "pre-push")
```
adityasaky, line 26: :+1:

internal/repository/hook_test.go
```diff
@@ -36,8 +36,8 @@ func TestUpdatePrePushHook(t *testing.T) {
 		require.NoError(t, err)
 		r := &Repository{r: repo}
 
-		hookDir := path.Join(tmpDir, ".git", "hooks")
-		hookFile := path.Join(hookDir, "pre-push")
+		hookDir := filepath.Join(tmpDir, ".git", "hooks")
```
adityasaky, line 39: :+1: to both lines.

internal/repository/hook_test.go
```diff
@@ -57,8 +57,8 @@ func TestUpdatePrePushHook(t *testing.T) {
...
```
adityasaky, line 60: :+1: to both lines.

internal/tuf/tuf.go
```diff
@@ -196,7 +196,7 @@ func (d *Delegations) AddDelegation(delegation Delegation) {
 // Matches checks if any of the delegation's patterns match the target.
 func (d *Delegation) Matches(target string) bool {
 	for _, pattern := range d.Paths {
-		if ok, _ := path.Match(pattern, target); ok {
+		if ok, _ := filepath.Match(pattern, target); ok {
```
adityasaky, line 199: this should certainly be path.Join as patterns will use `/`. We should likely make this clearer in the rule pattern docs. Also, #313 would change this anyway.

adityasaky (changes requested):
I think some of these changes are good to make. I think within Git itself, we ought to just use `/` regardless of the OS.

adityasaky (approved):

PR #414, opened 2024-05-24
Add new ssh package based on `ssh-keygen`

lukpueh:
* A Key struct, similar to tuf.Key, to be included in TUF metadata (not
  yet implemented), which implements the DSSE Verifier interface, to
  verify signatures created with Signer.

* A Signer struct, which implements the DSSE Signer interface, to create
  signatures using `ssh-keygen` and a path to a key.

* An Import function, to import a Key using `ssh-keygen` and a path to a
  key.

For signing and Key import paths to either public or private, plaintext
or encrypted, rsa, ecdsa or ed25519 keys are supported (akin to git's
user.signingKey configuration).

Also adds basic smoke tests for the `ssh` package, and replaces updates
dsse tests to use this module.


lukpueh:
Note: Tests fail because `ssh-keygen` refuses to consider world-readable signing keys. Since read permissions can't be checked into git, we'll need to fix this in test setup.

internal/signerverifier/ssh/ssh.go
```diff
@@ -0,0 +1,164 @@
...
+	"os/exec"
+	"strings"
+
+	"github.com/hiddeco/sshsig"
+	"golang.org/x/crypto/ssh"
+)
+
+// FIXME: make this the interface for Keys in TUF metadata
+type Key struct {
+	KeyType string
```
adityasaky, line 19: Do we just drop this in favour of scheme if we're stripping out go-sslib after this lands?

internal/signerverifier/ssh/ssh.go
```diff
@@ -0,0 +1,164 @@
...
+	KeyVal  KeyVal
```
adityasaky, line 20: I would have liked to move `Public` here directly but I think that's more painful for compat than dropping KeyType would be :(

internal/signerverifier/ssh/ssh.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (s *Signer) KeyID() (string, error) {
+	return s.Key.keyID, nil
+}
+
+// Import Key from path using ssh-keygen
+// Path can be public or private encrypted or plain key, much like git's
+// user.signingKey config
+func Import(path string) (*Key, error) {
+	cmd := exec.Command("ssh-keygen", "-e", "-f", path)
```
adityasaky, line 93: I don't expect this to change but maybe we explicitly pass in `-m rfc4716` so we don't rely on an upstream default or non-standard ssh-keygen implementation a user may have?

internal/signerverifier/ssh/ssh.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, fmt.Errorf("failed to run command %v: %v", cmd, err)
+	}
+
+	sshPub, err := parseSSH2Key(string(output))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse SSH2 key: %v", err)
+	}
+
+	return &Key{
+		keyID:   ssh.FingerprintSHA256(sshPub),
```
adityasaky, line 105: Do we keep the `sha256` prefix?

internal/signerverifier/ssh/ssh.go
```diff
@@ -0,0 +1,164 @@
...
+// Ignores:
+// - line length constraints
+// - other line termination charcters than "\n"
+// - exact header tag and header value format
+func parseSSH2Key(data string) (ssh.PublicKey, error) {
+
+	header := "---- BEGIN SSH2 PUBLIC KEY ----"
+	footer := "---- END SSH2 PUBLIC KEY ----"
+	lineSep := "\n"
+	commentSep := ":"
```
adityasaky, line 132: nit: I'd make this more generic as there a handful of defined headers + support for extended headers that a user may put in there separately?

adityasaky: Actually, can ssh-keygen accept a pubkey already formatted as rfc4716 and still "convert" to rfc4716? I ran into a little trouble first with the permissions being too open for a private key (despite passing in a public key) and then a libcrypto error (after setting perms to 600). Can we handle a case where a user has an RFC4716 encoded key on disk?

adityasaky:
This looks great! Left some comments from a first look.

internal/signerverifier/ssh/ssh.go
lukpueh: I think we could. I'm currently assigning the value `ssh`, but I could also add that to the scheme (if necessary).

internal/signerverifier/ssh/ssh.go
lukpueh: If we decide to not be compatible with TUF keys, we can do this too. If we do not want to the door to TUF or TUF key tooling (e.g. securesystemslib), we should maybe not drop `KeyType` either.

internal/signerverifier/ssh/ssh.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/hiddeco/sshsig"
+	"golang.org/x/crypto/ssh"
+)
+
+// FIXME: make this the interface for Keys in TUF metadata
+type Key struct {
+	KeyType string
+	KeyVal  KeyVal
+	Scheme  string
+	keyID   string
```
lukpueh, line 22: I'm a bit unsure about what to do with keyID. We want the field on the `Key` type, but not (necessarily) in the serialised metadata format. This means we cannot simply use Golang's neat json struct annotations for de/serialization.

There are a few things we could do:
1. include keyid in serialised metadata
   - pro: simple, can use json annotations
   - con:  not TUF spec compliant 
   - con:  duplicates data, as keys in TUF metadata are stored in a `keyid: key` map
2. write custom json parser (go-tuf does this, [for other reasons](https://github.com/theupdateframework/go-tuf/blob/ef106d86c4d7273f82a5ef680d852fb3224cbb10/metadata/marshal.go#L26-L27))
   - pro: TUF spec compliant
   - con: a bit more code
3. compute keyid on-the-fly (go-tuf [seems to do this](https://github.com/theupdateframework/go-tuf/blob/ef106d86c4d7273f82a5ef680d852fb3224cbb10/metadata/keys.go#L121-L133))
   - pros: of 1 and 2
   - con: does not work for arbitrary keyids

internal/signerverifier/ssh/ssh.go
lukpueh: Also helps code readability

internal/signerverifier/ssh/ssh.go
lukpueh: IMO this is a few bytes more on the wire vs. less complex code.  I don't have strong feelings, but I lean towards keeping it.

internal/signerverifier/ssh/ssh.go
lukpueh: > nit: I'd make this more generic as there a handful of defined headers + support for extended headers that a user may put in there separately?

What do you mean by "make this more generic"? Would you like to parse out the headers? I don't think we need them, do we?

adityasaky: purely in the var name: it's a separator for all headers, not just `comment`. as I said, it's a nit that shouldn't block this :D

lukpueh: > Can we handle a case where a user has an RFC4716 encoded key on disk?

Do you mean to load the key from file? Sure, we can do all sorts of things. But I find it quite attractive to just ask `ssh-keygen` for the key, because it hides so much complexity.

adityasaky: image

Possibly I'm doing something wrong? :sweat_smile:

lukpueh: I think it means that `ssh-keygen` cannot read rfc4716 keys.

lukpueh: Why do you want to support rfc4716 keys?

internal/signerverifier/ssh/ssh.go
adityasaky: I think an earlier version of this computed key IDs on the fly when loading it if we didn't already know it. As we usually have `{keyID: key}`, I think we could just set `key.keyID = keyID`? And this should work for arbitrary too? I'm trying to remember why we changed it, it's possible it actually got a bit messy since the GPG case is effectively an arbitrary key ID but we may be able to handle that better now. Let me dig a bit.

adityasaky: Is 1 the only option that supports arbitrary key IDs trivially or am I mistaken?

lukpueh: Maybe 2 isn't that complex either. I think we just need to implement custom unmarshall functions for the Key containers, and can use json annotations for the rest. I can check out.

adityasaky: Sounds good.

lukpueh:
Just pushed a bunch of code comments and renames (sorry for the force push). Let me fix lint and tests and mark it as ready for review. 

I think we can work on integration with TUF metadata in a follow-up PR.

internal/signerverifier/ssh/ssh.go
```diff
@@ -0,0 +1,171 @@
...
+	Path string
+}
+
+// SSH Signer.Sign() implementation of dsse.Signer interface.
+// Signs using "s.Path" to a public or private, encrypted or plaintext, rsa,
+// ecdsa or ed25519 key file supported by "ssh-keygen". This aligns with the
+// git "user.signingKey" option.
+// https://git-scm.com/docs/git-config#Documentation/git-config.txt-usersigningKey
+func (s *Signer) Sign(_ context.Context, data []byte) ([]byte, error) {
+	cmd := exec.Command("ssh-keygen", "-Y", "sign", "-n", "gittuf", "-f", s.Path) //nolint:gosec
```
lukpueh, line 78: Linter flagged this as: "G204: Subprocess launched with a potential tainted input or cmd arguments"

I think it's fine, because we are not running in a shell. Or should we be concerned?

Btw. the linter did not flag the very [similar line below](https://github.com/gittuf/gittuf/pull/414/files#diff-32b875bd523cb3ae662774771a22b3a4ac06ac2170939d1d7738886160c60e04R101).

adityasaky: Based on the docs for os/exec:

> Unlike the "system" library call from C and other languages, the os/exec package intentionally does not invoke the system shell and does not expand any glob patterns or handle other expansions, pipelines, or redirections typically done by shells. The package behaves more like C's "exec" family of functions. To expand glob patterns, either call the shell directly, taking care to escape any dangerous input, or use the path/filepath package's Glob function. To expand environment variables, use package os's ExpandEnv. 

I think this is fine, we can `nolint` and move on. @wlynch could you confirm?

adityasaky: > Btw. the linter did not flag the very [similar line below](https://github.com/gittuf/gittuf/pull/414/files#diff-32b875bd523cb3ae662774771a22b3a4ac06ac2170939d1d7738886160c60e04R101).

The linter's been hit or miss for me on these...I haven't analyzed what patterns trigger it and what do not with exec.Cmd.

internal/signerverifier/ssh/ssh_test.go
```diff
@@ -0,0 +1,101 @@
...
+		{"ed25519", keyidEd25519},
+		{"ed25519_enc", keyidEd25519},
+		{"ed25519.pub", keyidEd25519},
+	}
+
+	for _, test := range tests {
+		t.Run(test.keyName, func(t *testing.T) {
+			if strings.Contains(test.keyName, "_enc") {
+				if runtime.GOOS == "windows" {
+					t.Skip("TODO: test encrypted keys on windows")
```
lukpueh, line 41: I just spent way too long trying to fix this for Windows, with no success. Depending on the version of OpenSSH -- not the default, and not the one bundled with git either on the GHA runner -- it does indeed work as expected (e.g. with OpenSSH_for_Windows_8.6p1, LibreSSL 3.3.3 installed via choco). Unfortunately, I still can't test it, because the script defined via SSH_ASKPASS fails with a non-explanatory error message:
```
CreateProcessW failed error:193
ssh_askpass: posix_spawnp: Unknown error
```

I suggest to ticketize this issue and move on. That is:

1. Review & merge this PR
2. Integrate the signer/verifier from this PR with TUF metadata
3. Revisit the Windows issue later.

internal/signerverifier/ssh/ssh.go
```diff
@@ -0,0 +1,173 @@
...
+	signature, err := sshsig.Unarmor(sig)
+	if err != nil {
+		return fmt.Errorf("failed to parse ssh signature: %w", err)
+	}
+
+	message := bytes.NewReader(data)
+
+	// ssh-keygen uses sha512 to sign with **any*** key
+	hash := sshsig.HashSHA512
+	if err = sshsig.Verify(message, signature, pub, hash, "gittuf"); err != nil {
```
adityasaky, line 45: hm, this is a problem; we'd hit this to verify a commit or tag signature as well, when the namespace is not in our control. I wonder if the namespace needs to become part of the key and we have to `SetNamespace()` wherever we're using this? It's not super clean; we'd have to special case the SSH verifier, but I don't see another solution atm.

internal/signerverifier/ssh/ssh.go
```diff
@@ -0,0 +1,173 @@
...
+	cmd := exec.Command("ssh-keygen", "-Y", "sign", "-n", "gittuf", "-f", s.Path) //nolint:gosec
+
+	cmd.Stdin = bytes.NewBuffer(data)
+
+	output, err := cmd.Output()
+	if err != nil {
+		return nil, fmt.Errorf("failed to run command %v: %w", cmd, err)
+	}
+
+	return output, nil
```
adityasaky, line 87: So this implements the DSSE.Signer interface, meaning the signature returned here is meant to be embedded in a DSSE envelope primarily. Does this mean we should format the signature here to strip the header and footer? Remove newlines etc? DSSE signature should be b64 encoded which the rest of this is.

Or, we leave it as is and effectively double b64 encode it: https://github.com/gittuf/gittuf/blob/main/internal/signerverifier/dsse/dsse.go#L55.

internal/signerverifier/ssh/ssh.go
```diff
@@ -0,0 +1,173 @@
...
+func parseSSH2Key(data string) (ssh.PublicKey, error) {
+	beginMark := "---- BEGIN SSH2 PUBLIC KEY ----"
+	endMark := "---- END SSH2 PUBLIC KEY ----"
+	lineSep := "\n"
+	headerSep := ":"
+	continues := "\\"
+
+	// Normalize and trim newlines
+	data = strings.ReplaceAll(data, "\r\n", lineSep)
+	data = strings.Trim(data, lineSep)
```
adityasaky, line 144: strings.TrimSpace, alternatively

adityasaky (changes requested):
This is largely in great shape, IMO! I think most of my comments are with test setup, though we also need to work out how to pass in the ssh namespace for the signing and verifying flows.

internal/common/testutils.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// Copy all files from source path to destination path and set permissions
```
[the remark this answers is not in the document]
lukpueh: Ah right, I knew this once. Will update...

internal/signerverifier/ssh/ssh_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"strings"
+	"testing"
+
+	"github.com/gittuf/gittuf/internal/common"
+	"github.com/stretchr/testify/assert"
+)
+
+// Basic smoke test for ssh package for all supported keys
+func TestSSH(t *testing.T) {
+	testSSHKeys := common.TestSSHKeys(t)
```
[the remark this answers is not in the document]
lukpueh: Good idea.

internal/signerverifier/ssh/ssh_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	keyidRSA := "SHA256:ESJezAOo+BsiEpddzRXS6+wtF16FID4NCd+3gj96rFo"
```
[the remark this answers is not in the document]
lukpueh: >  I'm still a bit nervous about exposing this to a user as they can perform some manipulations using key IDs

What kind of manipulations? And why would that be a problem? IMO the system should just provide default unique keyids, but not rely on a specific format.

internal/testartifacts/testdata/scripts/askpass.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env sh
```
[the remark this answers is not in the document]
lukpueh: It did execute in the ssh session on the Windows runner. But I was also wondering if that's related to the mysterious `ssh_askpass: posix_spawnp: Unknown error`.

internal/signerverifier/dsse/dsse_test.go
```diff
@@ -28,49 +24,56 @@ func TestCreateEnvelope(t *testing.T) {
 }
 
 func TestSignEnvelope(t *testing.T) {
-	env, err := createSignedEnvelope()
+	keyPath := path.Join(common.TestSSHKeys(t), "rsa")
```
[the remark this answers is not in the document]
lukpueh: Thanks for the hint. I was wondering how to do this correctly in go. In Python I usually define test artifact paths relative to `__file__`. But using go:embed and then writing the cached bytes to a tmp dir, sounds like a good idea.

internal/testartifacts/testdata/scripts/askpass.sh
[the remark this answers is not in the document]
adityasaky: That was in the unix-like environment, right? I wonder if we want a separate powershell script or something for windows :thinking:

internal/signerverifier/ssh/ssh.go
lukpueh: We could also piggyback on git's namespace, i.e. `git`:
https://github.com/gittuf/gittuf/pull/414

internal/signerverifier/ssh/ssh_test.go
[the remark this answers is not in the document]
adityasaky:
```
Remove Policy key from gittuf root of trust

Usage:
  gittuf trust remove-policy-key [flags]

Flags:
  -h, --help                   help for remove-policy-key
      --policy-key-ID string   ID of Policy key to be removed from root of trust
```

for example. I don't know if it's a problem per se, it just feels like the hash alg is an implementation detail we should not be exposing to the user. I also have a question about consistency: should we use hex encoding to be consistent with gpg key fingerprints?

internal/signerverifier/ssh/ssh.go
lukpueh: A nice! I thought this was only for white space on the same line. Should have read the docs better.

internal/signerverifier/ssh/ssh.go
lukpueh: > i.e., just use git even for attestations? That could certainly work

Yes!

> might be unexpected for any other verifier

I think, we can and should hide the namespace from the verifier. It's a spill over from a different trust model: 
https://github.com/openssh/openssh-portable/blob/8054b906983ceaed01fabd8188d3dac24c05ba39/PROTOCOL.sshsig#L53-L57

AFAICS we don't need to consider a namespace value to determine the eligibility of a signature.

internal/signerverifier/ssh/ssh.go
lukpueh: I lean towards the latter. The DSSE Signer.sign interface asks me to return bytes, so I return bytes. 

Because even if I returned the b64 encoded bytes only, with begin/end marker and newlines stripped, then the current DSSE implementation would still re-b64-encode them.

So, I'd have to strip all that stuff and b64-decode the bytes before I return them, which sounds like a lot of work and room for error. 

Alternatively, we can also change the dsse Signer interface upstream.

lukpueh:
@adityasaky: IMO this is ready to be merged. I can create a ticket for testing encrypted ssh keys on windows.

lukpueh:
> I can create a ticket for testing encrypted ssh keys on windows.

--> https://github.com/gittuf/gittuf/issues/422

internal/signerverifier/dsse/dsse_test.go
```diff
@@ -28,49 +25,58 @@ func TestCreateEnvelope(t *testing.T) {
...
 	}
+
+	assert.Len(t, env.Signatures, 1)
+	assert.Equal(t, "SHA256:oNYBImx035m3rl1Sn/+j5DPrlS9+zXn7k3mjNrC5eto", env.Signatures[0].KeyID)
+
 	env, err = SignEnvelope(context.Background(), env, signer)
 	assert.Nil(t, err)
 	assert.Len(t, env.Signatures, 1)
-	assert.Equal(t, "a0xAMWnJ3Hzf8j2zLFmniyUxV58m2lUprgzDPkJIRUORR4aKlX23WB3teaVMjXLuRKrD5GAMN8NSCR1vaetxBA==", env.Signatures[0].Sig)
+	assert.Equal(t, "SHA256:oNYBImx035m3rl1Sn/+j5DPrlS9+zXn7k3mjNrC5eto", env.Signatures[0].KeyID)
```
adityasaky, line 46: with an RSA key, we could also test the value of the signature, right? Any reason for using ecdsa here? I imagine ed25519 was because of compat reasons.

internal/signerverifier/ssh/ssh_test.go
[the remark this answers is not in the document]
adityasaky: We'll work this out as we integrate this signer / key type into the metadata.

adityasaky (approved):
LGTM, thanks @lukpueh! Just a small question about using ecdsa vs rsa for signing in a test, but I think we can merge regardless.

internal/signerverifier/dsse/dsse_test.go
lukpueh: Yes, but it's a silly and no longer relevant reason related to https://github.com/gittuf/gittuf/pull/414/commits/a7aca9781cbc674a4c200287a802330d4847f2da:

> Depending on the version [and key type], ssh-keygen might require both public and
private key file, for (passwordless) export (-e) and sign (-Y sign)
operations.

TLDR: I switched to ecdsa, when rsa stopped working, after changing the test key file setup to only copy the  private key file. With the ecdsa private key file only it passed on linux/macOS but not on Windows, so ended up copying the key pair anyway.

If you want I can switch back to RSA. But I'd rather do that in a follow-up PR for improved testing for the new ssh key package.

PR #419, opened 2024-05-30
Support the local and remote ref names being different

### LINKED ISSUE #413: gittuf must support a mismatch in remote and local refs
### Add a description

gittuf assumes that a local branch and the remote equivalent of the branch have the same name. This may not be the case, as the refspec allows for specifying a source and destination name (src:dst).

When creating an RSL entry, gittuf ought to support recording the dst ref name in the entry while reading the tip from the src ref name. This would entail an enhancement to `gittuf rsl record`.

When verifying an RSL entry, gittuf ought to support specifying the local name of the ref being verified (which matches what's in the entry). Thus, at the very end, when gittuf inspects the tip of the local reference to see if it matches the latest valid entry for the ref, the explicitly specified local ref name would be used.

_No response_

### Code of Conduct

adityasaky:
Closes #413 

This PR makes two changes. First, when recording an entry in the RSL, it supports the src ref name being different from the dst ref name. For example, `git push origin local-branch-name:remote-branch-name` updates `remote-branch-name` on origin with the tip of `local-branch-name`. So, the RSL entry for this push must record `remote-branch-name` as the ref name updated, which is what everyone else sees, and use `local-branch-name` as the target ID.

Second, the verification workflow now accepts an optional `--remote-ref-name` flag. When this is set, the workflow uses the remote ref name to identify and verify the relevant RSL entries, but finally uses the local ref name to check that the tip matches. We need this for the following case:
```
git fetch origin remote-branch-name:local-branch-name
gittuf verify-ref --remote-ref-name=remote-branch-name local-branch-name
```

I debated whether the flag ought to be inverted, so we `gittuf verify-ref --local-ref-name=local-branch-name remote-branch-name`. This makes more sense from the implementation PoV because we use the remote ref name a lot more with the RSL verification, but to me it appears to be counterintuitive to the user. They ought to be providing the refname they see in their copy of the repository, I think.

### DIFF
--- a/internal/repository/rsl.go
+++ b/internal/repository/rsl.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/gittuf/gittuf/internal/dev"
 	"github.com/gittuf/gittuf/internal/gitinterface"
+	rslopts "github.com/gittuf/gittuf/internal/repository/options/rsl"
 	"github.com/gittuf/gittuf/internal/rsl"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 )
@@ -23,21 +24,42 @@ var (
 
 // RecordRSLEntryForReference is the interface for the user to add an RSL entry
 // for the specified Git reference.
-func (r *Repository) RecordRSLEntryForReference(refName string, signCommit bool) error {
+func (r *Repository) RecordRSLEntryForReference(refName string, signCommit bool, opts ...rslopts.Option) error {
+	options := &rslopts.Options{}
+	for _, fn := range opts {
+		fn(options)
+	}
+
 	slog.Debug("Identifying absolute reference path...")
-	absRefName, err := r.r.AbsoluteReference(refName)
+	refName, err := r.r.AbsoluteReference(refName)
 	if err != nil {
 		return err
 	}
 
-	slog.Debug(fmt.Sprintf("Loading current state of '%s'...", absRefName))
-	refTip, err := r.r.GetReference(absRefName)
+	// Track localRefName to check the expected tip as we may override refName
+	localRefName := refName
+
+	if options.RefNameOverride != "" {
+		// dst differs from src
+		// Eg: git push <remote> <src>:<dst>
+		slog.Debug("Name of reference overridden to match remote reference name, identifying absolute reference path...")
+		refNameOverride, err := r.r.AbsoluteReference(options.RefNameOverride)
+		if err != nil {
+			return err
+		}
+
+		refName = refNameOverride
+	}
+
+	// The tip of the ref is always from the localRefName
+	slog.Debug(fmt.Sprintf("Loading current state of '%s'...", localRefName))
+	refTip, err := r.r.GetReference(localRefName)
 	if err != nil {
 		return err
 	}
 
 	slog.Debug("Checking for existing entry for reference with same target...")
-	isDuplicate, err := r.isDuplicateEntry(absRefName, refTip)
+	isDuplicate, err := r.isDuplicateEntry(refName, refTip)
 	if err != nil {
 		return err
 	}
@@ -49,20 +71,25 @@ func (r *Repository) RecordRSLEntryForReference(refName string, signCommit bool)
 	// signCommit must be verified for the refName in the delegation tree.
 
 	slog.Debug("Creating RSL reference entry...")
-	return rsl.NewReferenceEntry(absRefName, refTip).Commit(r.r, signCommit)
+	return rsl.NewReferenceEntry(refName, refTip).Commit(r.r, signCommit)
 }
 
 // RecordRSLEntryForReferenceAtTarget is a special version of
 // RecordRSLEntryForReference used for evaluation. It is only invoked when
 // gittuf is explicitly set in developer mode.
-func (r *Repository) RecordRSLEntryForReferenceAtTarget(refName string, targetID string, signingKeyBytes []byte) error {
+func (r *Repository) RecordRSLEntryForReferenceAtTarget(refName, targetID string, signingKeyBytes []byte, opts ...rslopts.Option) error {
 	// Double check that gittuf is in developer mode
 	if !dev.InDevMode() {
 		return dev.ErrNotInDevMode
 	}
 
+	options := &rslopts.Options{}
+	for _, fn := range opts {
+		fn(options)
+	}
+
 	slog.Debug("Identifying absolute reference path...")
-	absRefName, err := r.r.AbsoluteReference(refName)
+	refName, err := r.r.AbsoluteReference(refName)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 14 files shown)


adityasaky:
On the first change, I briefly considered whether the RSL entry ought to also record the local branch name. I ultimately decided against this because it doesn't seem to have any value to _anyone_ else performing verification as they don't actually see a branch with that name. IMO, the user's choice to have some other name is a detail we don't particularly care about, but I may be missing some scenarios where it does matter. From my own experience, this comes up when I branch off locally to try new things, debug CI failures and so on. So a `cool-feature` branch becomes `cool-feature-debug-windows` or something, and when I have a fix, I sometimes just push `cool-feature-debug-windows:cool-feature`.

internal/repository/verify.go
```diff
@@ -23,59 +23,87 @@ import (
...
 
-	slog.Debug(fmt.Sprintf("Verifying gittuf policies for '%s'", target))
+	if remoteRefName == "" {
+		remoteRefName = localRefName
+	} else {
+		// remote ref name is different
+		// **We must consider RSL entries that have remoteRefName rather than
+		// localRefName**
+		slog.Debug("Name of reference differs on remote, identifying absolute reference path...")
+		remoteRefName, err = gitinterface.AbsoluteReference(r.r, remoteRefName)
```
adityasaky, line 45: One issue here: if the user only provides a short name, this will try and identify the full ref name based on what's in the local copy of the repository, and that'll fail as the remote ref is expressly called something else here. Two possible solutions:
a) we require the user to give us the full ref path
b) we assume it's a branch and/or try to see if we can prefix match what the localRefName has


Between the two, I lean towards the first solution, especially since the invocation of verify-ref is likely to be hidden away from the user with the transport or via command compatibility. At the transport level especially, we automatically have the full ref path.

The second can be really helpful in some cases, but can fall apart in confusing ways in others, I think. We can also add the second later after careful consideration but it'd be difficult to remove it...

neilnaveen: I believe that in any gittuf verification process, it's crucial to have steps involved, even if it means using a more complex method. It's important to know exactly what reference you are verifying, especially when it's based on pat clarity on the tern matching. Ambiguity in this process could lead to critical issues, such as verifying "refs/heads/f1" instead of "refs/f1." This level of clarity is essential regardless of whether it's being used in the transport or not. @patzielinski @adityasaky, do you think any other places in verification have any ambiguity? I know that VerifyTag reference uses AbsoulteRef, too, which could lead to the problems stated above.

adityasaky: AbsoluteRef is inherently not a problem, IMO, when it's user facing, as it expands, for eg, `main` to `refs/heads/main` _only if it's actually in the repository_. The second option I proposed above would just "guess" that `some-branch` is in fact `refs/heads/some-branch` as it's not actually in the repository.

adityasaky: the PR as it stands is in option A so I suggest we proceed. thanks for the discussion!

internal/repository/verify.go
```diff
@@ -22,47 +22,61 @@ import (
 // another is to create a new RSL entry for the current state.
 var ErrRefStateDoesNotMatchRSL = errors.New("Git reference's current state does not match latest RSL entry") //nolint:stylecheck
 
-func (r *Repository) VerifyRef(ctx context.Context, target string, latestOnly bool) error {
+func (r *Repository) VerifyRef(ctx context.Context, localRefName, remoteRefName string, latestOnly bool) error {
```
wlynch, line 25: I generally consider having funcs where you need to pass in zero values as an antipattern, since it's not clear whether these values are actually required or not.

I would recommend moving to variadic options for these if they are not actually required.

wlynch, line 25: nit: IIUC, this isn't really a remote ref, even though it's most likely going to be used in support of remote refs?

i.e. This is really verifying a local ref with a name override. This doesn't actually interact with a remote in any way. If this does interact with a remote, how do we know which remote to work with?

internal/repository/verify.go
```diff
@@ -22,47 +22,61 @@ import (
...
+		expectedTip, err = policy.VerifyRefFull(ctx, r.r, remoteRefName)
 	}
 	if err != nil {
 		return err
 	}
 
+	// To verify the tip, we _must_ use the localRefName
 	slog.Debug("Verifying if tip of reference matches expected value from RSL...")
-	if err := r.verifyRefTip(target, expectedTip); err != nil {
+	if err := r.verifyRefTip(localRefName, expectedTip); err != nil {
```
wlynch, line 63: What's the difference between VerifyRef and verifyRefTip, and why is okay to replace the ref name for one and not the other?

(probably worth adding context here)

internal/repository/verify.go
adityasaky: yes, it doesn't interact with the remote, this is about the name for the ref used on the remote, which maps to the name used in the RSL entry for pushes to that ref. I could rename this as a nameOverride or so?

adityasaky: I wonder if that can lead to confusion in the right order to pass in the strings? We'd also have to do length check for the variadic slice. Would a refToBeVerified opts struct make sense instead? Should we consider moving to opts structs more generally?

internal/repository/verify.go
adityasaky: verifyRefTip just checks the HEAD of the ref in the local repository to see that it matches the latest RSL entry for that ref. I can add the context as a comment.

adityasaky: added a docstring for verifyRefTip.

internal/repository/verify.go
adityasaky: I've renamed this to be refName and refNameOverride and changed the logic so we actually _use_ refName throughout, while using the override to update the refName value if needed.

adityasaky: I've also changed the debug message to be a bit clearer.

wlynch: Have an example where this might be confusing?

typically you'd have an internal opts struct that you populate based on the variadic options. e.g. https://go.dev/play/p/YvJLymVzEmX

adityasaky: Got it, I misunderstood your original comment with variadic "options"

adityasaky: done!

internal/repository/rsl.go
```diff
@@ -23,21 +23,34 @@ var (
 
 // RecordRSLEntryForReference is the interface for the user to add an RSL entry
 // for the specified Git reference.
-func (r *Repository) RecordRSLEntryForReference(refName string, signCommit bool) error {
+func (r *Repository) RecordRSLEntryForReference(srcRefName, dstRefName string, signCommit bool) error {
```
wlynch, line 26: I'd make a similar options change for Record as well.

internal/repository/verify.go
```diff
@@ -22,47 +22,93 @@ import (
 // another is to create a new RSL entry for the current state.
 var ErrRefStateDoesNotMatchRSL = errors.New("Git reference's current state does not match latest RSL entry") //nolint:stylecheck
 
-func (r *Repository) VerifyRef(ctx context.Context, target string, latestOnly bool) error {
+type verifyOptions struct {
```
wlynch, line 25: Looks great! Thanks for adding this.

wlynch (commented):
Almost there!

internal/repository/rsl.go
adityasaky: done!

adityasaky: i split them up into different opts packages to avoid a nam collision for the With... functions.

wlynch (approved):

PR #420, opened 2024-05-30
Ignore generated CLI docs in PR diff

adityasaky:
From: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

They render as follows now:
image

So we can still opt in to see specific changes, but IMO this makes reviewing bigger changes easier.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+docs/cli/*.md linguist-generated=true


adityasaky:
The message is different from large diffs. This says "Some generated files are not rendered by default." which is also what you see for codegen from protoc for example.

wlynch (approved):

PR #421, opened 2024-05-30
(WIP) test windows askpass

adityasaky:
ignore in favour of #414


adityasaky:
ah idk if this handles the env var issue on windows; also, AIUI the password will still fail because Write-Output necessarily adds a "\n" but getting to that stage would be a win I think

PR #429, opened 2024-06-05
Add ssh Key <-> SSlibKey conversion functions

lukpueh:
Add functions to convert an ssh Key to an SSlibKey and vice versa.

This can be used to include an ssh Key in TUF metadata (which expects
an SSlibKey) and to again extract it for ssh signature verification.

A basic unit test and a TUF metadata usage example (as test) is included.

### DIFF
--- a/internal/signerverifier/ssh/ssh.go
+++ b/internal/signerverifier/ssh/ssh.go
@@ -177,8 +177,8 @@ func parseSSH2Key(data string) (ssh.PublicKey, error) {
 	return parseSSH2Body(body)
 }
 
-// SSHKeyToSSlibKey converts an ssh Key into an SSlibKey.
-func SSHKeyToSSlibKey(key *Key) *sv.SSLibKey {
+// KeyToSSlibKey converts an ssh Key into an SSlibKey.
+func KeyToSSlibKey(key *Key) *sv.SSLibKey {
 	return &sv.SSLibKey{
 		KeyID:   key.keyID,
 		Scheme:  key.Scheme,
@@ -189,8 +189,8 @@ func SSHKeyToSSlibKey(key *Key) *sv.SSLibKey {
 	}
 }
 
-// SSHKeyToSSlibKey converts an SSlibKey into an ssh Key.
-func SSlibKeyToSSHKey(key *sv.SSLibKey) (*Key, error) {
+// KeyFromSSlibKey converts an SSlibKey into an ssh Key.
+func KeyFromSSlibKey(key *sv.SSLibKey) (*Key, error) {
 	if key.KeyType != SSHKeyType {
 		return nil, fmt.Errorf("cannot convert key of type '%s'", key.KeyType)
 	}
--- a/internal/signerverifier/ssh/ssh.go
+++ b/internal/signerverifier/ssh/ssh.go
@@ -10,11 +10,13 @@ import (
 	"os/exec"
 	"strings"
 
+	sv "github.com/gittuf/gittuf/internal/third_party/go-securesystemslib/signerverifier"
 	"github.com/hiddeco/sshsig"
 	"golang.org/x/crypto/ssh"
 )
 
 const SSHSigNamespace = "git"
+const SSHKeyType = "ssh"
 
 // Key is a container and dsse.Verifier implementation for SSH keys.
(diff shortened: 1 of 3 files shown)


internal/tuf/tuf_test.go
```diff
@@ -245,3 +252,67 @@ func TestDelegationMatches(t *testing.T) {
 		assert.Equal(t, test.expected, got, fmt.Sprintf("unexpected result in test '%s'", name))
 	}
 }
+
+func TestRootMetadataWithSSHKey(t *testing.T) {
```
lukpueh, line 256: This test function is a bit long and maybe too ssh-specific for `tuf_test.go`. I just wanted to include a metadata integration POC / usage example. Happy to drop the commit, if this feels misplaced.

internal/signerverifier/ssh/ssh.go
```diff
@@ -10,11 +10,13 @@ import (
 	"os/exec"
 	"strings"
 
+	sv "github.com/gittuf/gittuf/internal/third_party/go-securesystemslib/signerverifier"
 	"github.com/hiddeco/sshsig"
 	"golang.org/x/crypto/ssh"
 )
 
 const SSHSigNamespace = "git"
+const SSHKeyType = "ssh"
```
adityasaky, line 19:
```
const (
    SSHSigNamespace = ...
    SSHKeyType = ...
)
```

internal/signerverifier/ssh/ssh.go
```diff
@@ -174,3 +176,30 @@ func parseSSH2Key(data string) (ssh.PublicKey, error) {
 	body := strings.Join(lines[i:], "")
 	return parseSSH2Body(body)
 }
+
+// KeyToSSlibKey converts an ssh Key into an SSlibKey.
+func KeyToSSlibKey(key *Key) *sv.SSLibKey {
```
adityasaky, line 181: is there a reason for this to not be a method on `Key`? `key.SSLibKey()`

internal/signerverifier/ssh/ssh.go
```diff
@@ -174,3 +176,30 @@ func parseSSH2Key(data string) (ssh.PublicKey, error) {
...
+		Scheme:  key.Scheme,
+		KeyType: key.KeyType,
+		KeyVal: sv.KeyVal{
+			Public: key.KeyVal.Public,
+		},
+	}
+}
+
+// KeyFromSSlibKey converts an SSlibKey into an ssh Key.
+func KeyFromSSlibKey(key *sv.SSLibKey) (*Key, error) {
```
adityasaky, line 193:
```suggestion
func NewKeyFromSSlibKey(key *sv.SSLibKey) (*Key, error) {
```

internal/tuf/tuf_test.go
adityasaky: I think it's fine tbh. We can revisit when we have other signers / key types we want to test similarly.

internal/signerverifier/ssh/ssh.go
lukpueh: No reason. I can change it.

lukpueh:
I'd like to give some additional context for this PR. My original intention for ssh Key was to satisfy both the dsse.Verifier interface, and a hypothetical "Key-in-TUF-metadata" interface (currently this is the SSlibKey struct).

This idea is modeled after the python-tuf/secureystemslib Key API design, which provides a Key interface, defining the metadata container (keyid, keytype, scheme, keyval), de/serialization methods, and a verify method. To instantiate a concrete Key, we call a factory function, when deserializing TUF metadata. The factory looks up a global map of keytype/scheme pairs to Key implementations. This allows adding support for new keys without changing existing code. The key types just need to be registered in the global map.

I realize now that this idea does not fit well in the current `gittuf` design, where:
1. Key and Verifier are separated
2. Key-to-Verifier conversion is performed explicitly in `New*VerifierFrom*Key` functions

It seems easier to update my code to fit the `gittuf` model, than the other way around. Also, separating Key and Verifier does not contradict dynamic dispatch via global map. It would just happen at a different time.

I'll propose another PR on top of this one.

lukpueh:
Superseded by #433

PR #430, opened 2024-06-05
tuf: Add metadata version to gittuf metadata

patzielinski:
This PR adds a version field to the gittuf metadata to indicate what gittuf should expect when parsing metadata. This is useful for cases such as #326, where the underlying metadata is upgraded with a breaking change to support teams.

The TUF metadata has a `metadata_version` JSON property added for the root and targets metadata to signal what version said metadata aligns with.

### DIFF
--- a/internal/tuf/tuf.go
+++ b/internal/tuf/tuf.go
@@ -18,6 +18,8 @@ import (
 	"github.com/secure-systems-lab/go-securesystemslib/cjson"
 )
 
+const currentMetadataVersion = 0
+
 var (
 	ErrTargetsNotEmpty = errors.New("`targets` field in gittuf Targets metadata must be empty")
 )
@@ -83,6 +85,7 @@ type Role struct {
 // RootMetadata defines the schema of TUF's Root role.
 type RootMetadata struct {
 	Type                   string          `json:"type"`
+	MetadataVersion        int             `json:"metadata_version"`
 	Expires                string          `json:"expires"`
 	Keys                   map[string]*Key `json:"keys"`
 	Roles                  map[string]Role `json:"roles"`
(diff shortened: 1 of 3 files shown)


internal/tuf/tuf.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/secure-systems-lab/go-securesystemslib/cjson"
 )
 
+const currentMetadataVersion = 1
```
adityasaky, line 21: let's start with 0, this should allow for "automatic" versioning of existing metadata, as the new field will have a default zero value of `0` due to being unset.

docs/metadata-versions.md
```diff
@@ -0,0 +1,23 @@
...
+signal to gittuf what to expect when parsing the underlying data.
+
+The version of the metadata can be found in the JSON property
+`metadata_version`, located in the root and targets metadata of the TUF metadata
+used in gittuf.
+
+Different versions of metadata do not necessarily imply breaking changes, but
+there may be such changes present (i.e. V0 -> V1).
+
+## Version 0 (Legacy)
```
adityasaky, line 14: We have to think carefully of how this is communicated. For each version, I'd prefer to have schemas / migration descriptions so there's clarity on the changes.

adityasaky (commented):
Can you create a repo with gittuf metadata without this change, then use a build _with_ this change to update some metadata to confirm the field gets set explicitly? If you update the current to be `0` it should work with no additional changes, though with `1` my expectation is you'll see it be set to `0`.

JustinCappos:
Semver seems to be the standard / expected.  

One concern is whether it would match the gittuf client version number.  

I would argue for something that clearly doesn't match the client version (and is hard to confuse with it) but which follows semver.  Like spec 100.0.0

PR #436, opened 2024-06-13
(WIP) Wire up ssh-keygen signer to CLI

adityasaky:
This PR isn't there yet; it's highlighting some bugs. I ran into trouble getting this to work with gittuf/demo (even with some changes to the script to handle pem -> pub, etc). One thing in particular may be linked to key ID computation, though this might also mean the demo script needs to be updated.

Also, more generally, we may want to compute the key ID the old way for consistency with metadata that already exists (dogfood). It should be internally consistent--this should not break existing policies--but it can be hard to inspect and map old key ID to new key ID if necessary.


adityasaky:
cc @lukpueh

internal/cmd/common/common.go
```diff
@@ -77,12 +80,7 @@ func LoadPublicKey(key string) (*tuf.Key, error) {
 			},
 		}
 	default:
-		kb, err := os.ReadFile(key)
-		if err != nil {
-			return nil, err
-		}
-
-		keyObj, err = tuf.LoadKeyFromBytes(kb)
+		keyObj, err = ssh.NewKeyFromFile(key)
```
adityasaky, line 83: BREAKING CHANGE: we also end up with new key IDs for the same keys as before, I think.

internal/gitinterface/commit.go
```diff
@@ -291,7 +292,7 @@ func (r *Repository) verifyCommitSignature(ctx context.Context, commitID Hash, k
 		}
 
 		return nil
-	case signerverifier.RSAKeyType, signerverifier.ECDSAKeyType, signerverifier.ED25519KeyType:
+	case signerverifier.RSAKeyType, signerverifier.ECDSAKeyType, signerverifier.ED25519KeyType, ssh.SSHKeyType:
```
adityasaky, line 295: This should eventually drop the explicit RSA, ECDSA, etc types. We want just SSH here.

internal/gitinterface/tag.go
```diff
@@ -241,7 +242,7 @@ func (r *Repository) verifyTagSignature(ctx context.Context, tagID Hash, key *tu
 		}
 
 		return nil
-	case signerverifier.RSAKeyType, signerverifier.ECDSAKeyType, signerverifier.ED25519KeyType:
+	case signerverifier.RSAKeyType, signerverifier.ECDSAKeyType, signerverifier.ED25519KeyType, ssh.SSHKeyType:
```
adityasaky, line 245: same here

internal/repository/root.go
```diff
@@ -23,10 +24,18 @@ func (r *Repository) InitializeRoot(ctx context.Context, signer sslibdsse.Signer
...
-	rawKey := signer.Public()
-	publicKey, err := sslibsv.NewKey(rawKey)
-	if err != nil {
-		return err
+	var (
+		publicKey *sslibsv.SSLibKey
+		err       error
+	)
+	if signerSSH, isSSHSigner := signer.(*ssh.Signer); isSSHSigner {
+		publicKey = signerSSH.MetadataKey()
```
adityasaky, line 32: need to work this out a bit; I wonder if when we add GPG and Sigstore, we just extend the signer interface in gittuf to include MetadataKey. Right now, the else is only reachable in tests because of the legacy format if I'm not mistaken. We can't hit it from the CLI with this PR.

internal/signerverifier/signerverifier.go
```diff
@@ -40,14 +41,16 @@ type keyVal struct {
 // keys. While this is called signerverifier, tuf.Key only supports public keys.
 //
 // Deprecated: Switch to upstream key loading APIs.
-func NewSignerVerifierFromTUFKey(key *tuf.Key) (dsse.SignerVerifier, error) {
+func NewSignerVerifierFromTUFKey(key *tuf.Key) (dsse.Verifier, error) {
```
adityasaky, line 44: rename func too? I think we could undeprecate this tbh.

internal/signerverifier/ssh/ssh.go
```diff
@@ -56,10 +60,14 @@ func (v *Verifier) Public() crypto.PublicKey {
 	return v.sshKey.(ssh.CryptoPublicKey).CryptoPublicKey()
 }
 
+func (v *Verifier) MetadataKey() *sv.SSLibKey {
```
adityasaky, line 63: possibly a better way to do this? it's a bit wasteful to have both sshKey and sslibKey but I wonder if it's cheap enough to avoid the reverse conversion as well.

internal/signerverifier/ssh/ssh.go
```diff
@@ -56,10 +60,14 @@ func (v *Verifier) Public() crypto.PublicKey {
...
+	return v.sslibKey
+}
+
 // Signer is a dsse.Signer implementation for SSH keys.
 type Signer struct {
-	Verifier *Verifier
-	Path     string
+	Path string
+	*Verifier
```
adityasaky, line 70: This way, Signer inherits Verifier's methods.

internal/signerverifier/ssh/ssh.go
```diff
@@ -101,14 +104,22 @@ func NewKeyFromFile(path string) (*sv.SSLibKey, error) {
...
 		KeyID:   ssh.FingerprintSHA256(sshPub),
 		KeyType: SSHKeyType,
 		Scheme:  sshPub.Type(),
 		KeyVal: sv.KeyVal{
 			Public: base64.StdEncoding.EncodeToString(sshPub.Marshal()),
 		},
-	}, nil
+	}
+
+	keyID, err := calculateKeyID(key)
```
adityasaky, line 116: I made this change first, but we may not need it with the addition of MetadataKey(). I'm still not sure the best way to handle key IDs, we should discuss @lukpueh

internal/signerverifier/ssh/ssh.go
```diff
@@ -179,3 +191,20 @@ func parseSSH2Key(data string) (ssh.PublicKey, error) {
 	body := strings.Join(lines[i:], "")
 	return parseSSH2Body(body)
 }
+
+func calculateKeyID(k *sv.SSLibKey) (string, error) {
```
adityasaky, line 195: copied from the tuf package; we can maybe remove key APIs there and move them all to signerverifier? especially if we make MetadataKey() a thing here.

adityasaky (commented):
@lukpueh lmk what you think, thanks! This change is internally consistent enough to run gittuf/demo correctly now.

internal/repository/root.go
lukpueh: I do recommend to add the public key to the `Signer` interface. See https://github.com/secure-systems-lab/securesystemslib/issues/605

internal/signerverifier/ssh/ssh.go
lukpueh: Ah, right, it's the embedding thing. Had forgotten about that.

internal/signerverifier/ssh/ssh.go
lukpueh: I'd like to avoid this, because it implies that the two fields are in-sync, which is likely but not guaranteed.

IMO the most practical solution would be to have a struct that satisfies Key and Verifier interface, as in my original design, prior to https://github.com/gittuf/gittuf/pull/433. But that didn't fit in with the existing use of Key, Verifier, and Signer, and would thus require a more invasive patch.

Until then I recommend to ad-hoc create the SSlibKey from the contained sshKey . The code is already there, you could carve it out into a helper: 
https://github.com/gittuf/gittuf/blob/bcbb06557f29f4417a792877edb272b8a83a441c/internal/signerverifier/ssh/ssh.go#L107-L114 

(modulo passing the stored keyid, instead of regenerating it)

internal/signerverifier/ssh/ssh.go
lukpueh: Why should it be necessary?

You can take an ssh key with arbitrary keyid, create a verifier with `NewVerifierFromKey`, and verify a matching signature. So, if you are worried about verifying existing metadata, the keyid format should not be a problem.

What would be a problem is,  if `NewKeyFromFile` must compute the same default keyid as previous function that loaded the same key from file. But I don't think that's a necessary requirement.

lukpueh (commented):
Looks great. I only have vetos about:
- Verifier containing sshPub and sslibPub, and
- requiring cjson for the default keyid

internal/gitinterface/commit.go
adityasaky: Yeah, I think we should do that too.

internal/repository/root.go
adityasaky: `Public()` itself is on the Signer since it's on the Verifier now. The question is whether we expect it to return crypto.PublicKey or the metadata-specific format I've got with MetadataKey()?

internal/signerverifier/ssh/ssh.go
adityasaky: Okay, I've used this in MetadataKey, but I've still retained that method for now.

internal/signerverifier/ssh/ssh.go
adityasaky: Yeah, especially since our key IDs will change _regardless_ for the same ssh key we may have used before. I've dropped this and gone back to ssh fingerprint. It's also nice that we use the ssh fingerprint here, gpg fingerprint for gpg etc, this way.

internal/signerverifier/ssh/ssh.go
```diff
@@ -56,10 +56,19 @@ func (v *Verifier) Public() crypto.PublicKey {
 	return v.sshKey.(ssh.CryptoPublicKey).CryptoPublicKey()
 }
 
+func (v *Verifier) MetadataKey() *sv.SSLibKey {
+	return &sv.SSLibKey{
+		KeyID:   v.keyID,
+		KeyType: SSHKeyType,
+		Scheme:  v.sshKey.Type(),
+		KeyVal:  sv.KeyVal{Public: base64.StdEncoding.EncodeToString(v.sshKey.Marshal())},
+	}
```
lukpueh, line 65: IMO it would be a good idea to factor this into a helper to make sure that `MetadataKey` and `NewKeyFromFile` return the same Key.  The helper can have an optional keyid argument, to pass `v.KeyID` when called here, or compute a default, when called from `NewKeyFromFile`.

internal/repository/root.go
lukpueh: IMO we don't need `Public()`, if `MetadataKey()` is on the interface:

https://github.com/gittuf/gittuf/blob/e246f934a81279512b3676814d0bae8780af93d2/internal/signerverifier/ssh/ssh.go#L54-L55

internal/cmd/common/common.go
```diff
@@ -94,13 +92,26 @@ func LoadPublicKey(key string) (*tuf.Key, error) {
...
+			return nil, err
+		}
+		verifier, err := ssh.NewVerifierFromKey(keyObj)
+		if err != nil {
+			return nil, err
+		}
+		return &ssh.Signer{
+			Verifier: verifier,
+			Path:     key,
+		}, nil
```
lukpueh, line 113: A `ssh.NewSignerFromFile` for these 10 lines might be helpful.

lukpueh (commented):
Thanks for the changes. This looks really good.

internal/repository/root.go
adityasaky: let's resolve this upstream separately?

docs/testing/test-get-started-md.py
```diff
@@ -47,8 +47,9 @@ def test_commands():
     repo_keys_path = os.path.join(tmp_dir, "keys")
     shutil.copytree(keys_path, repo_keys_path)
     os.chdir(repo_keys_path)
+    os.chmod("root", 0o0600)
+    os.chmod("policy", 0o0600)
```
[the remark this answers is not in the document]
adityasaky: @lukpueh should we copy the specified key to a tmp dir when signing so we can enforce perms etc for ssh? too overkill?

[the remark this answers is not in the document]
lukpueh: What about including the `keygen` snippet in the tests and using ad-hoc generated keys? Advantages:
- increased test coverage
- reduced test setup (no key copying, etc.)
- ssh-keygen takes care of correct permissions

RSA might slow down the tests a bit. But we can switch back to ecdsa now that we have the new ssh Signer.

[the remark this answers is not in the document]
adityasaky: @patzielinski could you tackle that?

internal/cmd/common/common.go
lukpueh: Ah, silly me, we'd need a `NewSignerFromBytes` here. Let's punt on that.

lukpueh (approved):
LGTM. Just saw that the `NewSignerFromFile` I requested and you added is actually not needed in this PR. I'm okay if you want to keep it or not

internal/signerverifier/signerverifier.go
adityasaky: No, we need another pass to handle this and update a bunch of tests to drop the legacy format. Let's do it then.

Issue #73, opened 2023-08-07
Fix sigstore integration and testing


adityasaky:
#72 introduced gitsign verification support in gittuf using TUF's TAP-18. However, there's work to be done in testing the gitsign support.

1. It might be possible to prune the dependency graph by switching from cosign+gitsign+sigstore/sigstore to the https://github.com/sigstore/sigstore-go once it matures and has the capabilities we need.
2. We can't mock some of the necessary behavior with the current libraries pertaining to looking up the tlog entry for the signature. These are marked `internal` at the moment, the sigstore-go library may also help us out here.

PR #439, opened 2024-06-13
attestations: Add support for GitHub pull request approval attestation

adityasaky:
~This will change a fair bit with #372.~

Stacked over #489


internal/policy/verify.go
```diff
@@ -83,6 +83,186 @@ func VerifyRefFull(ctx context.Context, repo *gitinterface.Repository, target st
...
+		if err == nil {
+			// we've reached threshold with just the approvals
+			gitNamespaceVerified = true
+			break
+		} else if !errors.Is(err, ErrVerifierConditionsUnmet) {
+			// Unexpected error
+			return false, err
+		}
+
+		// If threshold == 1 and we have approval without gitObj signature, we'll have exited above
```
wlynch, line 150: Where is the gitObj signature being done? It's not immediately obvious from this code so this looks like we're checking the same thing twice.

internal/policy/verify.go
```diff
@@ -83,6 +83,186 @@ func VerifyRefFull(ctx context.Context, repo *gitinterface.Repository, target st
...
+		// Only try threshold-- if threshold > 1
+		if verifier.Threshold() > 1 {
+			// Make a copy before mutating verifier
+			verifier := verifier
+			verifier.threshold--
+
+			err := verifier.Verify(ctx, gitinterface.ZeroHash, authorization, githubApproval)
+			if err == nil {
+				rslEntrySignatureNeededForThreshold = true
```
wlynch, line 159: What is this tracking and why is it only checked the 2nd time?

internal/policy/verify.go
```diff
@@ -83,6 +83,186 @@ func VerifyRefFull(ctx context.Context, repo *gitinterface.Repository, target st
...
+	if !gitNamespaceVerified {
+		return false, fmt.Errorf("not enough approvals to meet Git namespace policies, %w", ErrUnauthorizedSignature)
+	}
+
+	hasFileRule, err := policyState.hasFileRule()
+	if err != nil {
+		return false, err
+	}
+	if !hasFileRule {
+		return rslEntrySignatureNeededForThreshold, nil
```
wlynch, line 177: Also not clicking what this has to do with file rules. If this is for mergeability, e.g. "the commit has the number of required RSL entries needed to meet the threshold", then this should be renamed since it reads like "RSL entries are needed to meet the threshold"

internal/policy/verify.go
```diff
@@ -83,6 +83,186 @@ func VerifyRefFull(ctx context.Context, repo *gitinterface.Repository, target st
...
+		return false, err
+	}
+
+	slog.Debug(fmt.Sprintf("Identifying latest RSL entry for '%s'...", featureRef))
+	featureEntry, _, err := rsl.GetLatestUnskippedReferenceEntryForRef(repo, featureRef)
+	if err != nil {
+		return false, err
+	}
+
+	mergeTreeID, err := repo.GetMergeTree(targetEntry.TargetID, featureEntry.TargetID)
```
wlynch, line 111: Why resolve the target ids via the RSL instead of via the refs directly?

Could the RSL potentially be behind the ref value if a new commit was pushed without updating the RSL?

internal/policy/verify.go
```diff
@@ -83,6 +83,186 @@ func VerifyRefFull(ctx context.Context, repo *gitinterface.Repository, target st
...
+			pathNamespaceVerified = false
+			break
+		}
+	}
+
+	if !pathNamespaceVerified {
+		return false, fmt.Errorf("not enough approvals to meet file namespace policies, %w", ErrUnauthorizedSignature)
+	}
+
+	return rslEntrySignatureNeededForThreshold, nil
```
wlynch, line 263: Taking a step back, there's a lot of verification logic here that feels like should be encapsulated elsewhere. Ideally I think this should look like:

What information do you need to be returned from verification in order to make policy decisions about mergability?

internal/policy/verify.go
```diff
@@ -656,35 +669,18 @@ func (v *Verifier) Threshold() int {
 // signature and signatures embedded in a DSSE envelope. Verify does not inspect
 // the envelope's payload, but instead only verifies the signatures. The caller
 // must ensure the validity of the envelope's contents.
-func (v *Verifier) Verify(ctx context.Context, gitObjectID gitinterface.Hash, env *sslibdsse.Envelope) error {
+func (v *Verifier) Verify(ctx context.Context, gitObjectID gitinterface.Hash, authorizationAttestationEnv, githubApprovalAttestationEnv *sslibdsse.Envelope) error {
```
adityasaky, line 672: One big question in my mind is that this interface is going from caring strictly about a git sig and dsse sigs to being aware of a specific predicate type. I think we may want to revert this change and handle the special casing of the github approval attestations separately.

adityasaky:
@wlynch I've moved your comments about verify-mergeable over to #455 so this PR can be focused on the github app approval attestation and related functions.

internal/policy/verify.go
adityasaky: working on rolling this back

adityasaky: i've pushed changes on this, i'll clean up tomorrow.

internal/cmd/trust/githubappapprovals/githubappapprovals.go
```diff
@@ -0,0 +1,67 @@
...
+		"disable",
+		false,
+		"mark GitHub app approvals as untrusted",
+	)
+
+	cmd.MarkFlagsOneRequired("enable", "disable") //nolint:errcheck
+	cmd.MarkFlagsMutuallyExclusive("enable", "disable")
+}
+
+func (o *options) Run(cmd *cobra.Command, _ []string) error {
```
adityasaky, line 37: this one is a deviation from how other CLI commands work. should this be enable and disable subcommands?

internal/tuf/tuf.go
```diff
@@ -82,10 +82,11 @@ type Role struct {
...
 type RootMetadata struct {
-	Type    string          `json:"type"`
-	Expires string          `json:"expires"`
-	Keys    map[string]*Key `json:"keys"`
-	Roles   map[string]Role `json:"roles"`
+	Type                   string          `json:"type"`
+	Expires                string          `json:"expires"`
+	Keys                   map[string]*Key `json:"keys"`
+	Roles                  map[string]Role `json:"roles"`
+	GitHubApprovalsTrusted bool            `json:"githubApprovalsTrusted"`
```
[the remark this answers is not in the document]
adityasaky: I lean towards keeping it vendor specific to start because the attestations we generate may differ. I wonder if the approvers predicate here should be accompanied by a github PR attestation, for example.

adityasaky:
Exploring if we can also track PR numbers here so we can better track dismissals

adityasaky:
This is now tracking reviewID for each approval so that dismissals can be handled. Dismissal tracking isn't perfect but I'd prefer to move ahead with something we can ship in the github app soon and iterate as needed later.

internal/policy/root_test.go
```diff
@@ -133,3 +133,87 @@ func TestDeleteTargetsKey(t *testing.T) {
...
+	rootMetadata, err = AddGitHubAppKey(rootMetadata, appKey)
+	assert.Nil(t, err)
+
+	_, err = DeleteGitHubAppKey(nil)
+	assert.ErrorIs(t, err, ErrRootMetadataNil)
+
+	rootMetadata, err = DeleteGitHubAppKey(rootMetadata)
+	assert.Nil(t, err)
+
+	assert.Equal(t, []string(nil), rootMetadata.Roles[GitHubAppRoleName].KeyIDs)
```
adityasaky, line 184: assert.Empty?

internal/repository/root_test.go
```diff
@@ -253,6 +253,205 @@ func TestRemoveTopLevelTargetsKey(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rootMetadata, err = state.GetRootMetadata()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, []string(nil), rootMetadata.Roles[policy.GitHubAppRoleName].KeyIDs)
```
adityasaky, line 328: assert.Empty

adityasaky: i wonder why we even have an empty slice, we should probably not initialize it if not needed

internal/repository/root_test.go
```diff
@@ -253,6 +253,205 @@ func TestRemoveTopLevelTargetsKey(t *testing.T) {
...
+		sv, err := signerverifier.NewSignerVerifierFromSecureSystemsLibFormat(keyBytes) //nolint:staticcheck
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		err = r.TrustGitHubApp(testCtx, sv, false)
+		assert.Nil(t, err)
+
+		_, err = policy.LoadCurrentState(testCtx, r.r, policy.PolicyStagingRef)
+		assert.NotNil(t, err)
```
adityasaky, line 350: this is actually because of state.Verify right? Can we assert a named error?

adityasaky:
@patzielinski thanks for adding tests!

@wlynch I think this is ready for your review.

The predicate now has an approvers list that we use in the policy package to verify. With the refactor:
* more of the github api logic is moved into gittuf rather than handling it in gittuf/github-app
* the predicate has a dismissedApprovers list for those that dismiss a prior approval (this isn't perfect to establishing a timeline of approve - dismiss - approve, but I think it's okay for now because a dismiss is likely because a new change landed meaning the next approve wouldn't be in the same attestation at all)
* we now track the unique github review ID, which is mapped to the <ref>/<from>-<to>; we need this because if a review is dismissed after changes land in either the base branch or the feature branch, we lose the ability to identify exactly which change the review was for when it was first issued

adityasaky:
gotcha, can we open a ticket for this similar to #73? We can probably mock github api responses but we don't have to do it here I reckon.

internal/repository/root.go
```diff
@@ -191,6 +191,132 @@ func (r *Repository) RemoveTopLevelTargetsKey(ctx context.Context, signer sslibd
 	return r.updateRootMetadata(ctx, state, signer, rootMetadata, commitMessage, signCommit)
 }
 
+// AddGitHubAppKey is the interface for the user to add the authorized key for
+// the special GitHub app role.
+func (r *Repository) AddGitHubAppKey(ctx context.Context, signer sslibdsse.SignerVerifier, appKey *tuf.Key, signCommit bool) error {
```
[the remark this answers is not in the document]
adityasaky: Yeah, it's really more `Set` rather than `Add`.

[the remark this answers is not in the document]
adityasaky: I propose we take a consistency pass for the other APIs as well later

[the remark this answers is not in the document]
adityasaky: Sure

internal/repository/attestations.go
```diff
@@ -255,6 +267,188 @@ func (r *Repository) AddGitHubPullRequestAttestationForNumber(ctx context.Contex
...
+		if err != nil {
+			return err
+		}
+
+		type tmpStatement struct {
+			Type          string                                             `json:"_type"`
+			Subject       []*ita.ResourceDescriptor                          `json:"subject"`
+			PredicateType string                                             `json:"predicateType"`
+			Predicate     *attestations.GitHubPullRequestApprovalAttestation `json:"predicate"`
+		}
```
wlynch, line 400: If you're defining this multiple times, you're probably better off moving this out into it's own type.

wlynch: In general, this looks largely similar to AddGitHubPullRequestApprover - can we refactor this to move common parsing/writing to a shared function?

internal/attestations/attestations.go
```diff
@@ -36,6 +43,24 @@ type Attestations struct {
...
+	githubPullRequestApprovalAttestations map[string]gitinterface.Hash
+
+	// githubPullRequestApprovalIndex is stored in-memory. It maps a GitHub pull
+	// request review ID to the gittuf identifier for a review,
+	// `<ref-path>/<from-id>-<to-id>`. We need this because when a review is
+	// dismissed, we need to unambiguously know what the review applied to when
+	// it was first submitted, which we cannot do with the information at the
+	// time of dismissal. This is serialized to the attestations namespace as a
+	// special blob in the githubPullRequestApprovalAttestations tree.
+	githubPullRequestApprovalIndex map[int64]string
```
wlynch, line 63: I'm a bit skeptical of this - I think we need to include remote information of some kind here if we're going to be keying off the pull request id, otherwise it's ambiguous across repos.

I'm also skeptical of using int64 if we ever want to expand this out to other types of presubmit attestations.

internal/repository/attestations.go
```diff
@@ -255,6 +267,188 @@ func (r *Repository) AddGitHubPullRequestAttestationForNumber(ctx context.Contex
...
+		stmt := &tmpStatement{Subject: []*ita.ResourceDescriptor{}, Predicate: &attestations.GitHubPullRequestApprovalAttestation{Approvers: []*tuf.Key{}, DismissedApprovers: []*tuf.Key{}}}
```
wlynch, line 318: Rule of thumb: only one-line if it's a very simple struct (usually 1 field)
```suggestion
		stmt := &tmpStatement{
		  Subject: []*ita.ResourceDescriptor{}, 
		  Predicate: &attestations.GitHubPullRequestApprovalAttestation{
		    Approvers: []*tuf.Key{},
		    DismissedApprovers: []*tuf.Key{},
		  }
		}
```

wlynch, line 318: Could this be replaced by `stmt := new(tmpStatement)`?

internal/repository/attestations.go
```diff
@@ -308,10 +502,104 @@ func (r *Repository) addGitHubPullRequestAttestation(ctx context.Context, signer
...
+	return baseRef, fromID, toID, nil
+}
+
+func getGitHubClient() (*github.Client, error) {
 	if githubClient == nil {
 		githubClient = github.NewClient(nil).WithAuthToken(os.Getenv("GITHUB_TOKEN"))
 	}
 
-	return githubClient
+	baseURL := os.Getenv("GITHUB_BASE_URL")
```
wlynch, line 569: I would add this in a separate change + add docs.

internal/policy/root.go
```diff
@@ -146,6 +149,58 @@ func DeleteTargetsKey(rootMetadata *tuf.RootMetadata, keyID string) (*tuf.RootMe
 	return rootMetadata, nil
 }
 
+// AddGitHubAppKey adds the 'appKey' as a trusted public key in 'rootMetadata'
+// for the special GitHub app role.
+func AddGitHubAppKey(rootMetadata *tuf.RootMetadata, appKey *tuf.Key) (*tuf.RootMetadata, error) {
```
wlynch, line 154: Is there anything special about this key that's specific to GitHub Apps, or is this a generic "Attestor" role?

wlynch: In particular, I want to avoid confusion with https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/managing-private-keys-for-github-apps

wlynch (commented):
Sorry about the delay here. This is a huge change so it's been hard to find the time to dedicate to sit down and review the whole thing 🙈

internal/policy/root.go
adityasaky: From the gittuf side, this is special. During the verification flow, we trust the contents of the github app predicate (which says X reviewed this change) only if it's signed by this key specified in the root.

FWIW, (I'll double check), I'm currently reusing the key used by the app to communicate with github to also sign the attestation the app generates, so it's currently the same key. We may want to change this, though, if the app uses an identity + sigstore for example.

https://github.com/gittuf/github-app/pull/1/files#diff-59d92a25c9a864d2527aa8c12116ba798ab65673cf86cd0759b4c1075d01653bR299

internal/repository/attestations.go
adityasaky: gotcha, will fix.

internal/attestations/attestations.go
adityasaky: just emphasizing, this isn't the PR number but rather the PR ID. I thought that was unique across repositories, but I can fix this to have more data points. TBF, there's also no easy way to look up a PR strictly by ID I think, so having additional information would be good.

int64 is because I'm using the ID field exactly as returned by go-github, so it matches what's used for the field in the API response. Do we expect presubmit attestations to be concentrated in how they're stored in refs/gittuf/attestations? Do you mean if we're supporting gitlab in addition to github?

internal/repository/attestations.go
adityasaky: :eyes:

internal/repository/attestations.go
adityasaky: https://github.com/gittuf/gittuf/pull/489

adityasaky:
@wlynch: I've opened #489 to split off the enterprise github support bits. I've also addressed a bunch of your other comments, including consolidating how we load the approval attestation predicate across the two APIs. I've held off on bigger changes on https://github.com/gittuf/gittuf/pull/439#discussion_r1681856606 and https://github.com/gittuf/gittuf/pull/439#discussion_r1681853263 so you can confirm:
1) we definitely want to store additional information next to the PR ID to track which repository it is
2) how we want to handle the signing key used by the app

I'll test the current version in a test deployment tomorrow.

>This is a huge change so it's been hard to find the time to dedicate to sit down and review the whole thing 🙈

Yeah, this one got bigger than I thought it would 😓

internal/attestations/attestations.go
adityasaky: Generalized this layer to support multiple code review systems.

internal/policy/root.go
adityasaky: fixed docs for this

adityasaky:
Note: I force pushed https://github.com/gittuf/gittuf/pull/439/commits/b78206675faafedab841774b2b0453e9d68c31ba because I noticed the dismiss-github-approval CLI command didn't have the `review-iD` parameter set

internal/attestations/attestations.go
```diff
@@ -36,6 +45,28 @@ type Attestations struct {
...
+	// codeReviewApprovalAttestations stores the blob ID of a code review
+	// approval attestation generated by or on behalf of a system like GitHub or
+	// Gerrit for the change it applies to. The key is a path of the form
+	// `<ref-path>/<from-id>-<to-id>/<system>`, where `ref-path` is the absolute
+	// ref path such as `refs/heads/main` and `from-id` and `to-id` determine
+	// how the ref in question moved. `<system>` identifies the code review
+	// system in question. For example, the key
+	// `refs/heads/main/<commit-A>-<tree-B>/github` indicates the approved
+	// change updated `refs/heads/main` from `commit-A` to a commit with
+	// `tree-B`, and that the approval originated on GitHub.
```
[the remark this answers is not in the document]
adityasaky: Yes, the predicate types may differ and this supports cases where the same change is approved simultaneously on multiple platforms. I don't know that this makes a lot of sense 99.99% of the time but I wonder if there are strange github+gerrit or so situations out there that we don't want to be backed into a corner on?

internal/attestations/attestations.go
```diff
@@ -12,11 +14,18 @@ import (
...
+
+	referenceAuthorizationsTreeEntryName = "reference-authorizations"
+
 	githubPullRequestAttestationsTreeEntryName = "github-pull-requests"
-	initialCommitMessage                       = "Initial commit"
-	defaultCommitMessage                       = "Update attestations"
+	githubPullRequestApprovalSystemName        = "github"
+
+	codeReviewApprovalAttestationsTreeEntryName = "code-review-approvals"
+	codeReviewApprovalIndexTreeEntryName        = "review-index.json"
```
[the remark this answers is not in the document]
adityasaky: I was trying to keep the reference authorizations (signed by reviewer) and platform reviews (signed by platform / app on platform) separate, but we could consolidate into one tree. Above, we're recording the platform at the end of the code review path. Should we use `gittuf` for the reference authorizations and consolidate? That also addresses the question above about whether we need to record the provider or not, as I can definitely see gittuf auth and forge approvals used together, even if two different forge provider approvals is rare.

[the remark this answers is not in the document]
wlynch: Yeah the situation I'm hoping to avoid is needed to guess which ref is the right place to place attestations.

The question I'd pose is what value do we get from keeping these separate? In many ways if you're using an app for delegate authorizations that IS the authorization, no different than if a human reviewer directly authorized it. If you're not relying on these directly and instead using it for supplemental information, than that seems similar to if someone approves a review but isn't authorized to be authoritative (e.g. if you don't have OWNERS).

I think my expectation is that these would be stored in the same location rather than needing to search N different ref spaces to find all potential authorizations.
That said, I do think it makes sense to have separate refs for things in the repo vs things being proposed to be added to the repo - we may want to take more liberties to limit history depth to keep the repo history clean for things that aren't necessarily trusted.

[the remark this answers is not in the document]
adityasaky: To clarify a couple things, we're using only one ref for attestations, and currently placing user-signed attestations (reference authorizations) and app-signed-on-behalf-of-scp attestations in two directories. The former is stored as a single blob (<ref>/<from>-<to> is the path in the tree, points to a blob object) while the latter is a directory in which each app can have an entry (<ref>/<from>-<to>/<scp>). I'm a little unclear on the "things being proposed to be added" bit here: do you mean attestations for a change that hasn't been merged yet are stored separately from attestations for changes that have been merged in?

wlynch (approved):

Issue #220, opened 2023-12-08
gittuf <-> Git command compatibility


adityasaky:
To improve gittuf's UX (#4), the CLI must provide some command compatibility with Git itself. This would allow users to use gittuf as a drop-in replacement for Git in common workflows such as syncing with remote repositories.

This is complementary to #158, where gittuf installs the right hooks to perform gittuf-specific actions for a repository. There, the user has gittuf installed but does not invoke it directly most of the time.

datosh:
After playing around with `gittuf` for a day now and having a think about the required verbs, I am not convinced someone would actually need `gittuf` to be a drop-in replacement for git. The usual day of a software engineer should only interface with `gittuf` when pushing, i.e. add an rsl record, right? The usual workflow of feature branch -> commit -> push, really only needs the one RSL operation we covered using the git hook. This would even hold true if the repository was freshly cloned, as the hook syncs the RSL before making the change.

On the other hand, all verification (primarily) happens in the CI system...?

Have you thought about creating something like user stories and personas to describe how different folks would use `git` and `gittuf` in a perfect world? We could even codify them as end 2 end tests (as the features become available in `gittuf`), so we can guarantee a feature does not break.

Do we have concrete usecases that are cumbersome today?

JustinCappos:
So in my view, gittuf verification is more likely to happen on the developer machines and eventually at the repository.  CI system, etc. verification is useful, but isn't the primary place most issues are detected.

Alice, who uses gittuf, does a pull and her client notices Bob, who does not use gittuf, has violated policy by pushing directly to master.  Alice's gittuf client remediates this issue and pushes a revised RSL entry.

We're really worried about Alice's workflow here.  An older design had Alice switching between git and gittuf commands as she does her daily operations.  The plan here is to have Alice be able to use / alias to gittuf and just use her normal workflow.  The only different commands she needs to type than Bob is when Alice needs to add a policy or do something gittuf specific.  Then since Bob doesn't even have a command of that type, of course Alice needs to type something Bob doesn't...

adityasaky:
+1, I think CI verification obviously makes a lot of sense, but there's a lot of value IMO in increasing the number of verifiers by having all gittuf users in a repo verify automatically.

naveensrinivasan:
I would like to help with `gittuf pull`.

flandweber:
[...]

PR #440, opened 2024-06-14
Git command compatibility #220

Raghava-Gatadi:
Added Git command to the root command.

### DIFF
--- a/internal/gittuf-git/args/args.go
+++ b/internal/gittuf-git/args/args.go
@@ -1,0 +1,119 @@
+// SPDX-License-Identifier: Apache-2.0
+
+// This file contains modified code from the hub project, available at
+// https://github.com/mislav/hub/blob/master/commands/args.go, and licensed
+// under the MIT License
+
+package args
+
+import (
+	"strings"
+)
+
+const (
+	noopFlag      = "--noop"
+	versionFlag   = "--version"
+	listCmds      = "--list-cmds="
+	helpFlag      = "--help"
+	configFlag    = "-c"
+	chdirFlag     = "-C"
+	flagPrefix    = "-"
+	defaultRemote = "origin"
+)
+
+// Args stores any global flags for the invocation of Git (i.e. those before the
+// operation), the command/operation, and the parameters to said
+// command/operation.
+type Args struct {
+	GlobalFlags []string
+	Command     string
+	Parameters  []string
+	ConfigIdx   int
+	ChdirIdx    int
+}
+
+// ProcessArgs takes in the arguments from the commandline, and then nicely
+// organizes them into an Args struct. This should only be called with 1 or more
+// arguments, excluding the base one (i.e. the executable name)
+func ProcessArgs(args []string) Args {
+	if len(args) < 1 {
+		return Args{
+			GlobalFlags: nil,
+			Command:     "",
+			Parameters:  nil,
+		}
+	}
+
+	// Find the command's index
+	cmdIndex, configIndex, chdirIndex := locateCommand(args)
+
+	// Organize and return args
+	return Args{
+		GlobalFlags: args[0:cmdIndex],
+		Command:     args[cmdIndex],
+		Parameters:  args[cmdIndex+1:],
+		ConfigIdx:   configIndex,
+		ChdirIdx:    chdirIndex,
+	}
+}
+
+// DetermineRemote attempts to find out what the remote is that the user would
+// like to push to. If there is no information to this regard in the original
+// command invocation, "origin" is assumed.
+func DetermineRemote(args []string) string {
+	// TODO: This should be more robust and thourougly tested...
+	if len(args) > 0 && args[0] != "origin" {
+		if len(args) == 1 {
+			return args[0]
+		}
+		for i, arg := range args {
+			if arg == ":" {
+				return args[i+1]
+			}
+		}
+	}
+	return defaultRemote
+}
+
+// locateCommand attempts to find where the main Git operation is in the
+// arguments. We need to do this as flags can be specified before the command.
+// In addition, it returns the location of any specified config directory and
+// change directory operation.
+// (The output format is (commandIndex, configIndex, chdirIndex))
+func locateCommand(args []string) (int, int, int) {
+	skip := false
+	idx, config, chdir := 0, 0, 0
+
+	for i, arg := range args {
+		switch {
+		case skip:
+			// If skip is set, then the current arg is a parameter to the
+			// previous arg. As we've processed it below, increment idx and
+			// proceed to the next arg.
+			idx = i + 1
+			skip = false
+		case arg == versionFlag || arg == helpFlag || strings.HasPrefix(arg, listCmds) || !strings.HasPrefix(arg, flagPrefix):
+			// If the arg is the verion flag, a help / list cmd flag, or does
+			// not have the "-" prefix, we've reached the location of the
+			// operation.
+			return idx, config, chdir
+		default:
+			// Otherwise, we need to keep processing args. Advance the index and
+			// check if the arg is one of interest.
+			idx = i + 1
+			switch arg {
+			case configFlag:
+				// The current arg indicates that the next value will be a
+				// config file. Set the config index appropriately.
+				config = idx
+				skip = true
+			case chdirFlag:
+				// The current arg indicates that the next value will be a
+				// directory. Set the config index appropriately.
+				chdir = idx
+				skip = true
+			}
+		}
+	}
+	return idx, config, chdir
+}
--- a/internal/gittuf-git/args/args_test.go
+++ b/internal/gittuf-git/args/args_test.go
@@ -1,0 +1,133 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package args
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestProcessArgs(t *testing.T) {
+	// TODO: Test this more
+	tests := map[string]struct {
+		args     []string
+		expected Args
+	}{
+		"no arguments": {
+			args: []string{},
+			expected: Args{
+				GlobalFlags: nil,
+				Command:     "",
+				Parameters:  nil,
+			},
+		},
+		"pull": {
+			args: []string{"pull"},
+			expected: Args{
+				GlobalFlags: []string{},
+				Command:     "pull",
+				Parameters:  []string{},
+			},
+		},
+		"push": {
+			args: []string{"push"},
+			expected: Args{
+				GlobalFlags: []string{},
+				Command:     "push",
+				Parameters:  []string{},
+			},
+		},
+		"fetch origin": {
+			args: []string{"fetch", "origin"},
+			expected: Args{
+				GlobalFlags: []string{},
+				Command:     "fetch",
+				Parameters:  []string{"origin"},
+			},
+		},
+		"-C ../somedir fetch origin": {
+			args: []string{"-C", "../somedir", "fetch", "origin"},
+			expected: Args{
+				GlobalFlags: []string{"-C", "../somedir"},
+				Command:     "fetch",
+				Parameters:  []string{"origin"},
+				ChdirIdx:    1,
+			},
+		},
+	}
+
+	for name, test := range tests {
+		got := ProcessArgs(test.args)
+		assert.Equal(t, test.expected, got, fmt.Sprintf("unexpected result in test '%s'", name))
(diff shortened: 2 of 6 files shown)


Raghava-Gatadi:
Removed the previous `gettuf git ...` sub-command, and added a new top-level command `gittuf-git ...`, which redirects to the git binary.

Raghava-Gatadi:
I have moved the git binary from gittuf/main.go to its own main.go file

Raghava-Gatadi:
I've implemented the requested modifications. Please review and verify if any further adjustments are needed.

internal/gittuf-git/git.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"errors"
+	"fmt"
+	"os/exec"
+
+	"github.com/spf13/cobra"
+)
+
+// gitCmd represents the base command when called without any subcommands
+var gitCmd = &cobra.Command{
```
adityasaky, line 14: How much control do we have over what flags package ends up being used by cobra? Do we have to resort to a go mod replace directive? I ask because cobra can be a bit tetchy about unrecognized args, and I'm not yet convinced cobra.ArbitraryArgs is ok. Do we want to just have a very lightweight wrapper around the stdlib flags library? Do we process flags ourselves?


I don't have answers to these questions yet but I'd like the both of you to think / explore a bit. :smile:

internal/gittuf-git/push.go
```diff
@@ -0,0 +1,32 @@
...
+	return rslRecordArgs
+}
+
+func gittufPush(gitArgs []string) {
+	rslRecordArgs := splitPushArgs(gitArgs[1:])
+	recordCmd := rsl.New()
+	rslArg := []string{}
+	for _, arg := range rslRecordArgs {
+		rslArg = append(rslArg, arg)
+		if err := recordCmd.RunE(nil, rslArg); err != nil {
```
neilnaveen, line 28: Now that https://github.com/gittuf/gittuf/pull/369 has been merged, we can run `gittuf rsl skip-rewritten` when recording in the RSL if the ref's state has changed, i.e., rebasing.

Raghava-Gatadi:
I've made changes to the /gittuf-git/push.go file to incorporate a rebase check. Please review the update and confirm if it functions as expected.

neilnaveen (commented):
This looks a lot better, a couple of comments.

internal/gittuf-git/push.go
```diff
@@ -0,0 +1,55 @@
...
+	rslRecordArgs := splitPushArgs(gitArgs[1:])
+	recordCmd := record.New()
+	rslArg := []string{}
+	for _, arg := range rslRecordArgs {
+		rslArg = append(rslArg, arg)
+		hasRebased, err := hasRefBeenRebased(arg)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to check for rebase: %v", err)
+		}
+		if hasRebased {
```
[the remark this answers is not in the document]
neilnaveen: both must be called, `gittuf skip-rewritten` just creates an annotation that skips those RSL entries, and `gittuf rsl record` actualy records a new RSL entry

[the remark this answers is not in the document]
Raghava-Gatadi: Check it once, I made changes to file to call both the functions

internal/cmd/root/root.go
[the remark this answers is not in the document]
neilnaveen: This hasn't been plugged into the cmd. Please check it out.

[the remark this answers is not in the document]
neilnaveen: Oh, I had assumed that gittuf-git would be a subcommand, and you would call it with `gittuf git ...`

Raghava-Gatadi:
If everything is completed, please check if `gittuf-git push` is functioning correctly as expected.

neilnaveen:
This looks good to me, but could you add some tests for push.go, maybe somthing like clone's tests https://github.com/gittuf/gittuf/blob/main/internal/repository/sync_test.go? 👀 @patzielinski @adityasaky

Raghava-Gatadi:
I've been assigned the task of identifying suitable libraries for extracting arguments from command-line input. Once I've found the most appropriate library and got approved to implement, I'll add test's for the push.go

Raghava-Gatadi:
while transferring the `gittuf-git push ..` command args to rslRecord command, should we also pass "origin" to rslRecord command? I guess rslRecord only requires 1 argument, i.e branch-name/main.

internal/gittuf-git/args/args.go
```diff
@@ -0,0 +1,123 @@
...
+
+	return returnArgs
+}
+
+// locateCommand attempts to find where the main Git operation is in the
+// arguments. We need to do this as flags can be specified before the command.
+// In addition, it returns the location of any specified config directory and
+// change directory operation.
+// (The output format is (commandIndex, configIndex, chdirIndex))
+func locateCommand(args []string) (int, int, int) {
```
adityasaky, line 70: This function is rather difficult to understand. Could you clarify what `skip` is, when we encounter it, etc.?

internal/gittuf-git/args/args.go
```diff
@@ -0,0 +1,123 @@
...
+	skip, isConfig, isChdir := false, false, false
+	idx, config, chdir := 0, 0, 0
+
+	for i, arg := range args { // Iterate through all args
+		switch {
+		case skip: // Is this value special?
+			switch {
+			case isConfig: // Config value?
+				config = idx
+				idx = i + 1
```
adityasaky, line 80: This can be moved out of the switch and a default isn't needed.

internal/gittuf-git/pull/pull.go
```diff
@@ -0,0 +1,57 @@
...
+package pull
+
+import (
+	"os"
+	"os/exec"
+
+	"github.com/gittuf/gittuf/internal/gittuf-git/args"
+)
+
+func GittufPull(gitArgs args.Args) error {
```
adityasaky, line 12: Just `Pull` is probably fine.

Is there a reason for this to be a separate package?

internal/gittuf-git/main.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// 2. Categorize the Git operation.
+	switch gitArgs.Command {
+	case "push": // 3a
+		err := push.GittufPush(gitArgs)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%s\n", err.Error())
+			os.Exit(1)
+		}
+	case "pull": // 3a
```
adityasaky, line 59: This should also handle `git fetch`

internal/gittuf-git/pull/pull.go
```diff
@@ -0,0 +1,57 @@
...
+	gitPullCmd.Stdout = os.Stdout
+	gitPullCmd.Stderr = os.Stderr
+
+	if err := gitPullCmd.Run(); err != nil {
+		return err
+	}
+
+	// 2
+	remote := args.DetermineRemote(gitArgs.Parameters)
+	rslPullCmd := exec.Command("gittuf", "rsl", "remote", "pull", remote)
```
adityasaky, line 38: the gittuf operations can be invoked using the API rather than a subprocess.

internal/gittuf-git/pull/pull.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// 0
+	if gitArgs.ChdirIdx > 0 {
+		env = append(env, "GIT_DIR="+gitArgs.GlobalFlags[gitArgs.ChdirIdx])
+	}
+
+	// 1
+	cmdArgs := []string{gitArgs.Command}
+	cmdArgs = append(cmdArgs, gitArgs.Parameters...)
+	gitPullCmd := exec.Command("git", cmdArgs...)
```
adityasaky, line 28: What happens if there's a symlink `~/.bin/git -> /usr/bin/gittuf-git`? We may not be able to guarantee this will use the correct `git` all the time, but perhaps we can do better. For now, a TODO is probably ok.

internal/gittuf-git/pull/pull.go
```diff
@@ -0,0 +1,57 @@
...
+	// 0. Determine appropriate env vars
+	// 1. Pull non-RSL changes
+	// 2. Pull RSL changes
+	// 3. Pull policy changes
+
+	var env = os.Environ()
+
+	// 0
+	if gitArgs.ChdirIdx > 0 {
+		env = append(env, "GIT_DIR="+gitArgs.GlobalFlags[gitArgs.ChdirIdx])
```
adityasaky, line 22: This may or may not be the git dir in question. The git dir may be that dir + `.git` etc. I'd also check if every git operation can work with just the git dir change (every gittuf specific git operation can be). Off the top of my head, this may not be true for `git mv` etc, which is probably why both git-dir and -C exist? A chdir may be unavoidable if so.

internal/gittuf-git/push/push.go
```diff
@@ -0,0 +1,58 @@
...
+
+	var env = os.Environ()
+
+	// 0
+	if gitArgs.ChdirIdx > 0 {
+		env = append(env, "GIT_DIR="+gitArgs.GlobalFlags[gitArgs.ChdirIdx])
+	}
+
+	// 1
+	rslRecordCmd := exec.Command("gittuf", "rsl", "record")
```
adityasaky, line 27: Is this recording HEAD? The push may be for a completely different branch than what's checked out. Unfortunately, you'd have to also evaluate what refs are being pushed (and also process refspecs if that's passed in).

adityasaky (commented):
This seems to be off to a pretty good start!

internal/gittuf-git/args/args.go
```diff
@@ -0,0 +1,124 @@
...
+
+import (
+	"strings"
+)
+
+/*
+	Argument processing for Git is a bit tricky, but we have the advantage of
+	only needing to support a few operations.
+
+	Significant inspiration drawn from
```
adityasaky, line 13: We may need a copyright notice for the hub authors here.

internal/gittuf-git/args/args.go
```diff
@@ -0,0 +1,124 @@
...
+	ChdirIdx    int
+}
+
+// ProcessArgs takes in the arguments from the commandline, and then nicely
+// organizes them into an Args struct. This should only be called with 1 or more
+// arguments, excluding the base one (i.e. the executable name)
+func ProcessArgs(args []string) Args {
+	if len(args) < 1 {
+		return Args{
+			GlobalFlags: []string{},
```
adityasaky, line 44:
```suggestion
			GlobalFlags: nil,
```

?

internal/gittuf-git/args/args.go
```diff
@@ -0,0 +1,124 @@
...
+			Command:     "",
+			Parameters:  []string{},
+		}
+	}
+
+	// Find the command's index
+	cmdIndex, configIndex, chdirIndex := locateCommand(args)
+
+	// Organize and return args
+	returnArgs := Args{
```
adityasaky, line 54:
```suggestion
	return Args{
```

internal/gittuf-git/args/args.go
```diff
@@ -0,0 +1,124 @@
...
+				// directory. Set the config index appropriately.
+				chdir = idx
+				skip = true
+			}
+		}
+	}
+	return idx, config, chdir
+}
+
+// determineRemote attempts to find out what the remote is that the user would
```
adityasaky, line 108:
```suggestion
// DetermineRemote attempts to find out what the remote is that the user would
```

internal/gittuf-git/args/args.go
```diff
@@ -0,0 +1,124 @@
...
+// like to push to. If there is no information to this regard in the original
```
adityasaky, line 109: fetch?

internal/gittuf-git/args/args.go
```diff
@@ -0,0 +1,124 @@
...
+		if len(args) == 1 {
+			return args[0]
+		}
+		for i, arg := range args {
+			if arg == ":" {
+				return args[i+1]
+			}
+		}
+	}
+	return "origin"
```
adityasaky, line 123: make this a const at the top?

internal/gittuf-git/args/args_test.go
```diff
@@ -0,0 +1,113 @@
...
+package args
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TODO: Test this more?
```
adityasaky, line 12: I'd move this into the test func

adityasaky (approved):

PR #442, opened 2024-06-14
ci: Enable `get-started` tester on certain docs changes

patzielinski:
We should run the `get-started` CI tests if there are changes to the `get-started.md` doc or testing data.

### DIFF
--- a/.github/workflows/get-started-tests.yml
+++ b/.github/workflows/get-started-tests.yml
@@ -4,9 +4,13 @@ on:
     branches: ['main']
     paths-ignore:
       - "docs/*"
+      - "!docs/testing/*"
+      - "!docs/get-started.md"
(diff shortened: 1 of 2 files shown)


adityasaky:
for the build gittuf step, could you also update the Makefile so that we have:

```
install : test just-install

just-install:
	CGO_ENABLED=0 go install -trimpath -ldflags "$(LDFLAGS)" github.com/gittuf/gittuf
```

And in the CI job, instead of `make`, let's run `make just-install`. I see no reason to run tests twice in CI effectively.

adityasaky:
See: https://github.com/gittuf/gittuf/actions/runs/9520506484/job/26246067112#step:5:12

adityasaky (approved):

Issue #384, opened 2024-05-07
`verify-commit` semantic refinement


flandweber:
This issue is intended to discuss a possible change in semantics for the `verify-commit` subcommand.

I'm still a little unclear on the specifics, but from what I understand from talking with @adityasaky it [gathers the first reference to a commit](https://github.com/gittuf/gittuf/blob/1a0878a25211a6fbaf553d065de27f847462fc71/internal/policy/policy.go#L150) ("first mention is RSL") and verifies if that was a valid change of the reference.

@adityasaky now proposed a change the semantics towards verifying a commit _with respect to a given reference_ (e. g. branch).

Example:

state 0:
```
180757b (feature) feature commit
5fa7115 (HEAD -> main) init
```

state 1 (main rebase to feature):
```
180757b (HEAD -> main, feature) feature commit
5fa7115 init
```

If my understanding of the current implementation is correct, if `gittuf verify-commit 180757b` was run after the merge (state 1), it would still verify the commiter of 180757b was allowed to commit to feature branch according to the policy at state 0.

The new implementation would require a reference as a second argument.
The command `gittuf verify-commit 180757b feature` could then be used to check if main was valid at state 0 whereas  `gittuf verify-commit 180757b main` would check if the original commit was according to policy.

Are reverts valid gittuf actions i. e. recorded to the rsl just the same?
If so, a commit could be added to the rsl, reverted and then readded, which means the combination commit/reference does not uniquely identify a rsl state.
How should gittuf behave in such situations?
Are there significant drawbacks to simply picking the first/last occurrence?

flandweber:
Come to think of it, maybe an added flag for `verify-ref` would be a better match for the proposed semantics:

`gittuf verify-ref main --at 180757b`

Which would have the additional perk of being able to throw out `verify-commit` and `verofy-tag` entirely and renaming `verify-ref` to `verify`.

neilnaveen:
We would not necessarily be able to throw out verify-tag, since while each tag is a ref itself, tags, and their verification work diffrently compared to regular refs. We maybe could route tag verification through gittuf verify-ref, but I think it would have to be its own feature under the hood.

neilnaveen:
[...]

PR #445, opened 2024-06-17
*: Drop verify-{commit,tag} workflows

adityasaky:
These workflows have some shortcomings, as discussed in https://github.com/gittuf/gittuf/issues/384#issuecomment-2099153472.

verify-commit doesn't currently do enough to ensure the right policy is identified for when a commit is first introduced. verify-ref, with some enhancements, is better.

verify-tag implements a subset of verify-ref already because it recognizes tags are refs. Thus, verify-ref is again a better option.

Note: this makes #431 a bit cleaner.


adityasaky:
cc @patzielinski @flandweber @neilnaveen

neilnaveen:
This makes much more sense since the verify commit cannot assure us that the ref is correctly verified, and the tags themselves are just special refs. We would have to add some code into verify-ref to handle tags for everything to work smoothly, or would we only want the basic verification we get from verify-ref? Thoughts?

flandweber:
I agree. I believe the most important benefit of this change is that it makes gittuf simpler to use.

adityasaky:
> This makes much more sense since the verify commit cannot assure us that the ref is correctly verified, and the tags themselves are just special refs. We would have to add some code into verify-ref to handle tags for everything to work smoothly, or would we only want the basic verification we get from verify-ref? Thoughts?

We already handle tags separately via verifyTagEntry, so I think we're covered actually. I also have a change (ab1fa83bc406a4394217f0047f699e6292f3a933) coming that updates verifyTagEntry to use the `verifier` bits to enable attestations.

wlynch (approved):

PR #451, opened 2024-06-19
Add functions and a command for reorder rules in a policy file

fr0m-scratch:
1. Add a function ReorderDelegations() to handle rule reordering from backend
2. Add a command ```gittuf policy reorder-rules [flags]``` to reorder rules from CLI

### DIFF
--- a/internal/cmd/policy/policy.go
+++ b/internal/cmd/policy/policy.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gittuf/gittuf/internal/cmd/policy/listrules"
 	"github.com/gittuf/gittuf/internal/cmd/policy/persistent"
 	"github.com/gittuf/gittuf/internal/cmd/policy/removerule"
+	"github.com/gittuf/gittuf/internal/cmd/policy/reorderrules"
 	"github.com/gittuf/gittuf/internal/cmd/policy/sign"
 	"github.com/gittuf/gittuf/internal/cmd/policy/updaterule"
 	"github.com/gittuf/gittuf/internal/cmd/trustpolicy/apply"
@@ -32,6 +33,7 @@ func New() *cobra.Command {
 	cmd.AddCommand(listrules.New())
 	cmd.AddCommand(remote.New())
 	cmd.AddCommand(removerule.New(o))
+	cmd.AddCommand(reorderrules.New(o))
 	cmd.AddCommand(sign.New(o))
 	cmd.AddCommand(updaterule.New(o))
 
--- a/internal/cmd/policy/reorderrules/reorderrules.go
+++ b/internal/cmd/policy/reorderrules/reorderrules.go
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package reorderrules
+
+import (
+	"github.com/gittuf/gittuf/internal/cmd/common"
+	"github.com/gittuf/gittuf/internal/cmd/policy/persistent"
+	"github.com/gittuf/gittuf/internal/repository"
+	"github.com/spf13/cobra"
+)
+
+type options struct {
+	p          *persistent.Options
+	policyName string
+	ruleNames  []string
+}
+
+func (o *options) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(
+		&o.policyName,
+		"policy-name",
+		"targets",
+		"name of policy file to reorder rules in",
+	)
+}
+
+func (o *options) Run(cmd *cobra.Command, args []string) error {
+	o.ruleNames = append(o.ruleNames, args...)
+
+	repo, err := repository.LoadRepository()
+	if err != nil {
+		return err
+	}
+
+	signer, err := common.LoadSigner(o.p.SigningKey)
+	if err != nil {
+		return err
+	}
+
+	err = repo.ReorderDelegations(cmd.Context(), signer, o.policyName, o.ruleNames, true)
+	return err
+}
+
+func New(persistent *persistent.Options) *cobra.Command {
+	o := &options{p: persistent}
+	cmd := &cobra.Command{
+		Use:               "reorder-rules",
+		Short:             "Reorder rules in the specified policy file",
+		Long:              "This command allows users to reorder rules in the specified policy file. By default, the main policy file is selected. The rule names need to be passed as arguments, in the new order they must appear in, starting from the first to the last rule. Rule names may contain spaces, so they should be enclosed in quotes if necessary.",
+		PreRunE:           common.CheckIfSigningViableWithFlag,
+		RunE:              o.Run,
+		DisableAutoGenTag: true,
+	}
+
+	o.AddFlags(cmd)
+
+	return cmd
+}
--- a/internal/policy/targets.go
+++ b/internal/policy/targets.go
@@ -4,14 +4,21 @@ package policy
 
 import (
 	"errors"
+	"fmt"
+	"strings"
 	"time"
 
+	"github.com/gittuf/gittuf/internal/common/set"
 	"github.com/gittuf/gittuf/internal/tuf"
 )
 
 const AllowRuleName = "gittuf-allow-rule"
 
-var ErrCannotManipulateAllowRule = errors.New("cannot change in-built gittuf-allow-rule")
+var (
+	ErrCannotManipulateAllowRule = errors.New("cannot change in-built gittuf-allow-rule")
+	ErrRuleNotFound              = errors.New("cannot find rule entry")
+	ErrMissingRules              = errors.New("some rules are missing")
+)
 
 // InitializeTargetsMetadata creates a new instance of TargetsMetadata.
 func InitializeTargetsMetadata() *tuf.TargetsMetadata {
@@ -92,6 +99,58 @@ func UpdateDelegation(targetsMetadata *tuf.TargetsMetadata, ruleName string, aut
 	allDelegations = append(allDelegations, AllowRule())
 
 	targetsMetadata.Delegations.Roles = allDelegations
+
+	return targetsMetadata, nil
+}
+
+// ReorderDelegations changes the order of delegations, and the new order is
+// specified in `ruleNames []string`.
+func ReorderDelegations(targetsMetadata *tuf.TargetsMetadata, ruleNames []string) (*tuf.TargetsMetadata, error) {
+	// Create a map of all existing delegations for quick look up
+	rolesMap := make(map[string]tuf.Delegation)
+	// Create a set of current rules in metadata, skipping the allow rule
+	currentRules := set.NewSet[string]()
+	for _, delegation := range targetsMetadata.Delegations.Roles {
+		if delegation.Name == AllowRuleName {
+			continue
+		}
+		rolesMap[delegation.Name] = delegation
+		currentRules.Add(delegation.Name)
+	}
+
+	specifiedRules := set.NewSet[string]()
+	for _, name := range ruleNames {
+		if specifiedRules.Has(name) {
+			return nil, fmt.Errorf("%w: '%s'", ErrDuplicatedRuleName, name)
+		}
+		specifiedRules.Add(name)
+	}
+
+	if !currentRules.Equal(specifiedRules) {
+		onlyInSpecifiedRules := specifiedRules.Minus(currentRules)
+		if onlyInSpecifiedRules.Len() != 0 {
+			if onlyInSpecifiedRules.Has(AllowRuleName) {
+				return nil, fmt.Errorf("%w: do not specify allow rule", ErrCannotManipulateAllowRule)
+			}
+
+			contents := onlyInSpecifiedRules.Contents()
+			return nil, fmt.Errorf("%w: rules '%s' do not exist in current rule file", ErrRuleNotFound, strings.Join(contents, ", "))
+		}
+
+		onlyInCurrentRules := currentRules.Minus(specifiedRules)
+		if onlyInCurrentRules.Len() != 0 {
+			contents := onlyInCurrentRules.Contents()
+			return nil, fmt.Errorf("%w: rules '%s' not specified", ErrMissingRules, strings.Join(contents, ", "))
+		}
+	}
+
+	// Create newDelegations and set it in the targetsMetadata after adding allow rule
+	newDelegations := make([]tuf.Delegation, 0, len(rolesMap)+1)
+	for _, ruleName := range ruleNames {
+		newDelegations = append(newDelegations, rolesMap[ruleName])
+	}
+	newDelegations = append(newDelegations, AllowRule())
+	targetsMetadata.Delegations.Roles = newDelegations
 
 	return targetsMetadata, nil
 }
--- a/internal/policy/targets_test.go
+++ b/internal/policy/targets_test.go
@@ -3,6 +3,7 @@
 package policy
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/gittuf/gittuf/internal/signerverifier"
@@ -84,6 +85,83 @@ func TestUpdateDelegation(t *testing.T) {
 		Terminating: false,
 		Role:        tuf.Role{KeyIDs: []string{key1.KeyID, key2.KeyID}, Threshold: 1},
 	}, targetsMetadata.Delegations.Roles[0])
+}
+
+func TestReorderDelegations(t *testing.T) {
+	targetsMetadata := InitializeTargetsMetadata()
+
+	key1, err := tuf.LoadKeyFromBytes(targets1PubKeyBytes)
+	if err != nil {
+		t.Fatal(err)
+	}
+	key2, err := tuf.LoadKeyFromBytes(targets2PubKeyBytes)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	targetsMetadata, err = AddDelegation(targetsMetadata, "rule-1", []*tuf.Key{key1}, []string{"path1/"}, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	targetsMetadata, err = AddDelegation(targetsMetadata, "rule-2", []*tuf.Key{key2}, []string{"path2/"}, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	targetsMetadata, err = AddDelegation(targetsMetadata, "rule-3", []*tuf.Key{key1, key2}, []string{"path3/"}, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	tests := map[string]struct {
+		ruleNames     []string
+		expected      []string
+		expectedError error
+	}{
+		"reverse order (valid input)": {
+			ruleNames:     []string{"rule-3", "rule-2", "rule-1"},
+			expected:      []string{"rule-3", "rule-2", "rule-1", AllowRuleName},
+			expectedError: nil,
+		},
+		"rule not specified in new order": {
+			ruleNames:     []string{"rule-3", "rule-2"},
+			expectedError: ErrMissingRules,
+		},
+		"rule repeated in the new order": {
+			ruleNames:     []string{"rule-3", "rule-2", "rule-1", "rule-3"},
+			expectedError: ErrDuplicatedRuleName,
+		},
+		"unknown rule in the new order": {
+			ruleNames:     []string{"rule-3", "rule-2", "rule-1", "rule-4"},
+			expectedError: ErrRuleNotFound,
+		},
+		"unknown rule in the new order (with correct length)": {
+			ruleNames:     []string{"rule-3", "rule-2", "rule-4"},
+			expectedError: ErrRuleNotFound,
+		},
+		"allow rule appears in the new order": {
+			ruleNames:     []string{"rule-2", "rule-3", "rule-1", AllowRuleName},
+			expectedError: ErrCannotManipulateAllowRule,
+		},
+	}
+
+	for name, test := range tests {
+		_, err = ReorderDelegations(targetsMetadata, test.ruleNames)
+		if test.expectedError != nil {
+			assert.ErrorIs(t, err, test.expectedError, fmt.Sprintf("unexpected error in test '%s'", name))
+		} else {
+			assert.Nil(t, err, fmt.Sprintf("unexpected error in test '%s'", name))
+			assert.Equal(t, len(test.expected), len(targetsMetadata.Delegations.Roles),
+				fmt.Sprintf("expected %d rules in test '%s', but got %d rules",
+					len(test.expected), name, len(targetsMetadata.Delegations.Roles)))
+			for i, ruleName := range test.expected {
+				assert.Equal(t, ruleName, targetsMetadata.Delegations.Roles[i].Name,
+					fmt.Sprintf("expected rule '%s' at index %d in test '%s', but got '%s'",
+						ruleName, i, name, targetsMetadata.Delegations.Roles[i].Name))
+			}
+		}
+	}
 }
 
 func TestRemoveDelegation(t *testing.T) {
--- a/internal/repository/targets.go
+++ b/internal/repository/targets.go
@@ -200,6 +200,59 @@ func (r *Repository) UpdateDelegation(ctx context.Context, signer sslibdsse.Sign
 	return state.Commit(r.r, commitMessage, signCommit)
 }
 
+// ReorderDelegations is the interface for the user to reorder rules in gittuf
+// policy.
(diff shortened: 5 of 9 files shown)


fr0m-scratch:
> This looks good! A couple of items:
> 
> * Each commit to be merged into gittuf needs a DCO signoff, which can be achieved with the `-s` flag when running `git commit` [(see here)](https://wiki.linuxfoundation.org/dco) . You'll need to amend the commit and then force-push.
> * `ReorderDelegations` in `repository/targets.go` and `policy/targets.go` will need tests. See [here](https://github.com/gittuf/gittuf/blob/main/internal/repository/targets_test.go#L132) and [here](https://github.com/gittuf/gittuf/blob/main/internal/policy/targets_test.go#L48) for examples. You can copy part of the existing tests and just add code that first reorders and then checks that reordering was successful.
> * I'm unsure on the ordering format at the moment (i.e. comma separated or otherwise), I need to look a bit more into that.

Thank you for your comments! I've made the requested change and tested out. Please help me check it : )

internal/policy/targets.go
```diff
@@ -96,6 +96,27 @@ func UpdateDelegation(targetsMetadata *tuf.TargetsMetadata, ruleName string, aut
...
+	for _, ruleName := range ruleNames {
+		if ruleName == AllowRuleName {
+			return nil, ErrCannotManipulateAllowRule
+		}
+
+		for _, delegation := range targetsMetadata.Delegations.Roles {
+			if delegation.Name == ruleName {
+				allDelegations = append(allDelegations, delegation)
+			}
+		}
```
neilnaveen, line 111: Instead of doing this for every single rulename, can we first precompute this in a map of rulename to roles? Then, when we iterate through the rule names, we can query the roles for that rule name and append it to allDelegations instead of iterating multiple times.

internal/cmd/policy/reorderrules/reorderrules.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/gittuf/gittuf/internal/cmd/common"
+	"github.com/gittuf/gittuf/internal/cmd/policy/persistent"
+	"github.com/gittuf/gittuf/internal/repository"
+	"github.com/spf13/cobra"
+)
+
+type options struct {
+	p          *persistent.Options
+	policyName string
+	ruleNames  string
```
neilnaveen, line 17: Instead of using a string, we could use cobra's StringArray, to take in an array of string, which means if the user mistakenly adds whitespace in front of a rulename, the reorder will still work

neilnaveen (commented):
This looks good, a couple of suggestions.

internal/cmd/policy/reorderrules/reorderrules.go
fr0m-scratch: Thanks, great idea!

internal/policy/targets_test.go
```diff
@@ -86,6 +86,39 @@ func TestUpdateDelegation(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	targetsMetadata, err = AddDelegation(targetsMetadata, "rule-2", []*tuf.Key{key2}, []string{"path2/"}, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Contains(t, targetsMetadata.Delegations.Keys, key1.KeyID)
+	assert.Contains(t, targetsMetadata.Delegations.Keys, key2.KeyID)
```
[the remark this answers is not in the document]
fr0m-scratch: Thanks, copy that

internal/cmd/policy/reorderrules/reorderrules.go
```diff
@@ -0,0 +1,69 @@
...
+
+	return nil
+}
+
+func New(persistent *persistent.Options) *cobra.Command {
+	o := &options{p: persistent}
+	cmd := &cobra.Command{
+		Use:   "reorder-rules",
+		Short: "Reorder rules in a policy file",
+	}
```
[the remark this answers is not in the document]
fr0m-scratch: Thanks, gocha

internal/policy/targets.go
fr0m-scratch: Thanks, I'll try that with a pre-computed map

fr0m-scratch:
> @fr0m-scratch Can you squash commits?

Sure, copy that

adityasaky:
I lean towards error-ing out. Too easy to forget a rule in the CLI. That said, can you record this flow using asciinema or something to get an idea of how this feature is being implemented? thanks!

fr0m-scratch:
I added checks for missing rules as well as relevant tests and squashed to one commit. Below is the link for my asciinema demo recording: https://asciinema.org/a/LnFiudjQD1RmgsnYb1Uy5b0TH

fr0m-scratch:
> Another edge case which we will want to catch: duplicate rule names with the correct number of rules total. For example, if the policy has `r1, r2, r3` as rules, calling `reorder-rules --rule-order r1 r2 r2` will work.
> 
> There's a `set` package that should be useful for this PR: https://github.com/gittuf/gittuf/blob/main/internal/common/set/set.go
> 
> Let's try and see how we can use that to simplify some operations in the PR (e.g. get the `Contents` of one set and loop over them, calling `Has` on the other set)

Appreciate for helping find the edge cases! Resolved with the suggested approach.
Screenshot 2024-08-02 at 2 56 48 AM

internal/repository/targets_test.go
```diff
@@ -167,6 +167,74 @@ func TestUpdateDelegation(t *testing.T) {
...
+		finalOrder := []string{}
+		for _, role := range targetsMetadata.Delegations.Roles {
+			finalOrder = append(finalOrder, role.Name)
+		}
+		expectedFinalOrder := append([]string{}, newOrder...)
+		expectedFinalOrder = append(expectedFinalOrder, "gittuf-allow-rule")
+		assert.Equal(t, expectedFinalOrder, finalOrder)
+	})
+
+	t.Run("invalid rule names", func(t *testing.T) {
```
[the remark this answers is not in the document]
fr0m-scratch: Copy that! I've pushed a new commit; please let me know If assert.ErrorIs() is good enough or I should make the asserts more specific.

internal/cmd/policy/policy.go
```diff
@@ -34,6 +35,7 @@ func New() *cobra.Command {
 	cmd.AddCommand(removerule.New(o))
 	cmd.AddCommand(sign.New(o))
 	cmd.AddCommand(updaterule.New(o))
+	cmd.AddCommand(reorderrules.New(o))
```
adityasaky, line 38: nit: let's order this alphabetically so this goes after removerule.

internal/cmd/policy/reorderrules/reorderrules.go
```diff
@@ -0,0 +1,70 @@
...
+	cmd.Flags().StringVar(
+		&o.policyName,
+		"policy-name",
+		"targets",
+		"name of policy file to reorder rules in",
+	)
+
+	cmd.Flags().StringSliceVar(
+		&o.ruleNames,
+		"rule-order",
```
adityasaky, line 28:
```suggestion
		"rule-name",
```

internal/cmd/policy/reorderrules/reorderrules.go
```diff
@@ -0,0 +1,70 @@
...
+		[]string{},
+		"a space-separated list of rule names, in the new order that they should appear in, from first to last",
```
adityasaky, line 30: does the flag not need to be repeated? as in `--rule-name <> --rule-name <> ...`

internal/cmd/policy/reorderrules/reorderrules.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return err
+	}
+
+	signer, err := common.LoadSigner(o.p.SigningKey)
+	if err != nil {
+		return err
+	}
+
+	err = repo.ReorderDelegations(cmd.Context(), signer, o.policyName, o.ruleNames, true)
```
adityasaky, line 48: just return here?

internal/cmd/policy/reorderrules/reorderrules.go
```diff
@@ -0,0 +1,70 @@
...
+
+	return nil
+}
+
+func New(persistent *persistent.Options) *cobra.Command {
+	o := &options{p: persistent}
+	cmd := &cobra.Command{
+		Use:               "reorder-rules",
+		Short:             "Reorder rules in the specified policy file",
+		Long:              `This command allows users to reorder rules in the specified policy file. Specify the names of rules in the new order they should appear in, starting from first to last rule. By default, the main policy file is selected. Note that authorized keys can be specified from disk, from the GPG keyring using the "gpg:<fingerprint>" format, or as a Sigstore identity as "fulcio:<identity>::<issuer>".`,
```
adityasaky, line 61: nit: I'd write this more like "The rule names must be specified in the new order they must appear in...".

We also don't have authorized keys here, no?

internal/policy/targets.go
```diff
@@ -96,6 +102,54 @@ func UpdateDelegation(targetsMetadata *tuf.TargetsMetadata, ruleName string, aut
...
+	// Create a map of all delegations for quick validation.
+	rolesMap := make(map[string]tuf.Delegation)
+	for _, delegation := range targetsMetadata.Delegations.Roles {
+		rolesMap[delegation.Name] = delegation
+	}
+
+	ruleSet := set.NewSet[string]()
+
+	for _, ruleName := range ruleNames {
+		if ruleName == AllowRuleName {
```
adityasaky, line 117: this brings up a good question: should users be specifying the allow-rule? IMO it's a convenience for us, so they should specify everything but.

internal/policy/targets.go
```diff
@@ -96,6 +102,54 @@ func UpdateDelegation(targetsMetadata *tuf.TargetsMetadata, ruleName string, aut
...
+			return nil, ErrCannotManipulateAllowRule
+		}
+
+		if ruleSet.Has(ruleName) {
+			return nil, fmt.Errorf("%w: %s", ErrDuplicatedRuleName, ruleName)
+		}
+
+		if delegation, valid := rolesMap[ruleName]; valid {
```
adityasaky, line 125: s/valid/exists -> we're not checking validity but that we know the delegation exists.

internal/policy/targets.go
```diff
@@ -96,6 +102,54 @@ func UpdateDelegation(targetsMetadata *tuf.TargetsMetadata, ruleName string, aut
...
+	// Check for any missing delegations
+	missingRules := []string{}
+	for ruleName := range rolesMap {
+		if ruleName != AllowRuleName {
+			missingRules = append(missingRules, ruleName)
+		}
+	}
+
+	if len(missingRules) > 0 {
+		return nil, fmt.Errorf("%w: %s", ErrMissingRules, missingRules)
```
adityasaky, line 145: I'd strings.Join the missingRules with a `, `, a comma and space basically.

internal/policy/targets.go
```diff
@@ -96,6 +102,54 @@ func UpdateDelegation(targetsMetadata *tuf.TargetsMetadata, ruleName string, aut
 	return targetsMetadata, nil
 }
 
+// ReorderDelegations changes the order of delegations in TargetsMetadata.
+func ReorderDelegations(targetsMetadata *tuf.TargetsMetadata, ruleNames []string) (*tuf.TargetsMetadata, error) {
```
adityasaky, line 106: I think this could generally be simplified a bit... cc @patzielinski

internal/repository/targets.go
```diff
@@ -200,6 +200,66 @@ func (r *Repository) UpdateDelegation(ctx context.Context, signer sslibdsse.Sign
...
+		return policy.ErrMetadataNotFound
+	}
+
+	targetsMetadata, err := state.GetTargetsMetadata(targetsRoleName)
+	if err != nil {
+		return err
+	}
+
+	slog.Debug("Checking if all rules exist...")
+	for _, ruleName := range ruleNames {
```
adityasaky, line 228: let's not split checks between here and the policy package, a future refactor could accidentally include one but not the other, etc.

adityasaky (commented):
This is pretty close!

internal/cmd/policy/reorderrules/reorderrules.go
fr0m-scratch: It can work both ways, either repeated or not repeated

internal/policy/targets.go
fr0m-scratch: Indeed, I think the allow-rule is more like a default setting which sits below user-level abstraction. And `policy --list-rules` does not include the allow-rule either.

internal/policy/targets.go
fr0m-scratch: I haven't figure out a way to do this, but all other requested changes have been made.

adityasaky: Taking a stab at making this easier to read:


```
func ReorderDelegations(targetsMetadata *tuf.TargetsMetadata, ruleNames []string) (*tuf.TargetsMetadata, error) {
	// Create a map of all existing delegations for quick look up
	rolesMap := make(map[string]tuf.Delegation)
	// Create a set of current rules in metadata, skipping the allow rule
	currentRules := set.NewSet[string]()
	for _, delegation := range targetsMetadata.Delegations.Roles {
		if delegation.Name == AllowRuleName {
			continue
		}
		rolesMap[delegation.Name] = delegation
		currentRules.Add(delegation.Name)
	}

	specifiedRules := set.NewSet[string]()
	for _, name := range ruleNames {
		if specifiedRules.Has(name) {
			return nil, fmt.Errorf("%w: '%s'", ErrDuplicatedRuleName, name)
		}
		specifiedRules.Add(name)
	}

	onlyInSpecifiedRules := specifiedRules.Minus(currentRules)
	if onlyInSpecifiedRules.Len() != 0 {
		if onlyInSpecifiedRules.Has(AllowRuleName) {
			return nil, fmt.Errorf("%w: do not specify allow rule", ErrCannotManipulateAllowRule)
		}

		contents := onlyInSpecifiedRules.Contents()
		return nil, fmt.Errorf("%w: rules '%s' do not exist in current rule file", ErrRuleNotFound, strings.Join(contents, ", "))
	}

	onlyInCurrentRules := currentRules.Minus(specifiedRules)
	if onlyInCurrentRules.Len() != 0 {
		contents := onlyInCurrentRules.Contents()
		return nil, fmt.Errorf("%w: rules '%s' not specified", ErrRuleNotFound, strings.Join(contents, ", "))
	}

	// Create newDelegations and set it in the targetsMetadata after adding allow rule
}
```

I think this may be slightly less optimal as we will end up iterating over `ruleNames` at least twice directly and some more in the implementation of `set.Minus` (which I haven't shown here), but I think that's okay because:
a) reordering is unlikely to be a regular operation
b) some of the cost (eg. `set.Contents()`) is going to only be during error cases

I personally think the result ends up being more readable.

adityasaky: We could also only do the `.Minus` when `specifiedRules` and `currentRules` are not equal, as `.Minus` is really only needed for the error message to be helpful.

internal/policy/targets_test.go
```diff
@@ -86,6 +86,53 @@ func TestUpdateDelegation(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	targetsMetadata, err = AddDelegation(targetsMetadata, "rule-2", []*tuf.Key{key2}, []string{"path2/"}, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Contains(t, targetsMetadata.Delegations.Keys, key1.KeyID)
```
adityasaky, line 111: I appreciate the thoroughness here :smile: 


That said, I think we can remove all of these assert statements as we want to (and do) test that in `TestAddDelegation`.

internal/policy/targets_test.go
```diff
@@ -86,6 +86,53 @@ func TestUpdateDelegation(t *testing.T) {
...
+	assert.Contains(t, targetsMetadata.Delegations.Keys, key2.KeyID)
+	assert.Equal(t, key1, targetsMetadata.Delegations.Keys[key1.KeyID])
+	assert.Equal(t, key2, targetsMetadata.Delegations.Keys[key2.KeyID])
+	assert.Contains(t, targetsMetadata.Delegations.Roles, AllowRule())
+
+	ruleNames := []string{"rule-2", "rule-1"}
+	targetsMetadata, err = ReorderDelegations(targetsMetadata, ruleNames)
+	assert.Nil(t, err)
+
```
adityasaky, line 120: let's set up a table of tests here:

```
	tests := map[string]struct {
		newOrder      []string
		expectedError error
	}{
		"reverse order": {...},
		"rule not specified in new order": {...},
		"rule repeated in new order": {...},
		"unknown rule in new order": {...},
		"allow rule in new order": {...},
	}
```

I think it's also easier for us to be sure we're testing every case, which I think isn't happening right now?

adityasaky: See: https://github.com/gittuf/gittuf/blob/bb9ced2c12c8ec070e59c6f9d600ebc9a22197b1/internal/gitinterface/hash_test.go#L15-L61

internal/repository/targets_test.go
```diff
@@ -167,6 +167,87 @@ func TestUpdateDelegation(t *testing.T) {
...
+		for _, role := range targetsMetadata.Delegations.Roles {
+			initialOrder = append(initialOrder, role.Name)
+		}
+		expectedInitialOrder := append([]string{}, "protect-main")
+		expectedInitialOrder = append(expectedInitialOrder, ruleNames...)
+		expectedInitialOrder = append(expectedInitialOrder, "gittuf-allow-rule")
+		assert.Equal(t, expectedInitialOrder, initialOrder)
+
+		// Case 1: Valid Input
+		newOrder := []string{"rule-3", "rule-1", "rule-2", "protect-main"}
```
adityasaky, line 206: Ah I see, the tests are more complete here. I think it makes more sense to test in the policy package, which is where we actually catch the different errors.

adityasaky (changes requested):

internal/policy/targets_test.go
fr0m-scratch: Will do!

internal/repository/targets_test.go
fr0m-scratch: It does make sense! Now, I kept only the valid case and move the complete test suite to the policy package

internal/policy/targets.go
fr0m-scratch: I haven't found the implementation for `.Minus`, so I added a quick impl in the common/set package

internal/cmd/policy/reorderrules/reorderrules.go
```diff
@@ -0,0 +1,66 @@
...
+		&o.ruleNames,
+		"rule-name",
+		[]string{},
+		"a space-separated list of rule names, in the new order that they should appear in, from first to last",
+	)
+	cmd.MarkFlagRequired("rule-name") //nolint:errcheck
+}
+
+func (o *options) Run(cmd *cobra.Command, args []string) error {
+	o.ruleNames = append(o.ruleNames, args...)
```
adityasaky, line 36: I missed this line before: what's this for?

internal/policy/targets.go
```diff
@@ -96,6 +103,58 @@ func UpdateDelegation(targetsMetadata *tuf.TargetsMetadata, ruleName string, aut
...
+
+	specifiedRules := set.NewSet[string]()
+	for _, name := range ruleNames {
+		if specifiedRules.Has(name) {
+			return nil, fmt.Errorf("%w: '%s'", ErrDuplicatedRuleName, name)
+		}
+		specifiedRules.Add(name)
+	}
+
+	if currentRules.Len() != specifiedRules.Len() {
```
adityasaky, line 129: This check is insufficient. If currentRules is `rule-1, rule2` and specifiedRules is `rule-1, rule-3`, we should error. We explicitly need a `set.Equal()` check (which in turn _can_ use .Len() as a first heuristic to return false as two sets cannot be equal if they're of different lengths).

internal/policy/targets.go
```diff
@@ -96,6 +103,58 @@ func UpdateDelegation(targetsMetadata *tuf.TargetsMetadata, ruleName string, aut
...
+
+		onlyInCurrentRules := currentRules.Minus(specifiedRules)
+		if onlyInCurrentRules.Len() != 0 {
+			contents := onlyInCurrentRules.Contents()
+			return nil, fmt.Errorf("%w: rules '%s' not specified", ErrMissingRules, strings.Join(contents, ", "))
+		}
+	}
+
+	// Create newDelegations and set it in the targetsMetadata after adding allow rule
+	newDelegations := []tuf.Delegation{}
```
adityasaky, line 148:
```suggestion
	newDelegations := make([]tuf.Delegation, 0, len(rolesMap) + 1)
```

internal/policy/targets_test.go
```diff
@@ -86,6 +86,75 @@ func TestUpdateDelegation(t *testing.T) {
...
+		},
+		"rule not specified in new order": {
+			ruleNames:     []string{"rule-3", "rule-2"},
+			expectedError: ErrMissingRules,
+		},
+		"rule repeated in the new order": {
+			ruleNames:     []string{"rule-3", "rule-2", "rule-1", "rule-3"},
+			expectedError: ErrDuplicatedRuleName,
+		},
+		"unknown rule in the new order": {
```
adityasaky, line 134: as noted above, this should also check when the specified list is the right length. Without this, we can panic when looking up a rule in rolesMap.

internal/cmd/policy/reorderrules/reorderrules.go
fr0m-scratch: Yes, it ensures `--rule-name` works either way

internal/policy/targets.go
fr0m-scratch: Will do!

internal/cmd/policy/reorderrules/reorderrules.go
adityasaky: hm, is there anything that stops a rule name from having a space in it? This would break those cases (compared to `--rule-name "rule 1" --rule-name "rule 2"` I think.

cc @patzielinski

fr0m-scratch: If we use multiple quotes following `rule-name`, it can still parse those with space in them. Though we probably should specify that to users in the doc

Screenshot 2024-09-06 at 4 45 34 PM
Screenshot 2024-09-06 at 4 45 52 PM

adityasaky: thinking about this some more, do we even need `--rule-name` then? Can we just use `args`?

fr0m-scratch: Do you mean the rule names goes directly as `args`? I suppose we can still keep `--rule-name`, just to signal it is the start of specifying the new order

adityasaky: No, I think we could just make that clear in the docs for the subcommand. Having a flag can be confusing if we really want them to pass in args. Let's get rid of the flag and use args only, and make that clear in the command docs, thanks!

fr0m-scratch: Gocha, I'll remove the flag and update the docs

adityasaky (approved):
Some minor nits, but I think this is otherwise good to go. Great work!

fr0m-scratch:
> Some minor nits, but I think this is otherwise good to go. Great work!

Have committed the new changes, appreciate it!

adityasaky (approved):
This looks great, thanks for working on this!

internal/cmd/policy/reorderrules/reorderrules.go
```diff
@@ -0,0 +1,58 @@
...
+	err = repo.ReorderDelegations(cmd.Context(), signer, o.policyName, o.ruleNames, true)
+	return err
+}
+
+func New(persistent *persistent.Options) *cobra.Command {
+	o := &options{p: persistent}
+	cmd := &cobra.Command{
+		Use:               "reorder-rules",
+		Short:             "Reorder rules in the specified policy file",
+		Long:              "This command allows users to reorder rules in the specified policy file. By default, the main policy file is selected. The rule names need to be passed as arguments, in the new order they must appear in, starting from the first to the last rule. Rule names may contain spaces, so they should be enclosed in quotes if necessary.",
```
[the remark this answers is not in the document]
adityasaky: No, the user isn't really aware of it at all. They can only encounter it when they go looking at the raw policy metadata, it's also not part of `policy list-rules`.

PR #455, opened 2024-06-21
Add verify-mergeable

adityasaky:
~Depends on #439~

~A bunch of comments for verify-mergeable were left there, will cross reference.~

I've applied the comments in threads inline for discussion. The idea of verify-mergeable is to be able to say "this PR can be merged as it has reached the required number of approvals" or "this PR can be merged as long as the merge is by someone with approval powers". I want to plug this into gittuf/github-app to add checkruns to a repository.

### DIFF
--- a/internal/policy/verify.go
+++ b/internal/policy/verify.go
@@ -32,12 +32,13 @@ import (
 )
 
 var (
-	ErrUnauthorizedSignature   = errors.New("unauthorized signature")
-	ErrInvalidEntryNotSkipped  = errors.New("invalid entry found not marked as skipped")
-	ErrLastGoodEntryIsSkipped  = errors.New("entry expected to be unskipped is marked as skipped")
-	ErrNoVerifiers             = errors.New("no verifiers present for verification")
-	ErrInvalidVerifier         = errors.New("verifier has invalid parameters (is threshold 0?)")
-	ErrVerifierConditionsUnmet = errors.New("verifier's key and threshold constraints not met")
+	ErrUnauthorizedSignature          = errors.New("unauthorized signature")
+	ErrInvalidEntryNotSkipped         = errors.New("invalid entry found not marked as skipped")
+	ErrLastGoodEntryIsSkipped         = errors.New("entry expected to be unskipped is marked as skipped")
+	ErrNoVerifiers                    = errors.New("no verifiers present for verification")
+	ErrInvalidVerifier                = errors.New("verifier has invalid parameters (is threshold 0?)")
+	ErrVerifierConditionsUnmet        = errors.New("verifier's key and threshold constraints not met")
+	ErrCannotVerifyMergeableForTagRef = errors.New("cannot verify mergeable into tag reference")
 )
 
 // VerifyRef verifies the signature on the latest RSL entry for the target ref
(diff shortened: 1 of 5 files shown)


adityasaky:
@wlynch left comments about verify-mergeable in #439, tracking those comments here.

https://github.com/gittuf/gittuf/pull/439#discussion_r1649068232

It's a little weird that the base branch is a flag and the feature branch is an arg.
I wonder if we should change this to match git behavior?

e.g. git merge main means merge main into the current branch, so gittuf verify-mergable feature == verify that the feature branch can be merged into the current branch?

https://github.com/gittuf/gittuf/pull/439#discussion_r1649082476

Everytime I read this it throws me for a loop because we're failing open. I wonder if we should allow restricting this somewhere to require at least 1 verifier. 🤔

https://github.com/gittuf/gittuf/pull/439#discussion_r1649084802

Where is the gitObj signature being done? It's not immediately obvious from this code so this looks like we're checking the same thing twice.

https://github.com/gittuf/gittuf/pull/439#discussion_r1649085936

What is this tracking and why is it only checked the 2nd time?

https://github.com/gittuf/gittuf/pull/439#discussion_r1649088759

Also not clicking what this has to do with file rules. If this is for mergeability, e.g. "the commit has the number of required RSL entries needed to meet the threshold", then this should be renamed since it reads like "RSL entries are needed to meet the threshold"

https://github.com/gittuf/gittuf/pull/439#discussion_r1649099274

Why resolve the target ids via the RSL instead of via the refs directly?

Could the RSL potentially be behind the ref value if a new commit was pushed without updating the RSL?

https://github.com/gittuf/gittuf/pull/439#discussion_r1649100626

Taking a step back, there's a lot of verification logic here that feels like should be encapsulated elsewhere. Ideally I think this should look like:

What information do you need to be returned from verification in order to make policy decisions about mergability?

internal/cmd/verifymergeable/verifymergeable.go
```diff
@@ -0,0 +1,45 @@
...
+	cmd.MarkFlagRequired("into") //nolint:errcheck
+}
+
+func (o *options) Run(cmd *cobra.Command, args []string) error {
+	repo, err := repository.LoadRepository()
+	if err != nil {
+		return err
+	}
+
+	return repo.VerifyMergeable(cmd.Context(), o.baseBranch, args[0])
```
adityasaky, line 30: > @wlynch: It's a little weird that the base branch is a flag and the feature branch is an arg.
I wonder if we should change this to match git behavior?
>
> e.g. git merge main means merge main into the current branch, so gittuf verify-mergable feature == verify that the feature branch can be merged into the current branch?

isn't this similar to the command we have here? The arg is what will be merged from, similar to `git merge`. The `--into` could be made optional and assume the HEAD branch if empty?

internal/policy/verify.go
```diff
@@ -329,6 +329,119 @@ func VerifyRelativeForRef(ctx context.Context, repo *gitinterface.Repository, fi
...
+	if err != nil {
+		return false, err
+	}
+
+	verifiers, err := policyState.FindVerifiersForPath(fmt.Sprintf("%s:%s", gitReferenceRuleScheme, targetRef))
+	if err != nil {
+		return false, err
+	}
+
+	_, rslEntrySignatureNeededForThreshold, err := verifyGitObjectAndAttestationsUsingVerifiers(ctx, verifiers, gitinterface.ZeroHash, authorizationAttestation, approverKeyIDs, true)
```
adityasaky, line 383: > @wlynch: Where is the gitObj signature being done? It's not immediately obvious from this code so this looks like we're checking the same thing twice.

This code has changed quite a bit since this comment was left. For clarity, I've applied it here and I think it's most relevant here. In a typical verification flow, we pass in the commit ID of the RSL entry. Here, we're passing in zero because the RSL entry doesn't exist yet. So, effectively we're saying only verify the signatures on the attestation, and we say verifyMergeable=true so that the called function knows to test for threshold-1. The idea is that if we meet threshold-1 using the attestations at present, then with a valid RSL entry we can meet threshold.

internal/policy/verify.go
```diff
@@ -599,22 +716,26 @@ func getCommits(repo *gitinterface.Repository, entry *rsl.ReferenceEntry) ([]git
...
 	if len(verifiers) == 0 {
 		// This target is not protected by gittuf policy
-		return "", nil
+		return "", false, nil
 	}
 
-	verifiedUsing := ""
+	var (
+		verifiedUsing                       string
+		rslEntrySignatureNeededForThreshold bool
```
adityasaky, line 736: > @wlynch: What is this tracking and why is it only checked the 2nd time?

This is tracking if we're already meeting threshold with approval attestations or if one final approval is needed via the signature on the RSL entry. Consider a rule that has threshold = 2. This can be met in two ways:
a) signature 1 on the RSL entry for the ref and signature 2 on an attestation
b) both signatures on the attestation

In the VerifyMergeable case, we may have all the approvals needed already, meaning the signature on the RSL entry does not matter. OTOH, something may still be good to merge _as long as the merge is by an authorized user who signs the RSL entry_. So, we track this and set it to true only if threshold-1 (rather than threshold) is met.

adityasaky:
> Taking a step back, there's a lot of verification logic here that feels like should be encapsulated elsewhere. Ideally I think this should look like:
> 
> What information do you need to be returned from verification in order to make policy decisions about mergability?

I think we may have addressed this in part with the previous PR, but fresh opinion would be great!

internal/policy/verify.go
```diff
@@ -329,6 +329,119 @@ func VerifyRelativeForRef(ctx context.Context, repo *gitinterface.Repository, fi
...
+
+	slog.Debug("Loading current set of attestations...")
+	attestationsState, err := attestations.LoadCurrentAttestations(repo)
+	if err != nil {
+		return false, err
+	}
+
+	var fromID gitinterface.Hash
+	slog.Debug(fmt.Sprintf("Identifying latest RSL entry for '%s'...", targetRef))
+	targetEntry, _, err := rsl.GetLatestReferenceEntry(repo, rsl.ForReference(targetRef), rsl.IsUnskipped())
```
adityasaky, line 352: > @wlynch: Why resolve the target ids via the RSL instead of via the refs directly?
>
> Could the RSL potentially be behind the ref value if a new commit was pushed without updating the RSL?

Possibly, yes. This is consistent with everywhere else we need this info. For example, for an approval attestation, we also look at the latest RSL entry for the feature branch from which we're approving a change. Should we track this separately?

internal/policy/verify.go
```diff
@@ -329,6 +329,119 @@ func VerifyRelativeForRef(ctx context.Context, repo *gitinterface.Repository, fi
...
+	if err != nil {
+		return false, fmt.Errorf("not enough approvals to meet Git namespace policies, %w", ErrUnauthorizedSignature)
+	}
+
+	hasFileRule, err := policyState.hasFileRule()
+	if err != nil {
+		return false, err
+	}
+	if !hasFileRule {
+		return rslEntrySignatureNeededForThreshold, nil
```
adityasaky, line 393: > @wlynch: Also not clicking what this has to do with file rules. If this is for mergeability, e.g. "the commit has the number of required RSL entries needed to meet the threshold", then this should be renamed since it reads like "RSL entries are needed to meet the threshold"

That's actually accurate, we're responding saying it's mergeable _as long as the RSL entry for the merge is by an authorized user_. IMO this flexibility is useful when merges may be outsourced to bots and all approvals come in via attestations. We may not want to add the bot as authorized to the rule since that might be too much power.

adityasaky: This is consumed in repository here: https://github.com/gittuf/gittuf/pull/455/files#diff-62c1c7d5c8c9f9d2c194f625d419809db12b1a2b514bc40dd8453a092c2571afR154-R158

Maybe that needs to be slog.Info so it's always printed, or plumbed through to cmd so it can print to stdout. We could also exit with a different exit code for programmatic knowledge. :thinking:

internal/policy/verify.go
```diff
@@ -599,22 +716,26 @@ func getCommits(repo *gitinterface.Repository, entry *rsl.ReferenceEntry) ([]git
...
-	return verifyGitObjectAndAttestationsUsingVerifiers(ctx, verifiers, gitID, authorizationAttestation, approverKeyIDs)
+	return verifyGitObjectAndAttestationsUsingVerifiers(ctx, verifiers, gitID, authorizationAttestation, approverKeyIDs, verifyMergeable)
 }
 
-func verifyGitObjectAndAttestationsUsingVerifiers(ctx context.Context, verifiers []*Verifier, gitID gitinterface.Hash, authorizationAttestation *sslibdsse.Envelope, approverKeyIDs *set.Set[string]) (string, error) {
+func verifyGitObjectAndAttestationsUsingVerifiers(ctx context.Context, verifiers []*Verifier, gitID gitinterface.Hash, authorizationAttestation *sslibdsse.Envelope, approverKeyIDs *set.Set[string], verifyMergeable bool) (string, bool, error) {
 	if len(verifiers) == 0 {
 		// This target is not protected by gittuf policy
-		return "", nil
+		return "", false, nil
```
adityasaky, line 731: > @wlynch: Everytime I read this it throws me for a loop because we're failing open. I wonder if we should allow restricting this somewhere to require at least 1 verifier. 🤔

That'd be a bit strange if someone were to `verify-ref <random unprotected feature ref>`. I agree on failing open, though...

internal/policy/verify_test.go
```diff
@@ -690,6 +693,305 @@ func TestVerifyRelativeForRefUsingPersons(t *testing.T) {
...
+		}
+
+		rslSignatureRequired, err := VerifyMergeable(testCtx, repo, refName, featureRefName)
+		assert.False(t, rslSignatureRequired)
+		assert.Nil(t, err)
+	})
+
+	t.Run("base commit not zero, mergeable with higher threshold using GitHub approval, RSL entry signature required", func(t *testing.T) {
+		if runtime.GOOS == "windows" {
+			t.Skipf("skipping test '%s' in windows as the commit signature is non-deterministic, leading to unmergeable histories", t.Name())
```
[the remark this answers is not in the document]
adityasaky: no, this is a gpg thing. fwiw, i had non determinism issues across the board (flaky) so i changed how the histories are managed. I think we can remove this skip now

internal/policy/verify.go
```diff
@@ -106,6 +106,144 @@ func VerifyRefFromEntry(ctx context.Context, repo *gitinterface.Repository, targ
 	return latestEntry.TargetID, VerifyRelativeForRef(ctx, repo, fromEntry, latestEntry, target)
 }
 
+func VerifyMergeable(ctx context.Context, repo *gitinterface.Repository, targetRef, featureRef string) (bool, error) {
```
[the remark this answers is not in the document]
adityasaky: added

internal/policy/verify.go
adityasaky: I think it's safe to mark this as addressed.

internal/policy/verify.go
adityasaky: This is captured in the docstring inline.

internal/policy/verify.go
adityasaky: Filed https://github.com/gittuf/gittuf/issues/668

internal/policy/verify.go
adityasaky: I think this has been resolved, we also don't have a cmd to plumb this through to. 😄

PR #456, opened 2024-06-21
Change how test coverage is calculated

patzielinski:
This PR prevents files that don't need / have tests from cluttering up our testing coverage results. By default, it looks like go includes everything in coverage, including things like `docs/cli`, which is irrelevant. This allows us to select what should be included.

This will probably need improvement for windows due to the use of `grep`...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,4 +22,7 @@ jobs:
         go-version: '1.22'
         cache: true
     - name: Test
-      run: go test -covermode atomic ./...
+      run: go test ./...
+    - name: Test Coverage
+      if: matrix.os == 'ubuntu-latest'
+      run: go test -covermode=atomic `go list ./... | grep -v -f .test_ignore.txt`
--- a/.test_ignore.txt
+++ b/.test_ignore.txt
@@ -1,0 +1,16 @@
+github.com/gittuf/gittuf/docs/cli
+github.com/gittuf/gittuf/internal/cmd/addhooks
+github.com/gittuf/gittuf/internal/cmd/clone
+github.com/gittuf/gittuf/internal/cmd/dev
+github.com/gittuf/gittuf/internal/cmd/policy
+github.com/gittuf/gittuf/internal/cmd/profile
+github.com/gittuf/gittuf/internal/cmd/root
+github.com/gittuf/gittuf/internal/cmd/rsl
+github.com/gittuf/gittuf/internal/cmd/trust
+github.com/gittuf/gittuf/internal/cmd/verifycommit
+github.com/gittuf/gittuf/internal/cmd/verifyref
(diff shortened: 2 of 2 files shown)


neilnaveen:
We could use `go test -covermode=atomic $(go list ./... | findstr /V /G:.test_ignore.txt)` for Windows

.github/workflows/ci.yml
```diff
@@ -22,4 +22,7 @@ jobs:
         go-version: '1.22'
         cache: true
     - name: Test
-      run: go test -covermode atomic ./...
+      run: go test ./...
+    - name: Test Coverage
+      if: matrix.os == 'ubuntu-latest'
+      run: go test -covermode=atomic `go list ./... | grep -v -f .test_ignore.txt`
```
adityasaky, line 28: @patzielinski and I discussed this and we explicitly decided to keep this separate from the main test step. This is so that we don't accidentally skip running tests on ubuntu because we forgot to update .test_ignore.txt or so. Even with test performance hits (#457), ubuntu is plenty fast that we can afford to effectively run tests twice.

neilnaveen: If we are ok with only having coverage for ubuntu, then this looks good to me.

.test_ignore.txt
```diff
@@ -0,0 +1,16 @@
...
+github.com/gittuf/gittuf/internal/cmd/dev
+github.com/gittuf/gittuf/internal/cmd/policy
+github.com/gittuf/gittuf/internal/cmd/profile
+github.com/gittuf/gittuf/internal/cmd/root
+github.com/gittuf/gittuf/internal/cmd/rsl
+github.com/gittuf/gittuf/internal/cmd/trust
+github.com/gittuf/gittuf/internal/cmd/verifycommit
+github.com/gittuf/gittuf/internal/cmd/verifyref
+github.com/gittuf/gittuf/internal/cmd/verifytag
+github.com/gittuf/gittuf/internal/cmd/version
```
neilnaveen, line 13: Nit

```suggestion
github.com/gittuf/gittuf/docs/
github.com/gittuf/gittuf/internal/cmd
```

neilnaveen (approved):

.test_ignore.txt
adityasaky: github.com/gittuf/gittuf/internal/cmd/common has tests

.github/workflows/ci.yml
adityasaky: yeah, we have almost no OS specific code. The exception is testing for encrypted ssh keys on windows. I think we can safely just test on ubuntu.

.test_ignore.txt
neilnaveen: Thanks, I did not see that.

adityasaky:
@patzielinski is this waiting on anything?

adityasaky (approved):

PR #480, opened 2024-07-10
speed up windows tests

zsun6:
Trying to speed up tests
@patzielinski

### DIFF
--- a/internal/policy/policy_test.go
+++ b/internal/policy/policy_test.go
@@ -298,7 +298,9 @@ func TestStateKeys(t *testing.T) {
 }
 
 func TestStateVerify(t *testing.T) {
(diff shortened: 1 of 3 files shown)


zsun6:
Apparently when using parallel, the non-deterministic scheduling nature of it will cause a possibility of tests to fail, the more parallel implemented, the higher the possibility. It is still feasible to safely implement parallel with care and making sure that running `go test -race ./...` raises no error, but upon checking, there are little combinations of tests that can add parallel and not raise error. The tests that take more time genuinely do not support parallel. Therefore it's not worth the risk of tests having a probability of failing (which causes confusion) to implement parallel

adityasaky (approved):

PR #484, opened 2024-07-16
transport: Fetch refs/gittuf/* before creating entries

adityasaky:
Fetch remote gittuf refs to ensure new RSL entries are appened at the end

### DIFF
--- a/internal/git-remote-gittuf/curl.go
+++ b/internal/git-remote-gittuf/curl.go
@@ -158,6 +158,15 @@ func handleCurl(remoteName, url string) (map[string]string, bool, error) {
 
 						// }
 
+						log("fetching remote gittuf refs")
+						cmd := exec.Command("git", "fetch", url, "refs/gittuf/*:refs/gittuf/*")
+						cmd.Stderr = os.Stderr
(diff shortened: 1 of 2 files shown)


internal/git-remote-gittuf/curl.go
```diff
@@ -158,6 +158,15 @@ func handleCurl(remoteName, url string) (map[string]string, bool, error) {
 
 						// }
 
+						log("fetching remote gittuf refs")
+						cmd := exec.Command("git", "fetch", url, "refs/gittuf/*:refs/gittuf/*")
+						cmd.Stderr = os.Stderr
+						cmd.Stdout = os.Stderr
```
neilnaveen, line 164: Shouldn't the `cmd.Stdout = os.Stdout`

adityasaky: No,stdout is used by the transport to communicate with Git. stderr is used to communicate with the user. Any error from this subprocess should go to the user rather than be parsed by Git, it wouldn't know how to handle the output actually

wlynch (approved):

PR #488, opened 2024-07-18
Support older Git versions

### LINKED ISSUE #434: Version Incompatibilities
### Add a description

This is to track any version incompatibilities we run into with gittuf's dependencies.

- `git` binary needs to be `>= 2.36.0` due to `--format` being introduced then. This is required for the `git ls-tree` functionality tests when building gittuf.
    - Ubuntu 22.04 LTS comes with `2.34.1`

_No response_

### Code of Conduct

adityasaky:
This allows us to support Git < 2.36, needed for users still on Ubuntu 22.04.

Closes #434

### DIFF
--- a/internal/gitinterface/tree.go
+++ b/internal/gitinterface/tree.go
@@ -9,8 +9,6 @@ import (
 	"strings"
 )
 
-const tempMergeComputationRef = "refs/gittuf/tmp/merge-computation"
-
 func (r *Repository) EmptyTree() (Hash, error) {
 	treeID, err := r.executor("hash-object", "-t", "tree", "--stdin").executeString()
 	if err != nil {
@@ -94,7 +92,6 @@ func (r *Repository) GetMergeTree(commitAID, commitBID Hash) (Hash, error) {
 	}
 
 	var stdOut string
-
 	if !niceGit {
 		// Older Git versions do not support merge-tree, and, as such, require
 		// quite a long workaround to find what the merge tree is. This
@@ -109,17 +106,12 @@ func (r *Repository) GetMergeTree(commitAID, commitBID Hash) (Hash, error) {
 			return ZeroHash, fmt.Errorf("currently in detached HEAD state, please switch to a branch")
 		}
 
-		err = r.SetReference(tempMergeComputationRef, commitAID)
+		_, err = r.executor("checkout", commitAID.String()).executeString()
 		if err != nil {
-			return ZeroHash, fmt.Errorf("unable to set the merge computation ref: %w", err)
+			return ZeroHash, fmt.Errorf("unable to enter detached HEAD state: %w", err)
 		}
 
-		_, err = r.executor("checkout", tempMergeComputationRef).executeString()
-		if err != nil {
-			return ZeroHash, fmt.Errorf("unable to create merge computation branch: %w", err)
-		}
-
-		_, err = r.executor("merge", commitBID.String()).executeString()
+		_, err = r.executor("merge", "-m", "Computing merge tree", commitBID.String()).executeString()
 		if err != nil {
 			// Attempt to abort the merge in all cases as a failsafe
 			_, abrtErr := r.executor("merge", "--abort").executeString()
@@ -130,7 +122,7 @@ func (r *Repository) GetMergeTree(commitAID, commitBID Hash) (Hash, error) {
 			return ZeroHash, fmt.Errorf("unable to perform merge: %w", err)
 		}
 
-		stdOut, err = r.executor("show", "-s", "--format=%T", tempMergeComputationRef).executeString()
+		stdOut, err = r.executor("show", "-s", "--format=%T").executeString()
 		if err != nil {
 			return ZeroHash, fmt.Errorf("unable to extract tree hash of merge commit: %w", err)
 		}
--- a/internal/gitinterface/tree.go
+++ b/internal/gitinterface/tree.go
@@ -26,7 +26,12 @@ func (r *Repository) EmptyTree() (Hash, error) {
 // GetAllFilesInTree returns all filepaths and the corresponding blob hashes in
 // the specified tree.
 func (r *Repository) GetAllFilesInTree(treeID Hash) (map[string]Hash, error) {
-	stdOut, err := r.executor("ls-tree", "-r", "--format=%(path) %(objectname)", treeID.String()).executeString()
+	// From Git 2.36, we can use --format here. However, it appears a not
+	// insignificant number of developers are still on Git 2.34.1, a side effect
+	// of being on Ubuntu 22.04. 22.04 is still widely used in WSL2 environments.
+	// So, we're removing --format and parsing the output differently to handle
+	// the extra information for each entry we don't need.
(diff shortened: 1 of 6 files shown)


adityasaky:
Converting to draft. The original change is good and works on 2.34.1. However, looks like we still need to find a workaround for merge-tree as that received updates ~ 2.38.

adityasaky:
Any thoughts on computing merge-tree across these version ranges? The only thing I can think of at the moment is to actually create a merge commit in a copy of the repository :\

internal/gitinterface/utils.go
```diff
@@ -37,3 +38,35 @@ func RemoteRef(refName, remoteName string) string {
...
+	case major >= 2:
+		switch {
+		case minor >= 38:
+			return true, nil
+		default:
+			return false, nil
+		}
+	default:
+		return false, nil
+	}
```
neilnaveen, line 71: Could this instead be a if statment like this?

``` go
if major >= 2 && minor >= 38 {
    return true, nil
}  
return false, nil
```

neilnaveen (commented):
This looks good to me, just a small nit.

wlynch (approved):

internal/gitinterface/tree.go
```diff
@@ -9,6 +9,8 @@ import (
 	"strings"
 )
 
+const TempMergeComputationRef = "refs/gittuf/tmp/merge-computation"
```
adityasaky, line 12: this can be private I think

internal/gitinterface/tree.go
```diff
@@ -74,9 +88,53 @@ func (r *Repository) GetMergeTree(commitAID, commitBID Hash) (Hash, error) {
...
+		if err != nil {
+			return ZeroHash, fmt.Errorf("unable to create merge computation branch: %w", err)
+		}
+
+		_, err = r.executor("merge", commitBID.String()).executeString()
+		if err != nil {
+			return ZeroHash, fmt.Errorf("unable to perform merge: %w", err)
+		}
+
+		stdOut, err = r.executor("show", "-s", "--format=%P", TempMergeComputationRef).executeString()
```
adityasaky, line 123: Isn't this listing the parents of the merge commit? Do we have a gap in our tests?

internal/gitinterface/tree.go
```diff
@@ -74,9 +88,53 @@ func (r *Repository) GetMergeTree(commitAID, commitBID Hash) (Hash, error) {
...
+		if err != nil {
+			return ZeroHash, fmt.Errorf("unable to extract tree hash of merge commit: %w", err)
+		}
+
+		// Switch back to the branch the user was on
+		_, err = r.executor("checkout", currentBranch).executeString()
```
adityasaky, line 129: Can we merge without checkouts (and thus modifications to the worktree?

internal/gitinterface/utils.go
```diff
@@ -37,3 +38,35 @@ func RemoteRef(refName, remoteName string) string {
 
 	return remotePath
 }
+
+// IsNiceGitVersion determines whether the version of git is "nice". Certain Git
+// subcommands that gittuf uses were added in newer versions than some common
+// client versions. Instead of using a workaround for all clients, we determine
+// if we can use the newer features or instead need to use workarounds.
+func IsNiceGitVersion() (bool, error) {
```
adityasaky, line 46: I think this can be private. Also, we can leave this be by now but if we encounter other cases where we have different minimum version constraints, we could have the constraint as a parameter here. Let's assess if / when we need it though.

internal/gitinterface/utils.go
```diff
@@ -37,3 +38,35 @@ func RemoteRef(refName, remoteName string) string {
...
+	cmd := exec.Command("git", "--version")
+	output, err := cmd.Output()
+	if err != nil {
+		return false, err
+	}
+
+	versionString := strings.TrimPrefix(strings.TrimSpace(string(output)), "git version ")
+	var major, minor, patch int
+
```
adityasaky, line 55:
```suggestion
```

internal/gitinterface/utils.go
```diff
@@ -37,3 +38,35 @@ func RemoteRef(refName, remoteName string) string {
...
+// if we can use the newer features or instead need to use workarounds.
+func IsNiceGitVersion() (bool, error) {
+	cmd := exec.Command("git", "--version")
+	output, err := cmd.Output()
+	if err != nil {
+		return false, err
+	}
+
+	versionString := strings.TrimPrefix(strings.TrimSpace(string(output)), "git version ")
+	var major, minor, patch int
```
adityasaky, line 54:
```suggestion

	var major, minor, patch int
```

internal/gitinterface/utils.go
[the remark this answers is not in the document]
adityasaky: Rather than enumerating capabilities, let's assess passing in the minimum version instead when the time comes. That way, the caller can control what the minimum version is where we'd need to know anyway.

internal/gitinterface/utils.go
adityasaky: Prefer the if statements as well.

internal/gitinterface/tree.go
```diff
@@ -74,9 +88,53 @@ func (r *Repository) GetMergeTree(commitAID, commitBID Hash) (Hash, error) {
...
+	}
+
+	var stdOut string
+
+	if !niceGit {
+		// Older Git versions do not support merge-tree, and, as such, require
+		// quite a long workaround to find what the merge tree is. This
+		// workaround boils down to:
+		// Create new branch > Merge into said branch > Extract tree hash
+		currentBranch, err := r.executor("branch", "--show-current").executeString()
```
adityasaky, line 103: can we run into trouble if we're in a detached head state here?

internal/gitinterface/tree.go
```diff
@@ -74,9 +88,53 @@ func (r *Repository) GetMergeTree(commitAID, commitBID Hash) (Hash, error) {
...
+			return ZeroHash, fmt.Errorf("unable to perform merge: %w", err)
+		}
+
+		stdOut, err = r.executor("show", "-s", "--format=%T", tempMergeComputationRef).executeString()
+		if err != nil {
+			return ZeroHash, fmt.Errorf("unable to extract tree hash of merge commit: %w", err)
+		}
+
+		// Switch back to the branch the user was on
+		_, err = r.executor("switch", currentBranch).executeString()
```
adityasaky, line 129: let's use switch / checkout consistently

internal/gitinterface/tree.go
```diff
@@ -74,9 +88,53 @@ func (r *Repository) GetMergeTree(commitAID, commitBID Hash) (Hash, error) {
...
+		if err != nil {
+			return ZeroHash, fmt.Errorf("unable to set the merge computation ref: %w", err)
+		}
+
+		_, err = r.executor("checkout", tempMergeComputationRef).executeString()
+		if err != nil {
+			return ZeroHash, fmt.Errorf("unable to create merge computation branch: %w", err)
+		}
+
+		_, err = r.executor("merge", commitBID.String()).executeString()
```
adityasaky, line 118: we should also see if merge --abort is necessary during conflicts.

PR #490, opened 2024-07-19
Verify file rules in specified branches only

### LINKED ISSUE #355: Verify file rules in certain branches only
### Add a description

Currently, file rules are applied on a per-commit basis regardless of the branch being verified. This means rules with threshold >=2 cannot be satisfied even in feature branches. We ought to support file rules against specific branches (possibly with patterns as well), and maybe even make that the default behavior.

_No response_

### Code of Conduct

patzielinski:
This PR adds functionality to verify file rules in specified branches only. Specifically,

- `add-branch-rule` and `add-file-rule` are new commands which take care of building their respective type of rules for the user. These are in addition to the existing workflow provided by `add-rule`, which allows unrestricted input.
- The `Matches` method in the `tuf` package now supports passing in branches in addition to a file pattern, enabling the namesake functionality of this PR.

**Note**: The changes in this PR now treat file rules without branch information as applying to all branches. For example, `file:README.md` will apply to `README.md` in **all** branches, while `file:README.md;git:refs/heads/main` will only apply to `README.md` in the `main` branch.

~~Still WIP, need to add some tests.~~

**Edit**: `add-branch-rule` and `add-file-rule` postponed to a future PR.

Closes #355.

### DIFF
--- a/internal/cmd/policy/listrules/listrules.go
+++ b/internal/cmd/policy/listrules/listrules.go
@@ -41,10 +41,10 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 		fmt.Printf(strings.Repeat("    ", curRule.Depth)+"Rule %s:\n", curRule.Delegation.Name)
 		gitpaths, filepaths := []string{}, []string{}
 		for _, path := range curRule.Delegation.Paths {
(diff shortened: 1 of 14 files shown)


adityasaky:
This looks like a major rebase is needed, let's revisit.

PR #504, opened 2024-08-02
Changed GetSigningCommand()

vamsikoneti:
Changed GetSigningCommand() to GetSigningConfig(), along with removing the program data while keeping the keyinfo

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,70 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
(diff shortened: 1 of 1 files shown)


adityasaky:
Hi @vamsikoneti, I'm not sure what editor you're using, but Go has an opinionated formatting tool that basically all projects use. I suggest configuring your editor to `go fmt` your file when you save, usually the Go language plugin for the editor in question takes care of that.

internal/gitinterface/keys.go
```diff
@@ -3,147 +3,121 @@
 package gitinterface
 
 import (
-	"errors"
+    "errors",
```
adityasaky, line 6: There shouldn't be a comma here. Also note that Go plugins should handle imports for you automatically too when you use something from the package in question.

internal/gitinterface/keys.go
```diff
@@ -3,147 +3,121 @@
...
-	ErrInvalidSignature           = errors.New("unable to parse signature / signature has unexpected header")
-)
-
-type SigningMethod int
-
-const (
-	SigningMethodGPG SigningMethod = iota
-	SigningMethodSSH
-	SigningMethodX509
+type SigningMethod intconst (
```
adityasaky, line 18: there should be a newline after int, unsure why this is collapsed.

PR #507, opened 2024-08-06
Fix Display Test Paging Issue

Raghava-Gatadi:


internal/display/display_test.go
```diff
@@ -47,7 +48,9 @@ func TestNewDisplayWriter(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			if gotOutput := defaultOutput.String(); gotOutput != tt.wantOutput {
+			// Trim the output to remove any extra whitespace or newlines
+			gotOutput := strings.TrimSpace(defaultOutput.String())
+			if gotOutput != tt.wantOutput {
```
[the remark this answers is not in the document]
Raghava-Gatadi: Done, reverted back to one-line format

adityasaky:
why is this specifically an issue on windows when we don't have newlines anyway? is this a _windows implementation issue in the go stdlib? also can we test on powershell (rather than something bash-like) in github actions?

Raghava-Gatadi:
In Windows, the end of the line is interpreted as \r\n. This discrepancy causes issues with newline handling in tests. I confirmed this by updating the test case as follows:

```
{
	name:            "simple test with paging",
	page:            true,
	contents:        []byte("Hello, world!"),
	wantOutput:      "Hello, world!\r\n",
	wantErrorOutput: "",
}
```
In this case, the test passed. After adding the trim function, the \r\n characters were removed, allowing the test to pass with wantOutput: "Hello, world!".

Regarding the PowerShell testing, I ran the following commands:

- `go test ./...` on Unix-like OS

- `pwsh -c 'go test ./...'` on Windows

All tests passed successfully. These commands are similar to those provided in the [GitHub documentation](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-powershell), ensuring compatibility with both Unix-like systems and Windows PowerShell.

adityasaky:
I think my question was more-so that we don't currently expect just `\n`, we don't expect newlines at all in the output. So I'm curious why only windows is injecting its newline sequence (`\r\n`). Effectively, we've discovered the display writer is _not_ writing what we expect it to, but is adding additional bytes for newlines.

internal/display/display_test.go
```diff
@@ -47,7 +48,8 @@ func TestNewDisplayWriter(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			if gotOutput := defaultOutput.String(); gotOutput != tt.wantOutput {
+			// Trim the output to remove any extra whitespace or newlines
+			if gotOutput := strings.TrimSpace(defaultOutput.String()); gotOutput != tt.wantOutput {
```
adityasaky, line 52: let's trim space on GOOS == windows.

Raghava-Gatadi: Are you asking to execute the `trimSpace` function only if the OS is Windows? If so, I've made the necessary changes; please check them.

adityasaky:
Looks like something's gone wonky with readme? 😄

Raghava-Gatadi:
Sorry, I was testing something else with Git, and this branch got pushed.

Raghava-Gatadi:
Now, please check if the changes are as per your request.

internal/display/display_test.go
```diff
@@ -46,8 +48,14 @@ func TestNewDisplayWriter(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
+			gotOutput := ""
```
adityasaky, line 51:
```
gotOutput := defaultOutput.String()
if runtime.GOOS == "windows" {
    gotOutput = strings.TrimSpace(gotOutput)
}
```

adityasaky (approved):
Thanks!

PR #520, opened 2024-08-18
[WIP] feat: adding TUI for managing rules in policy files

fr0m-scratch:
This pr adds TUI for managing rules in policy files. It is currently a very rough prototype that can be called with `gittuf policy TUI --signing-key [key]`.


internal/cmd/policy/TUI/TUI.go
adityasaky, line 1: no need to capitalize the TUI in either the dir or file name.

PR #531, opened 2024-08-30
Track verified states for refs locally

### LINKED ISSUE #245: Track verified states locally

adityasaky:
This allows us to not have to verify the full history of the repository each time.

Current status: it works but is untested and needs clean up. At the time of opening, I'm mostly just saving state to github.

ETA: looking at CI, it works _when you hold it particularly right_ seems to be a better description. :smile: 

Closes #245 

TODOS:
- [ ] Add tests for cache
- [x] Add tests for rsl.Get methods that use the entry number
- [x] Use non global cache variable
- [ ] Add tests for verifyrelativeforref that doesn't verify from the start every time
- [ ] Fix up expected user interface -> we shouldn't have verify-ref-cache by the time this will be merged
- [ ] Add tests where a repository goes from not having numbers to having numbers

### DIFF
--- a/docs/cli/gittuf.md
+++ b/docs/cli/gittuf.md
@@ -21,5 +21,6 @@ A security layer for Git repositories, powered by TUF
 * [gittuf rsl](gittuf_rsl.md)	 - Tools to manage the repository's reference state log
 * [gittuf trust](gittuf_trust.md)	 - Tools for gittuf's root of trust
 * [gittuf verify-ref](gittuf_verify-ref.md)	 - Tools for verifying gittuf policies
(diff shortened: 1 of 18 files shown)


neilnaveen:
Could we sign commits that are in the local cache ref? Maybe a flag to verify with signing?

adityasaky:
Closing in favor of #716.

PR #532, opened 2024-09-01
rsl: Add incrementing number to RSL entry

adityasaky:
This commit adds a uint64 Number field to RSL entries. This allows us to determine the ordering of distinct RSL entries, enabling optimizations such as knowing which policy to use for an entry without always walking the RSL.

The implementation maintains backwards compatibility, so the numbering can start from whenever this is enabled. This is made easy by the fact that we can guarantee gittuf is updated everywhere it is currently used.

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -3,9 +3,7 @@
 package common
 
 import (
-	"encoding/pem"
 	"fmt"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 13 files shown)


adityasaky:
Part of #245, WIP

internal/rsl/rsl.go
```diff
@@ -442,6 +562,105 @@ func GetLatestReferenceEntryForRefBefore(repo *gitinterface.Repository, refName
...
+}
+
+// GetLatestReferenceEntryForRefBeforeUntilNumber returns the latest reference
+// entry available locally in the RSL for the specified refName before the
+// specified anchor. The search is terminated when an entry with number less
+// than the until value passed in is encountered. Note that until is inclusive:
+// if the entry in question is for the requested reference, it is returned. If
+// the latest entry is not numbered (e.g., in a repository that predates this
+// feature), this function is identical to GetLatestReferenceEntryForRefBefore.
+func GetLatestReferenceEntryForRefBeforeUntilNumber(repo *gitinterface.Repository, refName string, anchor gitinterface.Hash, until uint64) (*ReferenceEntry, []*AnnotationEntry, error) {
```
neilnaveen, line 576: Could this func name be a bit smaller/simplified?

adityasaky: Probably. Do you have any suggestions?

neilnaveen: Could we change the UntilNumber to Anchor, or somthing else?

adityasaky: Anchor means an entry ID (ie its commit ID) rather than number, which is why I was trying to avoid confusion.

internal/rsl/rsl.go
```diff
@@ -75,6 +82,18 @@ func (e *ReferenceEntry) GetID() gitinterface.Hash {
 
 // Commit creates a commit object in the RSL for the ReferenceEntry.
 func (e *ReferenceEntry) Commit(repo *gitinterface.Repository, sign bool) error {
+	latestEntry, err := GetLatestEntry(repo)
+	if err == nil {
+		e.Number = latestEntry.GetNumber() + 1
+	} else {
+		if errors.Is(err, ErrRSLEntryNotFound) {
+			// First entry
+			e.Number = 1
```
wlynch, line 91: We should add docs for this, particularly if we're not starting index 0

internal/rsl/rsl.go
```diff
@@ -174,6 +228,48 @@ func (a *AnnotationEntry) Commit(repo *gitinterface.Repository, sign bool) error
...
+	if err == nil {
+		a.Number = latestEntry.GetNumber() + 1
+	} else {
+		if errors.Is(err, ErrRSLEntryNotFound) {
+			// First entry -> can an annotation actually be first? TODO
+			a.Number = 1
+		} else {
+			return err
+		}
+	}
```
wlynch, line 253: This looks highly duplicated from ReferenceEntry and AnnotationEntry. Can we combine these at all?

I'm worried these may skew in unintended ways in the future.

internal/rsl/rsl.go
wlynch: bikeshed: I'd rename this to GetLatestReferenceEntryBeforeNumber and rename until to number or n or something similar.

- `ReferenceEntryForRef` is redundant
- `until uint64` is going to look like an unix time out of context.

If you're worried about confusion with GetLatestReferenceEntryForRefUntilNumber make this variadic with options and do:

```go
GetLatestReferenceEntry(repo, ref, option.WithAnchor(hash), option.BeforeNumber(n))
```

This would let us cut down on a significant amount of the copy+pasta in this file.

internal/rsl/rsl.go
adityasaky: Added to docstrings

internal/rsl/rsl.go
adityasaky: moved to helper to `setEntryNumber` appropriately

internal/rsl/rsl.go
adityasaky: we should open a ticket for the completely untested conditions but also we probably want to test for combinations of conditions. I think the combinations we use currently are tested, luckily.

internal/common/common.go
```diff
@@ -37,84 +35,44 @@ var (
 // **signed** reference entry using the specified GPG key. It is used to
 // substitute for the default RSL entry creation and signing mechanism which
 // relies on the user's Git config.
+//
+// Update: This helper just wraps around CommitUsingSpecificKey in the rsl
+// package. We can probably get rid of it, but it's a pretty big delta.
```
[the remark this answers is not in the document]
adityasaky: I think it's fine to have a thin wrapper...

PR #536, opened 2024-09-04
ci: Update gittuf record workflows

adityasaky:
Summary:
* Set key permission to 600 for SSH signing key
* Use setup-gitsign workflow instead of installing from source
* Bump gittuf-installer to new release


adityasaky:
Related: https://github.com/in-toto/attestation-verifier/pull/41

wlynch (approved):

PR #539, opened 2024-09-04
Prepare v0.6.0 release

patzielinski:
This PR  adds the changelog for and prepares `v0.6.0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 This file tracks the changes introduced by gittuf versions.
 
+## v0.6.0
+
+- Added command to reorder policy rules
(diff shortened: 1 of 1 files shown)


adityasaky (approved):

Yasho-Bapat:
Leaving this here so it doesn't get forgotten:
To set the command line alias (and change the one that winget adds automatically), we can specify one in the manifest as such:
```
Commands:
- gittuf
```
This will go right under `InstallerType: portable` in the installer manifest.

Leaving this comment under this PR because we'll make this change when 0.6.0 goes to winget.

adityasaky:
What's the pipeline for updates on winget right now?

PR #543, opened 2024-09-06
policy: Refactor Verify APIs

adityasaky:
VerifyRelativeForRef is now responsible for identifying the right policy and attestation states to use and all other exported Verify APIs invoke it.

This is part of the same series as #532, #531, etc. It consolidates the functionality for identifying the policy so that it's easier to add support for reading from the persistent cache.


internal/policy/verify.go
```diff
@@ -108,52 +91,66 @@ func VerifyRefFromEntry(ctx context.Context, repo *gitinterface.Repository, targ
...
-	state, err := LoadState(ctx, repo, initialPolicyEntry)
-	if err != nil {
-		return err
+	// If firstEntry is for policy, we don't walk the RSL here
+	slog.Debug(fmt.Sprintf("Loading policy applicable at first entry '%s'...", firstEntry.ID))
+	var initialPolicyEntry *rsl.ReferenceEntry
+	if firstEntry.RefName == PolicyRef {
+		initialPolicyEntry = firstEntry
+	} else {
+		initialPolicyEntry, _, err = rsl.GetLatestReferenceEntryForRefBefore(repo, PolicyRef, firstEntry.ID)
```
adityasaky, line 115: This is a more explicit change from what we were doing before because now the start point for verification can be anywhere. Earlier, we had an implicit assumption it'd be from start of history.

So now, if the first entry is for policy, it's loaded to verify subsequent entries. If the first entry is not for policy and there's no policy entry before it, then the first entry must be _the very first entry_ in the RSL. So this supports the earlier case, but now checks explicitly.

Also, another change is how the very first entry case is handled. The odds are that the very first entry is for `policy-staging`, which we previously loaded and switched to the `policy` entry when it was encountered subsequently. Now, this isn't the case. We leave the policy unset so that `currentPolicy` is only ever set to an entry that was "applied" into `refs/gittuf/policy`. If everything is well, currentPolicy is set before any real entry is verified. If it's not set, then we error saying policy wasn't found. This closes a bug we previously had where an unapplied policy could be used to verify some RSL entry if the policy was not applied before said entry was verified.

adityasaky: Just realized my TODO to add a test checking for this case wasn't addressed where we have policy-staging <- main.

adityasaky: As it's written now, it'd fail. We could loosen this requirement: just leave policy unset so if the workflow encounters a non refs/gittuf ref entry, verification still fails as we didn't see a refs/gittuf/policy first.

adityasaky: I've taken out the check for now.

internal/policy/verify.go
```diff
@@ -108,52 +91,57 @@ func VerifyRefFromEntry(ctx context.Context, repo *gitinterface.Repository, targ
...
+		err                 error
 	)
 
 	// Load policy applicable at firstEntry
-	slog.Debug("Loading initial policy...")
-	state, err := LoadState(ctx, repo, initialPolicyEntry)
-	if err != nil {
-		return err
+	// If firstEntry is for policy, we don't walk the RSL here
+	slog.Debug(fmt.Sprintf("Loading policy applicable at first entry '%s'...", firstEntry.ID.String()))
```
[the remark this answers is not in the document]
adityasaky: removing policyEntry shouldn't affect the check that lastEntry is after firstEntry. but yes, the rsl API to identify all entries in the range will ensure that for us.

wlynch (approved):

PR #559, opened 2024-09-25
WIP: Add notes on managing versions and compatibility

adityasaky:
This is more notes than a concrete plan at this point for how we manage compatibility with versions of attestation predicates, policy metadata, and gittuf implementation itself which encompasses the rest and workflows.

### DIFF
--- a/docs/gaps/versioning-policy/README.md
+++ b/docs/gaps/versioning-policy/README.md
@@ -1,0 +1,322 @@
+# gittuf Versioning Policy
+
+## Metadata
+
+* **Number:** TBD
+* **Title:** gittuf Versioning Policy
+* **Implemented:** No
+* **Withdrawn/Rejected:** No
+* **Sponsors:** Aditya Sirish A Yelgundhalli (adityasaky)
+* **Last Modified:** December 31, 2025
+
+## Abstract
+
+As gittuf is intended to be used in a distributed manner by developers
+contributing to a repository, we need to have a plan in place for version skews.
+It will inevitably be the case that different developers have different versions
+of the client. Also, different versions of the client may support different
+versions of gittuf's policy and attestation schemas. If not handled with care,
+this can lead to unintended consequences, including degraded security
+guarantees.
+
+This GAP details how the gittuf implementation handles version updates and
+mismatches. The owners of a repository are empowered to set the supported
+versions of the policy metadata and attestation predicates, and the GAP details
+how clients must behave in different scenarios involving version mismatches.
+The emphasis is on ensuring that security guarantees are not degraded while also
+ensuring usability for developers using older versions of the gittuf client.
+
+## Motivation
+
+Over time, gittuf will evolve to support new features and security guarantees.
+This will lead to the creation of new schema versions for policy metadata and
+attestation predicates. It is necessary to have a systematic way to handle the
+introduction of new metadata versions, or such changes may lead to degraded
+security guarantees (e.g., a gittuf client fails to enforce new security checks
+it cannot understand) and poor usability.  The aim of this GAP is to assess
+different versioning strategies and settle on one that sufficiently balances
+security and usability.
+
+## Reasoning
+
+This GAP considers different versioning strategies and their tradeoffs. The key
+difference between the strategies is the entity responsible for selecting the
+metadata versions used in a repository.
+
+### Repository Owners Select Versions
+
+In this strategy, the owners of a repository (i.e., the root key holders)
+determine the gittuf metadata versions used in the repository at any point in
+time. This is the strategy adopted in this GAP's
+[specification](#specification). This approach ensures that the repository, at
+any point in time, has a consisent set of metadata versions, i.e., a policy
+state cannot have mismatched metadata versions. This prevents potential security
+gaps that may be caused by mixing different versions (e.g., a policy metadata
+file relies on a security guarantee introduced in another file in its version,
+but the corresponding file is of an older version that does not have the
+guarantee). In addition, this approach allows repository owners to determine
+when the repository should upgrade to newer versions, allowing them to plan for
+such upgrades (e.g., informing developers contributing to the repository to
+update their clients).
+
+#### Why must a client abort for unsupported newer versions?
+
+In this approach, the [specification](#specification) requires clients to abort
+when they encounter a newer version of metadata they do not support.  This is
+important because while this may be inconvenient for the user, it prevents
+security guarantees from being degraded. For example, if a client that does not
+support a newer version of policy metadata is used for verification, not
+aborting means that the client may allow changes that would have been prevented
+by a client that understands the newer version of the policy metadata.
+
+#### Why must a client only warn when using older versions?
+
+In this approach, the [specification](#specification) requires clients to only
+print warnings when they encounter older versions of metadata than they support.
+This is important to ensure users are aware of newer versions that may have
+improved security guarantees. However, the client must not abort the operation
+as this would lead to poor usability. For example, if a client encounters an
+older version of policy metadata during verification, aborting would prevent
+verification of historical repository activity.  Crucially, this does not lead
+to degraded security guarantees as the client still understands the older
+version and there is no expectation of enforcing security guarantees introduced
+in newer versions.
+
+### Clients Select Versions
+
+In this strategy, the client determines the versions of metadata used in a
+repository. For example, when creating a new attestation, the client may use the
+latest version of the attestation predicate it supports. Similarly, when
+creating new policy metadata (e.g., a new rule file), the client may use the
+latest version of the policy metadata it supports. While this approach improves
+usability as developers can use their existing clients without worrying about
+version mismatches, it can lead to security gaps. For example, an older client
+may inadvertently use a version of the policy metadata that has reduced security
+guarantees. Similarly, this can lead to inconsistent policy states where
+different policy metadata files have different versions, leading to potential
+incompatibilities.
+
+#### What if there are limits to which version a client can use?
+
+Some of the concerns stated with letting clients select versions can be
+mitigated by enforcing constraints on the versions of metadata that can be used.
+But, this leads to added complexity. First, this needs to be implemented
+correctly in the client. Second, the reason why certain versions are disallowed
+may not be clear to the user, leading to confusion. Third, this still does not
+prevent inconsistent policy states from being created, leading to potential
+incompatibilities and security gaps. Finally, this approach complicates the
+development of new policy schemas; the gittuf developers must also consider
+permutations of compatibility across metadata versions, which is more
+complicated than having to build translations between one policy version to the
+next.
+
+## Specification
+
+Currently, gittuf versions the following pieces:
+* attestation predicates: individual in-toto attestation predicates are
+  versioned in the predicate type URI; this is inherited from the in-toto
+  attestation specification
+* gittuf implementation: the implementation itself is versioned and encompasses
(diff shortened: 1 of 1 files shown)


docs/gaps/versioning-policy/README.md
```diff
@@ -0,0 +1,215 @@
...
+* **Withdrawn/Rejected:** No
+* **Sponsors:** Aditya Sirish A Yelgundhalli (adityasaky)
+* **Last Modified:** October 25, 2025
+
+## Abstract
+
+As gittuf is intended to be used in a distributed manner by developers
+contributing to a repository, we need to have a plan in place for version skews.
+It will inevitably be the case that different developers have different versions
+of the client. And, different versions of the client may support different
```
JustinCappos:
```suggestion
of the client. Also, different versions of the client may support different
```

docs/gaps/versioning-policy/README.md
```diff
@@ -0,0 +1,215 @@
...
+* gittuf policy metadata
+* gittuf implementation and behavior
+
+Note that all versioning follows semantic versioning.
+
+### General Guidelines
+
+Whenever a client encounters a newer metadata version than that supported by the
+client, it must prompt the user to update the client. This is the case for minor
+and patch version bumps as well.
```
JustinCappos: Whenever?  So if a repo contains 10 files, I would get prompted 10 times to ask if I want to update?

How do we stop this from being used to annoy users by stating a future version that doesn't exist?

docs/gaps/versioning-policy/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Client Workflows for Cross-Version Scenarios
+
+There are several aspects of version compatibility to consider.
+
+#### Attestation Predicate Versions
+
+##### Signing an older version of a predicate using a newer client
+
+If the predicate is of the same major version, the client must add a signature
```
JustinCappos: This needs more lead-in to explain the problem / solution sketch / idea.

docs/gaps/versioning-policy/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+If the predicate is from a previous major version, the client must generate an
+equivalent predicate with the current major version. Out of band, the user who
+generated the older predicate version must be prompted to update their client
+and add their signature to the new attestation with the current predicate
+version.
+
+##### Signing a newer version of a predicate using an older client
+
+If the predicate is of the same major version, the client must add a signature
```
JustinCappos: Same initial sentence as prior para.  Should combine and mention this is always done.

docs/gaps/versioning-policy/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+While there are similar scenarios for policy metadata versions as predicate
+versions, the client behavior is different. This is because all of the policy
+metadata in a policy state are related to each other, unlike attestations which
+exist in isolation. So, all metadata that makes up a policy state must have the
+same major policy version.
+
+##### Signing an older version of a policy file using a newer client
+
+The newer client can sign an older minor or patch version of the policy
```
JustinCappos: Why is this okay?

docs/gaps/versioning-policy/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+##### Signing a newer version of a policy file using an older client
+
+The older client may sign an older minor or patch version of the policy
+metadata, though it is unable to make other changes to the schema. The user must
+be alerted to update their client.
+
+##### Creating new policy metadata
+
+A developer's client must generate policy metadata using the same major policy
```
JustinCappos: Still missing the big picture

docs/gaps/8/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Security
+
+The GAP considers the impact of mixed versioning for attestation predicates and
+the policy metadata. Generally, the GAP proposes mechanisms to prevent rollback
+attacks for both types of metadata. In addition, the GAP ensures clients fail in
+actions that modify or verify metadata the client does not understand. This
+ensures that an older client does not allow a change that would be prevented
+with newer security features the client does not understand.
+
```
JustinCappos, line 318: Would it be sane for a future version of gittuf to also be able to set a minimum metadata version to deprecate them?  For example, v4.0 clients no longer need to support v0.9 metadata.  This would make it so clients could remove stale policy verification logic.

docs/gaps/8/README.md
```diff
@@ -0,0 +1,215 @@
...
+## Backwards Compatibility
+
+This GAP does not have impact on backwards compatibility of the existing gittuf
+design and implementation. Indeed, the GAP introduces granular versioning of
+metadata and behavior, which impacts how gittuf handles backwards compatibility
+going forward.
+
+## Security
+
```
JustinCappos, line 307: I don't see this document addressing how to handle mixed data from a variety of versions.  Does this get included in the version update notes?

JustinCappos (commented):
Needs a lot more high level description before I can really grok it.

docs/gaps/8/README.md
```diff
@@ -0,0 +1,215 @@
...
+* **Last Modified:** October 25, 2025
+
+## Abstract
+
+As gittuf is intended to be used in a distributed manner by developers
+contributing to a repository, we need to have a plan in place for version skews.
+It will inevitably be the case that different developers have different versions
+of the client. And, different versions of the client may support different
+versions of gittuf's policy and attestation schemas. If not handled with care,
+this can lead to unintended consequences, including degraded security
```
JustinCappos, line 19: I'm more worried about how this works from a compatibility standpoint.  Will users need to update their client?

docs/gaps/versioning-policy/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+Currently, gittuf versions the following pieces:
+* attestation predicates: individual in-toto attestation predicates are
+  versioned in the predicate type URI; this is inherited from the in-toto
+  attestation specification
+* gittuf implementation: the implementation itself is versioned and encompasses
+  the behavior / workflows as well as the default versions of the metadata
+  generated by the client at that version
+
+The gittuf policy metadata is not versioned for its schema, which will be
```
JustinCappos: What do you mean by "is not versioned for its schema"?

adityasaky: When this was written, we had a `version` field like in TUF which was an incrementing integer indicating version of the metadata file. We didn't have versioning for the metadata schema which we introduced here. I'll massage the text a bit.

docs/gaps/8/README.md
adityasaky: Sorry, how what works? Overall, yes, we say in this document in certain flows that we abort a workflow and ask users to update their client. I don't know that we can indefinitely let users postpone updating their clients...

docs/gaps/versioning-policy/README.md
adityasaky: It's up to the root key holders to update the policy version used in the repository. In the meantime, we should not prevent making rule changes or similar just because a newer policy schema version is available. I'm about to push a rework of this doc that indicates a warning to update the policy metadata should be printed.

docs/gaps/versioning-policy/README.md
adityasaky: Can you elaborate what you're looking for re big picture? This is strictly about defining client behavior written for people writing gittuf clients.

docs/gaps/8/README.md
adityasaky: I think no. We would of course disallow creating new policy metadata using v0.9 instead of vLatest supported by v4.0 client, but we should support it so that a repository which was created when v0.9 was latest can continue to have that old policy metadata still verified.

adityasaky: I'm about to push a rework that makes this explicit, in that clients must continue supporting older versions to ensure verification of historic activity works.

docs/gaps/8/README.md
adityasaky: Can you clarify what you mean by mixed data?

docs/gaps/versioning-policy/README.md
adityasaky: You're only ever operating over a single policy file at a time. This doesn't have any relation to the number of files in the repository. If the user runs 10 commands to modify 10 files, then yes they'll see 10 warnings.

docs/extensions/metadata-versioning.md
```diff
@@ -0,0 +1,102 @@
...
+client notices that there is already an authorization attestation for the change
+created by Alice and tries to add his signature to the existing attestation.
+
+#### Signing an older version of a predicate using a newer client
+
+If the predicate is of the same major version, this should be allowed. The
+validation of the predicate during the signing flow should use the validator for
+the version in question.
+
+TODO: major version drifts
```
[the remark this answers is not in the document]
adityasaky: My latest push does some of this.

tl;dr root key holders specify the explicit version used for all policy metadata with no drift permitted for minor / patch versions. root key holders can also specify a range of versions allowed for attestation predicates to enable flexibility in a client choosing a version when creating an attestation.

docs/gaps/versioning-policy/README.md
JustinCappos: I'm trying to understand the goals here and what you're trying to accomplish.  I'm getting a bit lost in: you must do X in case Y.

docs/gaps/8/README.md
JustinCappos: I mean metadata created by different versions of clients.  I'm assuming that different clients may have different versions of metadata support.  Can a single repo mix those?

docs/gaps/8/README.md
JustinCappos: Okay, so, can you be clear about the assumptions like this upfront?  I agree that users cannot indefinitely postpone this.  However, the frequency at which they need to do this and method they use to do so is important

docs/gaps/8/README.md
adityasaky: The edits I made today largely prevents that. For policy, the client must use the same version as that used by the current root of trust metadata. So we can't have mixed versions for policy metadata.

For attestations, the root can specify a range within which predicates are acceptable that the client has to abide by. So, we can have some mixed versions for predicates (eg an approval attestation). This is still fine because each predicate, irrespective of its version, has to provide the same set of information. The reference authorization predicate v0.1 and v0.2 can differ in field names and even include extra information in v0.2 but the information extracted from them during verification are the same. If we want to fundamentally change the set of information that must be used from the attestation, then we must do so as a new predicate altogether. This means that a repository can safely use v0.1 and v0.2 of a predicate (if allowed by root key holders).

We can separately add this to the design-document re attestations.

docs/gaps/versioning-policy/README.md
adityasaky: Please take a look at today's edits. I'm not certain it addresses your thought here though.

> you must do X in case Y.

This is intentional in the client behavior subsections. I've updated the reasoning section to more broadly discuss why we abort when the client sees a newer version of metadata than it supports while it only warns when it sees an older version than it supports.

docs/gaps/8/README.md
adityasaky: The current draft **requires** updates when the root key holders of the repo update to a metadata version which the client doesn't yet support. I can add this upfront in that client updates are only **required** at the discretion of the repository's owners.

adityasaky: I've added a note up front about this.

PratGit1606:
Will be working on this with @adityasaky.

PR #562, opened 2024-09-26
policy, attestations: Add Searcher interfaces

adityasaky:
This commit adds Searcher interfaces for the policy and attestations refs, which are used by VerifyRelativeForRef.


adityasaky:
Part of #531

adityasaky:
Maybe? I'm leaning towards starting with specific searchers after which we can decide if we can generalize them

Issue #266, opened 2024-01-12
Remove vendored go-securesystemslib


adityasaky:
### Add a description

go-securesystemslib is vendored in #264. We must remove this copy of the library once our changes are upstreamed or our requirements are met through other APIs. Note that an effort has started in the in-toto community to make go-securesystemslib more robust, which is the reason we haven't already upstreamed our changes.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adityasaky:
FWIW, I'm a maintainer of go-securesystemslib as well, so I'm going to be involved in the go-securesystemslib improvement effort. I will take this issue on when the library is ready for direct use in gittuf.

PR #563, opened 2024-09-26
Vendor go-sslib/dsse and add support for setting a signature extension

adityasaky:
This PR prepares #524 by vendoring the dsse library and adding support for signature extensions.


adityasaky:
Updating #266 to track this

PR #566, opened 2024-09-27
Use signerverifier/ssh for git signing / verifying

### LINKED ISSUE #448: Drop `verifySSHKeySignature` in favour of `ssh.Verifier.Verify`
Is the plan to also remove `verifySSHKeySignature` in favour of ssh.Verifier.Verify?

_Originally posted by @lukpueh in https://github.com/gittuf/gittuf/pull/436#discussion_r1639672959_

adityasaky:
Closes #448 

I ran into an import cycle when I was trying something and decided to change this. Signing flow needs changing still, and possibly careful testing to ensure we're compatible.

Edit: I don't know if we want to change the signing flow: we're working off key bytes and it's used more for tests at the moment.

### DIFF
--- a/internal/signerverifier/signerverifier.go
+++ b/internal/signerverifier/signerverifier.go
@@ -6,19 +6,14 @@ package signerverifier
 import (
 	"github.com/gittuf/gittuf/internal/signerverifier/common"
 	"github.com/gittuf/gittuf/internal/signerverifier/ssh"
-	sslibsv "github.com/gittuf/gittuf/internal/third_party/go-securesystemslib/signerverifier"
 	"github.com/gittuf/gittuf/internal/tuf"
 	"github.com/secure-systems-lab/go-securesystemslib/dsse"
 )
 
 const (
-	ED25519KeyType  = sslibsv.ED25519KeyType
-	ECDSAKeyType    = sslibsv.ECDSAKeyType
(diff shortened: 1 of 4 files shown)


internal/signerverifier/signerverifier.go
```diff
@@ -6,33 +6,22 @@ package signerverifier
...
 // Deprecated: Switch to upstream key loading APIs.
 func NewSignerVerifierFromTUFKey(key *tuf.Key) (dsse.Verifier, error) {
-	switch key.KeyType {
-	case ED25519KeyType:
-		return sslibsv.NewED25519SignerVerifierFromSSLibKey(key)
-	case ECDSAKeyType:
-		return sslibsv.NewECDSASignerVerifierFromSSLibKey(key)
-	case RSAKeyType:
-		return sslibsv.NewRSAPSSSignerVerifierFromSSLibKey(key)
+	switch key.KeyType { //nolint:gocritic
```
adityasaky, line 24: This will be reworked when I rebase #588

adityasaky, line 24: This will be reworked when I rebase #588

adityasaky: I noticed this is only really needed in the policy package and likely doesn't need to be imported from another package because we've done a lot of consolidation around the verification flows!

PR #570, opened 2024-09-30
transport: Fix curl pushes on Windows

adityasaky:
counterpart to #569 on the push side of things.

### DIFF
--- a/internal/git-remote-gittuf/curl.go
+++ b/internal/git-remote-gittuf/curl.go
@@ -387,34 +387,48 @@ func handleCurl(repo *repository.Repository, remoteName, url string) (map[string
 				return nil, false, err
 			}
 
-			helperStdOutScanner := bufio.NewScanner(helperStdOut)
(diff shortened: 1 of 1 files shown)


internal/git-remote-gittuf/curl.go
```diff
@@ -387,34 +387,48 @@ func handleCurl(repo *repository.Repository, remoteName, url string) (map[string
...
+					}
+
+					if bytes.Equal(output, []byte("\n")) {
+						seenTrailingNewLine = true
+						break
 					}
 				}
 
-				// flushPkt indicates end of message
-				if bytes.Equal(output, flushPkt) {
```
adityasaky, line 417: interestingly, the flushPkt check is not required because git-remote-http itself doesn't give us a flushPkt. We were still terminating correctly because on non-windows environments, we were getting that last newline without having to wrap the scan like we are in this PR.

PR #572, opened 2024-10-01
cmd: Drop devmode requirement for root and policy thresholds

patzielinski:
This PR removes the devmode requirement for updating the thresholds for the root of trust and policy.


adityasaky (approved):
Can you double check we don't have a secondary check in the repository package for these flows?

Issue #554, opened 2024-09-24
Reference authorizations don't cleanly support tags


adityasaky:
### What happened?

With the expectation that a reference authorization is for merging commits, we don't currently support tags very well. I propose we bump the predicate to v0.2 and update how we apply its contents.

When the target ref is a tags ref: from -> zero/empty, target -> underlying commit the tag points to

When the target ref is a non-tags ref: from -> previous commit for the ref / zero (if approving for a new ref), target -> underlying Git tree

While refs / tags can point to any Git objects, I think we're safe in that attestations are likely to be used for:
a) branches to move from one commit to another
b) tag refs that point to tag objects that point to commit objects

We could be flexible for what the tag object points to, possibly, but I think on the first, we should expect commits (and therefore use trees for the target in the attestation).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adityasaky:
I don't know if we support lightweight tags today fwiw. I think we expect tag objects at several points for `refs/tags` in our verification flow.

PR #575, opened 2024-10-01
policy: Fix attestation load for tag verification

adityasaky:
Currently, attestations are loaded assuming the approval target is a Git tree. Entries for tags don't have trees. Separately, we can add support for approval attestations for tags that know not to compute merge trees when approving a tag, instead using the commit ID the tag is expected to point to.

I included this patch in #556 which does add the support for approving a tag using expected commit ID as the `to`. Testing for this case requires 556 (we need a new version of the reference authorization).

Partially addresses #554 as verification will now work correctly.


adityasaky:
How on earth did coverage increase? https://coveralls.io/builds/70111120

PR #583, opened 2024-10-04
*: Remove legacy key types

adityasaky:
This commit drops support for the legacy / custom securesystemslib key format. This format was used in two forms:
a) In tests
b) On disk in the policy state for the root keys

To address the removal, tests have been updated to use SSH keys (via the ssh-keygen signer). This makes up the majority of the diff for this commit, and includes some additions to the ssh package to more easily load test artifacts.

Additionally, we don't need to store a policy state's root keys on disk for that ref. This was an error in our initial design, and it actually leads to complications in ensuring that the policy state's on disk keys match the keys listed in the state's root metadata. This commit updates it so only the root metadata's record of the root keys are used, with the keys directory omitted for future policy states. However, we maintain backwards compatibility for policy states that include the keys on disk, we just ignore them.


adityasaky:
We have something flaky with ssh-keygen 🤔

adityasaky:
note: drop in coverage in ssh is a test-only func we're exposing for use elsewhere

PR #614, opened 2024-10-21
docs: Improvements to the design document

adityasaky:
This PR also moves some WIP sections to the extensions directory,


JustinCappos (approved):
I took a quick look and while I don't have all the needed context to comment deeply on the examples, data formats, etc. it seems sensible to me.

adityasaky:
> WDYT about adding some figures (either from the paper or made specifically for this doc)?

I suggest we add figures from the paper once that's live, if we're allowed to.

PR #615, opened 2024-10-21
docs: Add app / code review tool attestations

adityasaky:
Adds app / code review tool attestations as an extension doc after some discussions with @wlynch re https://github.com/gittuf/github-app.

### DIFF
--- a/docs/gaps/5/README.md
+++ b/docs/gaps/5/README.md
@@ -1,0 +1,233 @@
+# Principals, not Keys
+
+## Metadata
+
+* **Number:** 5
+* **Title:** Principals, not Keys
+* **Implemented:** No
+* **Withdrawn/Rejected:** No
+* **Sponsors:** Aditya Sirish A Yelgundhalli (adityasaky)
+* **Last Modified:** March 25, 2025
+
+## Abstract
+
+gittuf's policy, like TUF metadata that it is inspired by, assumes that all
+actors are [represented by their signing keys](/docs/design-document.md#actors).
+This assumption leads to usability problems in Git development in contrast to
+traditional TUF deployments in packaging ecosystems and the like.
+
+## Motivation
+
+There are more contexts where a developer may sign in a Git repository protected
+by gittuf compared to a standard TUF deployment. Developers may contribute code
+from different devices or cloud based IDEs that have distinct signing keys. The
+current gittuf policy would represent these different signing keys as
+independent actors, making policy misconfigurations that allow a developer to
+meet a threshold using multiple keys they control more likely.
+
+Additionally, as gittuf may integrate multiple systems (a code forge like
+GitHub, a code review tool like Gerrit), the policy ought to be able to track a
+developer's identity across all of these systems. Doing so would reduce
+cognitive overload for maintainers who manage policies, and would enable
+integrating into metadata such as in-toto attestations produced by these other
+systems.
+
+## Specification
+
+This GAP introduces the notion of a generic "principal". A principal can be a
+single key or a more complex entity that contains more than one signing key and
+other custom metadata. A principal has the following attributes:
+* ID: A unique identifier for the principal
+* Keys: A set of public keys that can be used to verify signatures, no
+  duplicates are allowed
+* CustomMetadata: Additional custom metadata that must be associated with the
+  principal
+
+Additionally, this GAP introduces the first new principal type, to represent a
+"person". A "person" has the following schema:
+
+```
+"<id>": {
+    "personID": "<id>",
+    "publicKeys": {
+        "<keyID1>": {
+            "keyid": "<keyID1>",
+            "keyid_hash_algorithms": null,
+            "keytype": "sigstore-oidc",
+            "keyval": {
+                "identity": "<username>",
+                "issuer": "<issuer>"
+            },
+            "scheme": "fulcio"
+          },
+        "<keyid2>>": ...
+    },
+    "associatedIdentities": {
+        "<integrationID1>": "<usernameA>",
+        "<integrationID2>": "<usernameB>",
+        ...
+    }
+    "customMetadata": {...}
+}
+```
+
+Here, `personID` can be set to any unique (within the policy) identifier of a
+person such as their name or email address, similar to
+[TAP-12](https://github.com/theupdateframework/taps/blob/master/tap12.md). Each
+key associated with a person follows the schema already used by keys in gittuf's
+policy today. `associatedIdentities` records the person's identity on
+integrations. Finally, `customMetadata` allows for recording arbitrary
+attributes with each person for further extensibility. To meet the generic
+principal schema, the person type considers a combination of `customMetadata`
+and `associatedIdentities` as the principal's custom metadata.
+
+Here is an example person using this schema:
+
+```
+"aditya@example.com": {
+    "personID": "aditya@example.com",
+    "publicKeys": {
+        "B83110D012545604": {
+            "keyid": "B83110D012545604",
+            "keytype": "gpg",
+            "keyval": {
+                "public": "<pubkey>"
+            },
+            "scheme": "gpg"
+        },
+        "aditya@example.com::https://github.com/login/oauth": {
+            "keyid": "aditya@example.com::https://github.com/login/oauth",
+            "keytype": "sigstore-oidc",
+            "keyval": {
+                "identity": "aditya@example.com",
+                "issuer": "https://github.com/login/oauth"
+            },
+            "scheme": "fulcio"
+        },
+    },
+    "associatedIdentities": {
+        "https://github.com": "adityasaky+8928778",
+    }
+    "customMetadata": {}
+}
+```
+
+Each associated identity SHOULD be recorded with an immutable value. For
+example, some source control platforms allow a user to change their username,
+with the old username up for grabs for another user to claim. Here, the username
+alone is insufficient to identify the user uniquely. In the example above,
+`8928778` is a unique identifier on github.com that does not change even if the
+username is updated. For readability, the associated identity includes both the
+usename and the unchanging ID, with only the latter ensuring uniqueness.
+
+### Updating Principal Definitions
+
+When dealing directly with keys, a delegating entity must update their
+delegation when a developer must rotate their key. With principals, this remains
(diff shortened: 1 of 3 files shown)


docs/extensions/code-review-tool-attestations.md
```diff
@@ -0,0 +1,114 @@
...
+gittuf policy must be updated to extend the information tracked for each trusted
+person. Currently, each trusted person is represented directly by their public
+key in policy metadata, similar to how TUF represents actors. To support code
+review tool attestations, this must be extended to support the identity of an
+approver from the perspective of the code review tool. For example, Alice may
+use an SSH key for commits and RSL entries but use GitHub's cloud offering for
+code review. Thus, gittuf policy must associate her public key and her GitHub
+username / identifier. Also see: https://github.com/gittuf/gittuf/issues/586.
+
+This extension proposes the following format for the policy schema:
```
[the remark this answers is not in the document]
adityasaky: I think I get what you're getting at, but could you rephrase to be certain?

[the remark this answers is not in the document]
adityasaky: Good points, and I'll raise you one!

Right now, I'm focusing on allowing "alice" to sign using any of her keys. I think we could explore parametrizing it in future, like `alice:<keyid>` or something, but I'm unsure how frequently this is going to be necessary. I also see no reason we don't continue supporting `Key` as a type where this is a concern. But overall, I think the point you raise is an open question: do we support setting constraints on a person as to what keys they ought to use in a particular context? What about thresholds? Right now, it only counts towards the threshold for the rule but perhaps in a context, Alice can only count as having signed if she uses _two of her keys_ rather than just one.

The other thing I was considering: consistency in Person definitions. Is a `Person` redefined in multiple rule files where they're delegated to? How do we keep them consistent across definitions?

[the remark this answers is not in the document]
adityasaky: This is really a two-parter:
a) Storing metadata alongside a person definition pertaining to their github etc identities
b) Using those identities to verify code review attestations from SCPs

docs/extensions/code-review-tool-attestations.md
```diff
@@ -0,0 +1,114 @@
...
+## Attestation Schema
+
+The attestation is similar to the reference authorization, but also includes
+information indicating the approvers.
+
+```
+TargetRef    string
+FromTargetID string
+ToTargetID   string
+Approvers    []string
```
wlynch, line 22: Worth adding comments + examples for each.

docs/extensions/code-review-tool-attestations.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+The attestation must indicate the code review tool via the predicate type, and
```
wlynch, line 25: > via the predicate type

Where should predicate type be specified?

docs/extensions/code-review-tool-attestations.md
```diff
@@ -0,0 +1,114 @@
...
+review tool attestations, this must be extended to support the identity of an
+approver from the perspective of the code review tool. For example, Alice may
+use an SSH key for commits and RSL entries but use GitHub's cloud offering for
+code review. Thus, gittuf policy must associate her public key and her GitHub
+username / identifier. Also see: https://github.com/gittuf/gittuf/issues/586.
+
+This extension proposes the following format for the policy schema:
+
+```
+"<id>": {
```
wlynch, line 44: what would be the primarily id here?

docs/extensions/code-review-tool-attestations.md
```diff
@@ -0,0 +1,114 @@
...
+            "keytype": "sigstore-oidc",
+            "keyval": {
+                "identity": "<username>",
+                "issuer": "<issuer>"
+            },
+            "scheme": "fulcio"
+          },
+        "<keyid2>>": ...
+    },
+    "associatedAppIdentities": {
```
wlynch, line 59: bikeshed:

```suggestion
    "associatedIdentities": {
```

I don't think we need to over-index on "app" terminology here. In my mind, technically a human should also be able to make this kind of assertion if they wanted.

docs/extensions/code-review-tool-attestations.md
```diff
@@ -0,0 +1,114 @@
...
+        "<appID1>": "<usernameA>",
```
wlynch, line 60: I know we also talked about there would be information needed to describe the appID's trust - e.g. the keys it uses, etc. what do we need and where would it be specified?

docs/extensions/code-review-tool-attestations.md
```diff
@@ -0,0 +1,114 @@
...
+during verification, but this has usability concerns. To balance usability and
+security, this extension proposes the use of special attestations created by
+popular code review tools.
+
+## Attestation Schema
+
+The attestation is similar to the reference authorization, but also includes
+information indicating the approvers.
+
+```
```
wlynch, line 18: Would the diff codeblock type be useful here to show what's new?

docs/extensions/code-review-tool-attestations.md
adityasaky: That's in the attestation, it's a required field in the in-toto statement.

docs/extensions/code-review-tool-attestations.md
adityasaky: I wanted to leave it up to the users, it could be anything they want? Right now we use keyids but I don't see why we can't have it be usernames or other identifiers based on the context.

docs/extensions/code-review-tool-attestations.md
adityasaky: I think it's just the signer, since all we really have is a signed attestation. The public key is in the root.json file. I think the app ID could be any human picked name again? The root.json would have `appID: <principal>` mappings effectively. `principal` is what I'm calling the interface right now that could be a person, a single pubkey, etc.

docs/extensions/code-review-tool-attestations.md
adityasaky: This is a new predicate type altogether rather than an extension of the reference authorization type. I think that makes it easier to ensure we have no confusion in how these attestations are used.

adityasaky:
#626 is a first pass at supporting this in the policy metadata, we could use that to discuss the changes proposed here.

docs/gaps/5/README.md
```diff
@@ -0,0 +1,233 @@
...
+        "<integrationID2>": "<usernameB>",
+        ...
+    }
+    "customMetadata": {...}
+}
+```
+
+Here, `personID` can be set to any unique (within the policy) identifier of a
+person such as their name or email address, similar to
+[TAP-12](https://github.com/theupdateframework/taps/blob/master/tap12.md).  Each
```
patzielinski, line 76:
```suggestion
[TAP-12](https://github.com/theupdateframework/taps/blob/master/tap12.md). Each
```

docs/gaps/5/README.md
```diff
@@ -0,0 +1,233 @@
...
+        },
+    },
+    "associatedIdentities": {
+        "https://github.com": "adityasaky+8928778",
+    }
+    "customMetadata": {}
+}
+```
+
+Each associated identity SHOULD be recorded with an immutable value. For
```
patzielinski, line 115: Do we follow any special meaning for capitalized words like SHOULD, e.g. https://www.rfc-editor.org/rfc/rfc2119?

docs/gaps/5/README.md
```diff
@@ -0,0 +1,233 @@
...
+single key or a more complex entity that contains more than one signing key and
+other custom metadata. A principal has the following attributes:
+* ID: A unique identifier for the principal
+* Keys: A set of public keys that can be used to verify signatures, no
+  duplicates are allowed
+* CustomMetadata: Additional custom metadata that must be associated with the
+  principal
+
+Additionally, this GAP introduces the first new principal type, to represent a
+"person".  A "person" has the following schema:
```
patzielinski, line 47:
```suggestion
"person". A "person" has the following schema:
```

docs/gaps/5/README.md
```diff
@@ -0,0 +1,233 @@
...
+existing policies that use keys directly.
+
+## Security
+
+This GAP introduces some concepts that have security implications.
+
+### Out of Sync Principal Definitions
+
+The delegating entity is responsible for associating a unique identifier for the
+delegatee "person".  While this is unique within the context of a delegated
```
patzielinski, line 184:
```suggestion
delegatee "person". While this is unique within the context of a delegated
```

docs/gaps/5/README.md
```diff
@@ -0,0 +1,233 @@
...
+file, other delegated files may use the same ID for another person. To avoid
+policy misconfiguration or confusion issues, the person definitions SHOULD be
+consistent across metadata.  In other words, the definition for `Bob` SHOULD be
```
patzielinski, line 187:
```suggestion
consistent across metadata. In other words, the definition for `Bob` SHOULD be
```

patzielinski (approved):
I think we can merge this and iterate on it with updates as needed.

docs/gaps/6/README.md
```diff
@@ -0,0 +1,162 @@
...
+Then, the attestations's `Approvers` are used to extend the list of `personIDs`
+verified via their signatures. Each approver is matched to a person using gittuf
+policy, and the corresponding `personID` is added to `personIDs`. Note that
+`personIDs` must contain no duplicates, to ensure the same person is not counted
+twice towards the threshold. If the threshold is met with the approvers in the
+code review tool attestation, then verification passes.
+
+## Reasoning
+
+TODO: flesh out reasoning.
```
wlynch, line 123: Do we want to wait for this?

wlynch (commented):
Generally LGTM!

wlynch (approved):

docs/gaps/5/README.md
adityasaky: It's based on that, but if so, the RFC should be listed in the GAP process itself. I don't want us to put too fine a point on it.

docs/gaps/6/README.md
adityasaky: added!

PR #629, opened 2024-10-30
debug: Add Dockerfile for debugging gittuf on Git 2.34.1

adityasaky:
This should make life easier when debugging on Git 2.34.1.


adityasaky:
I think the current setup is fine.

PR #632, opened 2024-10-30
gittuf: Improve testing coverage

yongjae354:
cc @patzielinski

### DIFF
--- a/experimental/gittuf/repository_test.go
+++ b/experimental/gittuf/repository_test.go
@@ -13,21 +13,38 @@ import (
 )
 
 func TestLoadRepository(t *testing.T) {
-	tmpDir := t.TempDir()
-	gitinterface.CreateTestGitRepository(t, tmpDir, false)
+	t.Run("load with no repo, unsuccessful", func(t *testing.T) {
+		tmpDir := t.TempDir()
+		currentDir, err := os.Getwd()
+		if err != nil {
+			t.Fatal(err)
+		}
+		if err := os.Chdir(tmpDir); err != nil {
+			t.Fatal(err)
+		}
+		defer os.Chdir(currentDir) //nolint:errcheck
 
-	currentDir, err := os.Getwd()
-	if err != nil {
-		t.Fatal(err)
-	}
-	if err := os.Chdir(tmpDir); err != nil {
-		t.Fatal(err)
-	}
-	defer os.Chdir(currentDir) //nolint:errcheck
+		repository, err := LoadRepository()
+		assert.NotNil(t, err)
+		assert.Nil(t, repository)
+	})
 
-	repository, err := LoadRepository()
-	assert.Nil(t, err)
-	assert.NotNil(t, repository.r)
+	t.Run("successful load", func(t *testing.T) {
+		tmpDir := t.TempDir()
+		currentDir, err := os.Getwd()
+		if err != nil {
+			t.Fatal(err)
+		}
+		if err := os.Chdir(tmpDir); err != nil {
+			t.Fatal(err)
+		}
+		defer os.Chdir(currentDir) //nolint:errcheck
+
+		gitinterface.CreateTestGitRepository(t, tmpDir, false)
+		repository, err := LoadRepository()
+		assert.Nil(t, err)
+		assert.NotNil(t, repository.r)
+	})
 }
 
 func TestUnauthorizedKey(t *testing.T) {
@@ -53,4 +70,13 @@ func TestUnauthorizedKey(t *testing.T) {
 		err := r.RemoveTopLevelTargetsKey(testCtx, targetsSigner, "some key ID", false)
 		assert.ErrorIs(t, err, ErrUnauthorizedKey)
 	})
+}
+
+func TestGetGitRepository(t *testing.T) {
+	tmpDir := t.TempDir()
+	repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
+
+	r := &Repository{r: repo}
+	result := r.GetGitRepository()
+	assert.Equal(t, repo, result)
 }


experimental/gittuf/repository_test.go
```diff
@@ -54,3 +54,29 @@ func TestUnauthorizedKey(t *testing.T) {
...
+	tmpDir := t.TempDir()
+	repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
+
+	r := &Repository{r: repo}
+	result := r.GetGitRepository()
+
+	assert.Equal(t, repo, result, "Error: GetGitRepository does not return repo")
+}
+
+func TestLoadRepository_NoRepo(t *testing.T) {
```
adityasaky, line 68: you could make this just `TestLoadRepository` and have a `t.Run` for a no repo failure case. Also a `t.Run` for a successful load.

adityasaky: Ah `TestLoadRepository exists, I'd move everything in that into a `t.Run` for the successful load case and add this as a `t.Run` for the expected failure case.

experimental/gittuf/repository_test.go
```diff
@@ -54,3 +54,29 @@ func TestUnauthorizedKey(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := os.Chdir(tmpDir); err != nil {
+		t.Fatal(err)
+	}
+	defer os.Chdir(currentDir) //nolint:errcheck
+
+	repository, err := LoadRepository() // Should err
+	assert.NotNil(t, err)
```
adityasaky, line 80: Do we have a named error we can use here?

experimental/gittuf/policy_test.go
```diff
@@ -114,3 +114,18 @@ func TestPullPolicy(t *testing.T) {
...
+func TestApplyPolicy(t *testing.T) {
+	localRepo := createTestRepositoryWithPolicy(t, "")
+	err := localRepo.ApplyPolicy(testCtx, false)
+	assert.Nil(t, err)
+}
+
+func TestListRules(t *testing.T) {
+	localRepo := createTestRepositoryWithPolicy(t, "")
+
+	_, err := localRepo.ListRules(testCtx, policy.PolicyRef)
```
adityasaky, line 127: this should assert the rules returned as well...

experimental/gittuf/policy_test.go
```diff
@@ -114,3 +114,18 @@ func TestPullPolicy(t *testing.T) {
 		assert.ErrorIs(t, err, ErrPullingPolicy)
 	})
 }
+
+func TestApplyPolicy(t *testing.T) {
+	localRepo := createTestRepositoryWithPolicy(t, "")
```
adityasaky, line 119:
```suggestion
	repo := createTestRepositoryWithPolicy(t, "")
```

local, remote prefix makes sense for the synchronization tests.

experimental/gittuf/policy_test.go
```diff
@@ -114,3 +114,18 @@ func TestPullPolicy(t *testing.T) {
...
+	err := localRepo.ApplyPolicy(testCtx, false)
```
adityasaky, line 120: isn't the policy already applied by the helper?

experimental/gittuf/repository_test.go
```diff
@@ -25,9 +23,18 @@ func TestLoadRepository(t *testing.T) {
 	}
 	defer os.Chdir(currentDir) //nolint:errcheck
 
-	repository, err := LoadRepository()
-	assert.Nil(t, err)
-	assert.NotNil(t, repository.r)
+	t.Run("Load with no repo, unsuccessful", func(t *testing.T) {
```
adityasaky, line 26: I would inline the tmpDir declaration too so that each t.Run has its own tmpDir where a repository is created (or not) with corresponding chdir calls as well. This makes the tests less fragile / dependent on order of execution.

adityasaky, line 26:
```suggestion
	t.Run("load with no repo, unsuccessful", func(t *testing.T) {
```

experimental/gittuf/repository_test.go
```diff
@@ -25,9 +23,18 @@ func TestLoadRepository(t *testing.T) {
...
+		repository, err := LoadRepository() // Should err
+		assert.NotNil(t, err)
+		assert.Nil(t, repository)
+	})
+
+	t.Run("Successful load", func(t *testing.T) {
```
adityasaky, line 32:
```suggestion
	t.Run("successful load", func(t *testing.T) {
```

experimental/gittuf/repository_test.go
```diff
@@ -54,3 +61,13 @@ func TestUnauthorizedKey(t *testing.T) {
...
 }
+
+func TestGetGitRepository(t *testing.T) {
+	tmpDir := t.TempDir()
+	repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
+
+	r := &Repository{r: repo}
+	result := r.GetGitRepository()
+
+	assert.Equal(t, repo, result, "Error: GetGitRepository does not return repo")
```
adityasaky, line 72:
```suggestion
	assert.Equal(t, repo, result)
```

experimental/gittuf/repository_test.go
```diff
@@ -54,3 +61,13 @@ func TestUnauthorizedKey(t *testing.T) {
...
 	})
 }
+
+func TestGetGitRepository(t *testing.T) {
+	tmpDir := t.TempDir()
+	repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
+
+	r := &Repository{r: repo}
+	result := r.GetGitRepository()
+
```
adityasaky, line 71:
```suggestion
```

adityasaky (commented):
good stuff, some comments!

experimental/gittuf/policy_test.go
yongjae354: Thank you for your feedback. Yes, it seems like `ApplyPolicy` simply makes a call to `policy.Apply` which is already done inside `createTestRepositoryWithPolicy`. 
Do you think I could change the `createTestRepositoryWithPolicy` function in helpers_test.go to call the `ApplyPolicy` function instead, for coverage?

adityasaky: I think it might be worth moving it out of this PR.

experimental/gittuf/repository_test.go
```diff
@@ -13,21 +13,38 @@ import (
 )
 
 func TestLoadRepository(t *testing.T) {
-	tmpDir := t.TempDir()
-	gitinterface.CreateTestGitRepository(t, tmpDir, false)
+	t.Run("load with no repo, unsuccessful", func(t *testing.T) {
```
adityasaky, line 16: the tests in this file are fine, I suggest we merge this PR with just those and circle back on the policy_test.go ones?

adityasaky (approved):

PR #636, opened 2024-10-31
*: Move signing checks around

### LINKED ISSUE #633: Move the signing key check into the `gittuf` package
To have a unified way to check if signing is possible (for gittuf, the transport, and command compatibility layer), the signing check currently in [`cmd`](https://github.com/gittuf/gittuf/blob/main/internal/cmd/common/common.go#L48) should be moved into the public `gittuf` package.

patzielinski:
This PR moves the `CheckIfSigningViable` check to the gittuf API and updates the signing key argument check as well.

Closes #633

### DIFF
--- a/experimental/gittuf/attestations.go
+++ b/experimental/gittuf/attestations.go
@@ -47,6 +47,14 @@ func (r *Repository) AddReferenceAuthorization(ctx context.Context, signer sslib
 		return dev.ErrNotInDevMode
 	}
 
+	if signCommit {
+		slog.Debug("Checking if Git signing is configured...")
+		err := r.r.CanSign()
+		if err != nil {
+			return err
+		}
+	}
+
 	var err error
 
 	targetRef, err = r.r.AbsoluteReference(targetRef)
@@ -173,6 +181,14 @@ func (r *Repository) RemoveReferenceAuthorization(ctx context.Context, signer ss
 		return dev.ErrNotInDevMode
 	}
 
+	if signCommit {
+		slog.Debug("Checking if Git signing is configured...")
+		err := r.r.CanSign()
+		if err != nil {
+			return err
+		}
+	}
+
 	// Ensure only the key that created a reference authorization can remove it
 	slog.Debug("Evaluating if key can sign...")
 	_, err := signer.Sign(ctx, nil)
@@ -245,10 +261,19 @@ func (r *Repository) AddGitHubPullRequestAttestationForCommit(ctx context.Contex
 		return dev.ErrNotInDevMode
 	}
 
+	if signCommit {
+		slog.Debug("Checking if Git signing is configured...")
+		err := r.r.CanSign()
+		if err != nil {
+			return err
+		}
+	}
+
 	options := githubopts.DefaultOptions
 	for _, fn := range opts {
 		fn(options)
 	}
+
 	if options.GitHubToken == "" {
 		options.GitHubToken = os.Getenv(githubTokenEnvKey)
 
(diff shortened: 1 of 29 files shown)


experimental/gittuf/common.go
```diff
@@ -0,0 +1,17 @@
+// Copyright The gittuf Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package gittuf
+
+import "github.com/gittuf/gittuf/internal/gitinterface"
+
+// CheckIfSigningViable checks if we are able to sign RSL entries given the
+// current environment
+func CheckIfSigningViable() error {
```
adityasaky, line 10:
```suggestion
func CheckIfGitSigningEnabled() error {
```

experimental/gittuf/common.go
```diff
@@ -0,0 +1,17 @@
+// Copyright The gittuf Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package gittuf
+
+import "github.com/gittuf/gittuf/internal/gitinterface"
+
+// CheckIfSigningViable checks if we are able to sign RSL entries given the
```
adityasaky, line 8:
```suggestion
// CheckIfSigningViable checks if we are able to sign Git commits given the
```

experimental/gittuf/common.go
```diff
@@ -0,0 +1,17 @@
...
+// current environment
```
adityasaky, line 9:
```suggestion
// current configuration.
```

experimental/gittuf/common.go
adityasaky: Should this be a method on Repository in this package?

internal/cmd/common/common.go
```diff
@@ -40,16 +40,11 @@ func CheckIfSigningViableWithFlag(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("required flag \"signing-key\" not set")
 	}
 
-	return CheckIfSigningViable(cmd, nil)
+	return CheckIfSigningViable(nil, nil)
```
adityasaky, line 43: I'd keep this in in case this ever changes. It's going to be dropped by the receiver anyway.

experimental/gittuf/common.go
adityasaky: I meant should it be a method on `gittuf.Repository`, not `gitinterface.Repository`.

adityasaky: Using `r.r.CanSign` makes a lot of sense fwiw.

adityasaky (commented):
So I see this is replacing the pre run check but don't we still have the original problem when accessing directly with the API?

experimental/gittuf/attestations.go
```diff
@@ -49,6 +49,14 @@ func (r *Repository) AddReferenceAuthorization(ctx context.Context, signer sslib
 
 	var err error
 
+	if signCommit {
+		slog.Debug("Checking if git signing is configured...")
```
adityasaky, line 53:
```suggestion
		slog.Debug("Checking if Git signing is configured...")
```

experimental/gittuf/root.go
```diff
@@ -28,6 +28,15 @@ func (r *Repository) InitializeRoot(ctx context.Context, signer sslibdsse.Signer
 		publicKeyRaw *signerverifier.SSLibKey
 		err          error
 	)
+
+	if signCommit {
```
adityasaky, line 32: move this above the var?

experimental/gittuf/targets.go
```diff
@@ -259,6 +291,14 @@ func (r *Repository) RemoveDelegation(ctx context.Context, signer sslibdsse.Sign
 		return err
 	}
 
+	if signCommit {
+		slog.Debug("Checking if git signing is configured...")
```
adityasaky, line 295: move this above signer.KeyID? in fulcio, this would trigger an oidc flow to get the keyID.

adityasaky (approved):

Issue #557, opened 2024-09-25
Add tests for currently-unused combinations in GetLatestReferenceEntry


patzielinski:
#532 introduces a unified `GetLatestReferenceEntry` which takes in options instead of having multiple methods for the getting the latest reference entry.

Tests are present for the combinations currently used, but we need to figure out the combinations / options that are not tested, and add tests accordingly.

Originally raised here: https://github.com/gittuf/gittuf/pull/532#discussion_r1771814769

jasisboredsojoinedgithub:
Hi, @patzielinski Can I work on this? I’d like to add the missing tests. Thanks!

adityasaky:
@jasisboredsojoinedgithub go for it!

patzielinski:
@jasisboredsojoinedgithub quick ping, are you still working on this?

PR #650, opened 2024-11-08
add two new unit test cases

rishabhBudhouliya:
Linked to: https://github.com/gittuf/gittuf/issues/557
This PR is part of a larger effort to improve the coverage of the codebase

Before coverage `internal/rsl` package %age: 83.2%
After coverage `internal/rsl` %age: 83.9%
Attaching the coverage report for reference as a zip(apparently Github doesn't allow .out format attachments): 
[finalcoverage.out.zip](https://github.com/user-attachments/files/17684214/finalcoverage.out.zip)

### DIFF
--- a/internal/rsl/rsl_test.go
+++ b/internal/rsl/rsl_test.go
@@ -178,6 +178,15 @@ func TestGetLatestReferenceEntry(t *testing.T) {
 		}
 		_, _, err = GetLatestReferenceEntry(repo, UntilEntryNumber(10))
 		assert.ErrorIs(t, err, ErrInvalidUntilEntryNumberCondition)
+
+		// a reference older than until entry number is being asked for (configuration mismatch)
+		if err := NewReferenceEntry("feature", gitinterface.ZeroHash).Commit(repo, false); err != nil {
+			t.Fatal(err)
+		}
+		entry, annotations, err := GetLatestReferenceEntry(repo, ForReference("main"), BeforeEntryID(emptyTreeID), UntilEntryNumber(2))
+		assert.Nil(t, annotations)
+		assert.Nil(t, entry)
+		assert.ErrorIs(t, err, ErrInvalidGetLatestReferenceEntryOptions)
 	})
 
 	t.Run("with ref name and until entry number", func(t *testing.T) {
@@ -392,6 +401,28 @@ func TestGetLatestReferenceEntry(t *testing.T) {
 		assert.Equal(t, entryIDs[2], entry.ID)
 
 		_, _, err = GetLatestReferenceEntry(repo, ForReference("feature"), BeforeEntryID(entryIDs[3]), UntilEntryNumber(3))
+		assert.ErrorIs(t, err, ErrRSLEntryNotFound)
+	})
+
+	t.Run("with ref name but before entry ID is not found in initial walk", func(t *testing.T) {
+		tempDir := t.TempDir()
+		repo := gitinterface.CreateTestGitRepository(t, tempDir, false)
+		// RSL structure for the test
+		// 1    <- 2
+		// main <- feature
+		testRefs := []string{"main", "feature"}
+		for _, ref := range testRefs {
+			if err := NewReferenceEntry(ref, gitinterface.ZeroHash).Commit(repo, false); err != nil {
+				t.Fatal(err)
+			}
+		}
+		emptyTreeID, err := repo.EmptyTree()
+		if err != nil {
+			t.Fatal(err)
+		}
+		entry, annotations, err := GetLatestReferenceEntry(repo, ForReference("main"), BeforeEntryID(emptyTreeID))
+		assert.Nil(t, annotations)
+		assert.Nil(t, entry)
 		assert.ErrorIs(t, err, ErrRSLEntryNotFound)
 	})
 


internal/rsl/rsl_test.go
```diff
@@ -395,6 +395,50 @@ func TestGetLatestReferenceEntry(t *testing.T) {
...
+		// RSL structure for the test
+		// 1    <- 2
+		// main <- feature
+		testRefs := []string{"main", "feature"}
+		for _, ref := range testRefs {
+			if err := NewReferenceEntry(ref, gitinterface.ZeroHash).Commit(repo, false); err != nil {
+				t.Fatal(err)
+			}
+		}
+		hash, hashErr := gitinterface.NewHash("a94a8fe5ccb19ba61c4c0873d391e987982fbbd3")
```
adityasaky, line 410: it's better to handle the err before moving on. Reuse `err` here and make it:

```
hash, err := gitinterface.NewHash("a94a8fe5ccb19ba61c4c0873d391e987982fbbd3")
if err != nil {
    t.Fatal(err)
}

entry, annotations, err := GetLatestReferenceEntry(...)
...
```

internal/rsl/rsl_test.go
```diff
@@ -395,6 +395,50 @@ func TestGetLatestReferenceEntry(t *testing.T) {
...
+			if err := NewReferenceEntry(ref, gitinterface.ZeroHash).Commit(repo, false); err != nil {
+				t.Fatal(err)
+			}
+			latest, err := GetLatestEntry(repo)
+			if err != nil {
+				t.Fatal(err)
+			}
+			entryIDs = append(entryIDs, latest.GetID())
+		}
+		entry, annotations, err := GetLatestReferenceEntry(repo, ForReference("main"), BeforeEntryID(entryIDs[0]), UntilEntryNumber(2))
```
adityasaky, line 436:
```suggestion

		entry, annotations, err := GetLatestReferenceEntry(repo, ForReference("main"), BeforeEntryID(entryIDs[0]), UntilEntryNumber(2))
```

internal/rsl/rsl_test.go
```diff
@@ -395,6 +395,50 @@ func TestGetLatestReferenceEntry(t *testing.T) {
...
+		assert.Empty(t, annotations)
```
adityasaky, line 437:
```suggestion
		assert.Nil(t, annotations)
```

internal/rsl/rsl_test.go
```diff
@@ -395,6 +395,50 @@ func TestGetLatestReferenceEntry(t *testing.T) {
...
+		assert.Empty(t, entry)
```
adityasaky, line 438:
```suggestion
		assert.Nil(t, entry)
```

internal/rsl/rsl_test.go
```diff
@@ -395,6 +395,50 @@ func TestGetLatestReferenceEntry(t *testing.T) {
...
+		testRefs := []string{"main", "feature"}
+		for _, ref := range testRefs {
+			if err := NewReferenceEntry(ref, gitinterface.ZeroHash).Commit(repo, false); err != nil {
+				t.Fatal(err)
+			}
+		}
+		hash, hashErr := gitinterface.NewHash("a94a8fe5ccb19ba61c4c0873d391e987982fbbd3")
+		entry, annotations, err := GetLatestReferenceEntry(repo, ForReference("main"), BeforeEntryID(hash), UntilEntryNumber(0))
+		assert.Empty(t, hashErr)
+		assert.Empty(t, annotations)
```
adityasaky, line 413:
```suggestion
		assert.Nil(t, annotations)
```

internal/rsl/rsl_test.go
```diff
@@ -395,6 +395,50 @@ func TestGetLatestReferenceEntry(t *testing.T) {
...
+		// main <- feature
+		testRefs := []string{"main", "feature"}
+		for _, ref := range testRefs {
+			if err := NewReferenceEntry(ref, gitinterface.ZeroHash).Commit(repo, false); err != nil {
+				t.Fatal(err)
+			}
+		}
+		hash, hashErr := gitinterface.NewHash("a94a8fe5ccb19ba61c4c0873d391e987982fbbd3")
+		entry, annotations, err := GetLatestReferenceEntry(repo, ForReference("main"), BeforeEntryID(hash), UntilEntryNumber(0))
+		assert.Empty(t, hashErr)
```
adityasaky, line 412: side note: Empty is the wrong one fwiw. For errors, we want to assert Nil if we don't expect an error.

internal/rsl/rsl_test.go
```diff
@@ -395,6 +395,50 @@ func TestGetLatestReferenceEntry(t *testing.T) {
...
+		assert.Empty(t, annotations)
+		assert.Empty(t, entry)
```
adityasaky, line 414:
```suggestion
		assert.Nil(t, entry)
```

adityasaky (changes requested):
this is pretty close! Some minor nits, but I think this is good otherwise.

internal/rsl/rsl_test.go
rishabhBudhouliya: addressed

internal/rsl/rsl_test.go
rishabhBudhouliya: Understood, addressed

internal/rsl/rsl_test.go
rishabhBudhouliya: Ah, that makes sense. We would not want the execution to go forward if there is an actual error at that point.

rishabhBudhouliya:
@adityasaky I've addressed the comments, please let me know if you see any other issues that need to be addressed.

internal/rsl/rsl_test.go
```diff
@@ -395,6 +395,53 @@ func TestGetLatestReferenceEntry(t *testing.T) {
...
+		// RSL structure for the test
+		// 1    <- 2
+		// main <- feature
+		testRefs := []string{"main", "feature"}
+		for _, ref := range testRefs {
+			if err := NewReferenceEntry(ref, gitinterface.ZeroHash).Commit(repo, false); err != nil {
+				t.Fatal(err)
+			}
+		}
+		hash, err := gitinterface.NewHash("a94a8fe5ccb19ba61c4c0873d391e987982fbbd3")
```
adityasaky, line 410: is this just a random hash?

rishabhBudhouliya: Yes, and while setting the test up I was slightly concerned with using a random hash that is not part of the reference tree (since that would never be an actual case). However, to me, it seems impossible to reach to the branch where the error RSLNotFound is reached with a valid reference tree and before entry and until number options.

rishabhBudhouliya: So, I sort of forced the test to reach to that state with the goal of covering that line of code. But this may not be a logical failure test case for the method involved.

adityasaky: what about using https://github.com/gittuf/gittuf/blob/main/internal/gitinterface/tree.go#L13-L25?

internal/rsl/rsl_test.go
```diff
@@ -395,6 +395,53 @@ func TestGetLatestReferenceEntry(t *testing.T) {
...
+				t.Fatal(err)
+			}
+			latest, err := GetLatestEntry(repo)
+			if err != nil {
+				t.Fatal(err)
+			}
+			entryIDs = append(entryIDs, latest.GetID())
+		}
+
+		entry, annotations, err := GetLatestReferenceEntry(repo, ForReference("main"), BeforeEntryID(entryIDs[0]), UntilEntryNumber(2))
```
adityasaky, line 439: can you add this to the t.Run in line 155? we have an "invalid conditions" test that this is a good fit for.

rishabhBudhouliya: Makes sense, I'll do that.

internal/rsl/rsl_test.go
```diff
@@ -395,6 +395,53 @@ func TestGetLatestReferenceEntry(t *testing.T) {
...
+		for _, ref := range testRefs {
+			if err := NewReferenceEntry(ref, gitinterface.ZeroHash).Commit(repo, false); err != nil {
+				t.Fatal(err)
+			}
+		}
+		hash, err := gitinterface.NewHash("a94a8fe5ccb19ba61c4c0873d391e987982fbbd3")
+		if err != nil {
+			t.Fatal(err)
+		}
+		entry, annotations, err := GetLatestReferenceEntry(repo, ForReference("main"), BeforeEntryID(hash), UntilEntryNumber(0))
```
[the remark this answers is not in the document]
rishabhBudhouliya: Removed

internal/rsl/rsl_test.go
rishabhBudhouliya: I think we need to able to walk the reference chain to be able to reach to that statement. To be clear, to reach [here](https://github.com/gittuf/gittuf/blob/e6c43368733f3e19ddc61a5c18f52e421b898467/internal/rsl/rsl.go#L561), we need an empty reference when the code tries to retrieve a parent. An empty tree would not have a parent in my understanding.

adityasaky: A tree would fail the getparent, yeah. But how certain are we that the hardcoded hash exists? Is this a deterministic hash from the test?

adityasaky: I would at least make it programmatic. We may change the entry structure down the line that changes the hash, and manually updating it will be tough.

rishabhBudhouliya: I didn't understand the relationship between the calculation of hash and the entry structure? To me it seems like they both are unrelated. The hash is being calculated by this utility" https://pkg.go.dev/encoding/hex which won't change unless we change the input.

adityasaky: a94a8fe5ccb19ba61c4c0873d391e987982fbbd3 is an actual commit ID, right? That's created as part of the test? I'm saying that hard coding this commit ID could lead to problems if we ever change the RSL entry structure as the commit ID will then differ.

rishabhBudhouliya: I see. I was weighing in on other options like creating a commit without a number(assuming that we not establish a parent-child relationship between two subsequent commits on the same repo) but the part of code that I test would  depends on numbered commits.

At the moment, I can't think of a way to programmatically calculate the unrelated hash (commit id) without it being tied to the RSL itself. I'll take some time and think more.

adityasaky: It's now set to empty tree ID so we don't have to hardcode a hash. :)

adityasaky (approved):

adityasaky:
thanks for working on this @rishabhBudhouliya!

rishabhBudhouliya:
@adityasaky apologies for not being able to fix the last review recommendation, thank you for fixing it for me!

PR #662, opened 2024-11-12
transport: Fix server response on push

adityasaky:
On pushes, we weren't bubbling up the status of pushed ref correctly because the `dstRefs` check included a trailing newline. This fixes that.

Now:
```
...
To gittuf::github:gittuf/gittuf
 * [new branch]      fix-transport-push-status -> fix-transport-push-status
 ```
 
 Before:
```
...
Writing objects: 100% (6/6), 990 bytes | 990.00 KiB/s, done.
Total 6 (delta 5), reused 0 (delta 0), pack-reused 0
Everything up-to-date
```

### DIFF
--- a/internal/git-remote-gittuf/ssh.go
+++ b/internal/git-remote-gittuf/ssh.go
@@ -35,6 +35,8 @@ func handleSSH(repo *gittuf.Repository, remoteName, url string) (map[string]stri
 	stdInScanner := &logScanner{name: "git-remote-gittuf stdin", scanner: bufio.NewScanner(os.Stdin)}
 	stdInScanner.Split(splitInput)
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: internal/git-remote-gittuf/ssh.go [hunk 1]
```diff
@@ -638,26 +640,26 @@ func handleSSH(repo *gittuf.Repository, remoteName, url string) (map[string]stri
 
 				output = output[4:] // remove length prefix
 				outputSplit := bytes.Split(output, []byte(" "))
+				pushedRef := strings.TrimSpace(string(outputSplit[1]))
```
adityasaky, line 643: here's the fix in ssh.go

adityasaky:
To make debugging easier, this also wraps os.Stdout in a logWriteCloser.

Issue #630, opened 2024-10-30
`gittuf rsl log` should write output incrementally


adityasaky:
The command currently identifies all entries and then writes them to the specified output. This means we have to wait quite a while to see the output of `gittuf rsl log`. Instead, we should pass in the writer and write incrementally.

PR #666, opened 2024-11-19
rsl: log now incrementally writes output

haotran-california:
Aimed to solve issue #630 

The writer, which is now modified to call a buffered `Write` by default, is now passed into the command to identify all entries to the log and write them to output. 

The main command `PrintAllEntriesInRangeFor` is mostly a copy and paste of `GetReferenceEntriesInRangeForRef`. So there is duplication of code in 
`internal/rsl/rsl.go`. 

My reasoning  for this approach is mainly to follow the idea that each function should have one job. If I were to use optional parameters in and conditionals to get `GetReferenceEntriesInRangeForRef` to print, then this function would both be a getter function as well as a printing function. I think this is worse for the health of the codebase than just repeating myself. Additionally there are four other instances of  `GetReferenceEntriesInRangeForRef` being used in various parts of the codebase so I reasoned it was best not to tamper with it. What do you ya'll think? 

I can add tests to this PR later to test the different pagers `less` and `more` because they may not implement a buffered `Write` by default. But I wanted to get feedback before proceeding. 

(P.S I think this project is pretty cool I'd be happy to hack on this with ya'll)

### DIFF
--- a/experimental/gittuf/rsl.go
+++ b/experimental/gittuf/rsl.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"log/slog"
 	"slices"
 	"strings"
@@ -513,4 +514,18 @@ func GetRSLEntryLog(repo *Repository) ([]*rsl.ReferenceEntry, map[string][]*rsl.
 
 	slices.Reverse(entries)
 	return entries, annotationMap, nil
+}
+
+func DisplayRSLLogIncrementally(repo *Repository, bufferedWriter io.WriteCloser) error {
+	firstEntry, _, err := rsl.GetFirstEntry(repo.r)
+	if err != nil {
+		return err
+	}
+
+	lastEntry, err := rsl.GetLatestEntry(repo.r)
+	if err != nil {
+		return err
+	}
+
+	return rsl.PrintAllEntriesInRangeFor(repo.r, firstEntry.GetID(), lastEntry.GetID(), "", bufferedWriter)
 }
--- a/internal/cmd/rsl/log/log.go
+++ b/internal/cmd/rsl/log/log.go
@@ -38,11 +38,6 @@ func (o *options) Run(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
-	entries, annotationMap, err := gittuf.GetRSLEntryLog(repo)
-	if err != nil {
-		return err
-	}
-
 	output := os.Stdout
 	if o.filePath != "" {
 		output, err = os.Create(o.filePath)
@@ -52,11 +47,8 @@ func (o *options) Run(_ *cobra.Command, _ []string) error {
 		o.page = false // override page since we're not writing to stdout
 	}
 
-	outputContents := display.PrepareRSLLogOutput(entries, annotationMap)
 	writer := display.NewDisplayWriter(output, o.page)
-
-	_, err = writer.Write([]byte(outputContents))
-	return err
+	return gittuf.DisplayRSLLogIncrementally(repo, writer)
 }
 
 func New() *cobra.Command {
--- a/internal/display/display.go
+++ b/internal/display/display.go
@@ -38,29 +38,30 @@ func pageWriter(pagerBin string, defaultOutput io.Writer) *pager {
 }
 
 type pager struct {
-	command *exec.Cmd
+	command     *exec.Cmd
+	stdInWriter io.WriteCloser
+	started     bool
 }
 
 func (p *pager) Write(contents []byte) (int, error) {
-	stdInWriter, err := p.command.StdinPipe()
+	// Initialize the command and stdin pipe if not already started
+	if !p.started {
+		stdInWriter, err := p.command.StdinPipe()
+		if err != nil {
+			return -1, err
+		}
+		p.stdInWriter = stdInWriter
+
+		if err := p.command.Start(); err != nil {
+			return -1, err
+		}
+
+		p.started = true
+	}
+
+	// Write to the pager's stdin pipe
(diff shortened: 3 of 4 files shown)


haotran-california:
I ended up sleeping on it and tried a different approach. A boolean parameter called `isReturnEntriesOrdered`, which dictates if  all the reference entries are returned in sorted or reversed order, was added to `GetReferenceEntriesInRangeForRef`. It should be cleaner implementation because it deletes two functions I had previously added as well as leaves the `display` package isolated to `cmd/rsl/log`

adityasaky:
Hey @haotran-california, thanks for taking a stab at this! I think it might be worth taking a step back and identifying the key issue. Basically, `gittuf rsl log` takes a long time if the repository has a bunch of entries in `refs/gittuf/reference-state-log`, especially on some operating systems where loading git objects is more expensive than others. I think we might want to amend the original issue because it's not just about writing this in a buffered way but also just how many entries are loaded before it's displayed. This might mean more surgery with popping into `internal/rsl` in how entries are returned to the writer? WDYT?

also cc @wlynch.

haotran-california:
It was honestly pretty fun and the codebase is pretty well organized. I'd be happy to give a more thoughtful stab at this issue after it is redefined. Also when I am in agreement with you @wlynch 

Here's my understanding so far. 

Suppose you have thousands of commits in the refs/gittuf/reference-state-log namespace. In my current solution, the user has to wait for gittuf to read all entries from beginning to end before anything starts printing on screen. It could take a few seconds and make the gittuf rsl log command look broken. So the ideal solution is to write to the screen while reading the commits in the rsl. This way the user instantaneously gets a logs printed to their console while they wait.

Perhaps this can be implemented in rsl/internal by creating and function like GetNextReferenceEntriesFromCurrent(arg1, arg2) which takes in the current reference entry as arg1 and the number of entries to read as arg2 and returns some array of entries. I feel like this is what you're hinting at by saying it's also about how many entries are loaded. 

The reason we need to load arg2 number of entries is that to display a log we need to account for AnnotationEntries. Let's say a ReferenceEntry appears at index x, then it's corresponding  annotation could possibly exist at index x + a, where a is len(rsl) - x. So it's possible for the Annotation to exist anywhere in the rsl. But if we assume users will generally annotate a entry within like 50 or so changes to the rsl, then we can just read 50 commits and see if there is an annotation or otherwise move on. Although this might miss some annotations the buffered reading will allow the for user to have immediate output. 

I kind of had the idea of reworking the internals/rsl more extensively at first. But I was a bit scared to make such a major change as my first issue. And I honestly didn't have all the ideas in place either.

adityasaky:
I think what you're describing aligns with my thinking on this. No amount of buffered writing will solve this issue if we still spend a bunch of time loading all entries into memory. I'm not sure of the best way to handle this. Do we just continually load the next n entries behind the scenes and write to the buffer? Does that interplay well with pager programs? Can we track what the user has _seen_ and lazy load only when necessary? etc

haotran-california:
@zilinski @adiyita

Thank you for both getting back to me so quick. I was going to respond earlier but decided to spend some time hacking on things first before replying. 

What if we split up the entries in the output like so? 

--------------------------------
PRINT ALL REF ENTRIES
--------------------------------
PRINT ALL ANNOTATIONS
--------------------------------

This would mean we can do buffered read of entries followed by buffered write of just the references. While collecting all the annotations along the way. The user will see continuous output without an initial delay. And this log correctly contains and displays all annotations without missing any, which was not guaranteed by the previous methods we thought of together. 

My previous buffer method could miss some annotations if they were beyond the length of the buffer. 

Apparently in a pager you can track keystrokes as inputs. So we could do some algebra to map this to what the user has seen and lazy load the next buffer when appropriate which is really clever. In my head I reasoned that if the user spams down arrow or the j key then there's a chance not enough time has passed for the program to get all relevant annotations for that next buffer. But now that I'm writing that down it seems pretty unlikely.

What do ya'll think?

haotran-california:
10,000 RSL Entries 
command: time gittuf rsl log

```
Error: write |1: broken pipe
real    1m42.868s
user    1m3.208s
sys     1m1.677s
```

I also ended up getting curious about the 1000 entries and subsequently the 10,000 entries test case and timed a few different runs. Here's an example of one of those runs with my implementation and some of my observations. 

First off I ended up breaking the pipe a bunch of times although all the entries still displayed in the pager every-time. Secondly, I noticed that 95% of the runtime of the log function was just the `GetFirstEntry` as it loads all entries into memory via a cache. So when I wrote optimized code for displaying entries it just executed the other 5% more quickly. This meant that my function was marginally faster if not slower sometimes. Thirdly, I suppose test cases will never have to account for more than around 20,000 commits as the largest projects I'm aware of has roughly that amount. 

Not sure if this information is super useful besides the pipe breaking but I thought I might share.

adityasaky:
@haotran-california take a look at https://github.com/gittuf/gittuf/pull/666/commits/0b20028bf04adcf25dbfce8e9fde6c739b7ac4ca, thanks.

cc @patzielinski

haotran-california:
@adityasaky Yeah it can totally run without the `GetNextReferenceEntryBuffer` function, the changes are succinct

adityasaky:
@haotran-california one note, could you clear up the commit history? Also, we prefer not having merge commits when the feature branch is out of date, instead opting for a rebase based workflow. Thanks!

haotran-california:
Sure! I'd be happy to try and clean this up, thank you for all the guidance and approvals by @patzielinski

haotran-california:
@patzielinski 

I ended up failing to cleanup the commit history and made a mess. I think I failed to include the main branch in rebasing. Would it be okay to use a new PR?

PR #676, opened 2024-12-04
transport: Check for empty list for-push result

### LINKED ISSUE #674: transport: Incorrect access token causes the transport to freeze
When using the transport + authenticating with GitHub (over HTTPS), an incorrect personal access token will cause the transport to freeze without exiting.

patzielinski:
If the user fails to authenticate with the remote using the http helper, then `list for-push` returns nothing, which is now handled by exiting with an error.

Closes #674.

### DIFF
--- a/internal/git-remote-gittuf/curl.go
+++ b/internal/git-remote-gittuf/curl.go
@@ -7,6 +7,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


internal/git-remote-gittuf/curl.go
```diff
@@ -347,6 +350,12 @@ func handleCurl(repo *gittuf.Repository, remoteName, url string) (map[string]str
 			for helperStdOutScanner.Scan() {
 				output := helperStdOutScanner.Bytes()
 
+				// If nothing is returned, the user has likely failed to
+				// authenticate with the remote
+				if len(output) == 0 {
```
adityasaky, line 355: we should test this change on windows.

adityasaky: is the single byte a newline?

adityasaky (approved):

Issue #280, opened 2024-01-17
Enable expiry enforcement on latest policy


adityasaky:
### Add a description

The gittuf policy metadata has an "expires" field that is currently not enforced. We should start enforcing it but only when the policy is the _latest_ one. In historic policy states, we should ignore the expiry and assume it was valid for changes at the time.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #702, opened 2024-12-18
docs: Major reworking of design document

adityasaky:
This cleans up various parts of the design document, including updating it to reflect some changes. The document now also points to the accepted NDSS paper.

### DIFF
--- a/docs/design-document.md
+++ b/docs/design-document.md
@@ -1,43 +1,52 @@
 # gittuf Design Document
 
-Last Modified: October 21, 2024
+Last Modified: December 18, 2024
 
 ## Introduction
 
-This document describes gittuf, a security layer for Git repositories. gittuf
-applies several key properties part of the
-[The Update Framework (TUF)](https://theupdateframework.io/) such as
-delegations, secure key distribution, key revocation, trust rotation, read /
-write access control, and namespaces to Git repositories. This enables owners of
-the repositories to distribute (and revoke) contributor signing keys and define
-policies about which contributors can make changes to some namespaces within the
-repository. gittuf also protects against reference state attacks by extending
-the Reference State Log design which was originally described in an
-[academic paper](https://www.usenix.org/conference/usenixsecurity16/technical-sessions/presentation/torres-arias).
-Finally, gittuf can be used as a foundation to build other desirable features
-such as cryptographic algorithm agility, the ability to store secrets, storing
-in-toto attestations pertaining to the repository, and more.
+This document describes gittuf, a security layer for Git repositories.
+With gittuf, any developer who can pull from a Git repository can independently
+verify that the repository's security policies were followed. gittuf's policy,
+inspired by [The Update Framework (TUF)](https://theupdateframework.io/),
+handles key management for all trusted developers in a repository, allows for
+setting permissions for repository namespaces such as branches, tags, and files,
+and provides protections against
+[attacks targeting Git metadata](https://www.usenix.org/conference/usenixsecurity16/technical-sessions/presentation/torres-arias).
+At the same time, gittuf is backwards compatible with existing source control
+platforms ("forges") such as GitHub, GitLab, and Bitbucket. gittuf is currently
+a sandbox project at the
+[Open Source Security Foundation (OpenSSF)](https://openssf.org/) as part of the
+[Supply Chain Integrity Working Group](https://github.com/ossf/wg-supply-chain-integrity).
+The core concepts of gittuf described in this document have been
+[peer reviewed](https://ssl.engineering.nyu.edu/papers/yelgundhalli_gittuf_ndss_2025.pdf).
 
-This document is scoped to describing how TUF's access control policies are
-applied to Git repositories. It contains the corresponding workflows for
-developers and their gittuf clients. Note that gittuf is designed in a manner
-that enables other security features. These descriptions will be in standalone
-specifications alongside this one, and will describe modifications or extensions
-to the "default" workflows in this document.
+This document is scoped to describing how gittuf's write access control policies
+are applied to Git repositories. Other additions to gittuf's featureset are
+described in standalone ["extensions" documents](/docs/extensions/).
 
 ## Definitions
 
 This document uses several terms or phrases in specific ways. These are defined
 here.
 
-### Git Reference (Ref)
+### Git References (Refs) and Objects
 
 A Git reference is a "simple name" that typically points to a particular Git
 commit. Generally, development in Git repositories are centered in one or more
 refs, and they're updated as commits are added to the ref under development. By
-default, Git defines two of refs: branches (heads) and tags. Git allows for the
-creation of other arbitrary refs that users can store other information as long
-as they are formatted using Git's object types.
+default, Git defines two of refs: branches ("heads") and tags. Git allows for
+the creation of other arbitrary refs that users can store other information as
+long as they are formatted using Git's object types.
+
+Git employs a content addressed object store, with support for four types of
+objects. An essential Git object is the "commit", which is a self-contained
+representation of the whole repository. Each commit points to a "tree" object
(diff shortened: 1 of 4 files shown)


docs/design-document.md
```diff
@@ -64,76 +73,170 @@ Repository
...
-policy allows an actor to make certain changes, it's actually allowing the
-person who has the specific signing key to make those changes. To maintain
-security, all actions made in the repository, such as adding or modifying files,
-are checked for authenticity. This is done by verifying the digital signature
-attached to the action, which must match the public key associated with the
-actor who is supposed to have made the change.
+In a Git repository, an "actor" is any party, human or bot, who makes changes to
+the repository. These changes can involve any part of the repository, such as
+modifying files, branches or tags. In gittuf, each actor is identified by a
+unique signing key that they use to cryptograhpically sign their contributions.
```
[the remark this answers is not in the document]
adityasaky: that's an extension (#615)

docs/design-document.md
```diff
@@ -64,76 +73,170 @@ Repository
...
+_In this example, the repository administrator grants write permissions to Carol 
+for the main branch, to Alice for the alice-dev branch, and to Bob for the
+/tests folder (under any of the existing branches)._
+
+This mechanism is employed when verifying both RSL entries for Git ref updates
+_and_ when verifying the commits introduced between two ref updates. The latter
+option allows for defining policies to files and directories tracked by the
+repository. It also enables repository owners to define closed sets of
+developers authorized to make changes to the repository. Note that gittuf does
+not by default use Git commit metadata to identify the actor who created it as
```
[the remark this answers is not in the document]
adityasaky: Moved this line and after move, I've said email.

docs/design-document.md
```diff
@@ -449,73 +473,7 @@ repository. Clara and Dave are only authorized to work on files within `bar/*`
...
-```bash
-gittuf rsl record
-gittuf rsl annotate
-```
-
-Note: the commands listed here are examples and not exhaustive. Please refer to
-gittuf's help documentation for more specific information about gittuf's usage.
+## gittuf Workflows
 
 ### Syncing gittuf namespaces with the main repository
```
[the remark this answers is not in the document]
adityasaky: we could, happy to do that separately

docs/design-document.md
```diff
@@ -64,76 +73,192 @@ Repository
...
-attached to the action, which must match the public key associated with the
-actor who is supposed to have made the change.
+In a Git repository, an "actor" is any party, human or bot, who makes changes to
+the repository. These changes can involve any part of the repository, such as
+modifying files, branches or tags. In gittuf, each actor is identified by a
+unique signing key that they use to cryptographically sign their contributions.
+gittuf uses cryptographic signatures to authenticate actors as these signatures
+can be verified by anyone who has the corresponding public key, fundamental to
+gittuf's mission to enable independent verification of repository actions. Note
+that gittuf does not use Git commit metadata (e.g., author email, committer
```
JustinCappos, line 85: Is it true we never use this information for anything?   What if it doesn't match what we expect for a user?   Would it be more accurate to position this as information we don't rely on?

docs/design-document.md
```diff
@@ -64,76 +73,192 @@ Repository
...
+email) to identify the actor who created it as that may be trivially spoofed.
+
+In practice, when a gittuf policy allows an actor to make certain changes, it's
+actually allowing the person who has the specific signing key to make those
```
JustinCappos, line 89: should 'person' -> 'identity'?   If so, don't just make the change because identities aren't known to the reader.   

Also, you talk about 'actor' before this and 'person' here?   Is an 'actor' a 'person'?   Do we need two terms?

docs/design-document.md
```diff
@@ -554,15 +526,19 @@ The workflow is as follows:
...
    been successfully verified, each ref's remote state can be applied to the
    local repository, so `refs/heads/<ref>` matches `refs/remotes/origin/<ref>`.
 1. Set local RSL to the remote RSL's tip.
 
-#### RSLPush: Submitting local RSL changes
+NOTE: Some aspects of this workflow are under discussion and are subject to
+change. The gittuf implementation does not implement precisely this workflow.
+Specifically, the implementation does not verify new entries in the remote
+automatically. Additionally, the RSL may contain entries for references a client
+does not have, making verification of those entries unfeasible.
```
JustinCappos, line 539: This seems weird to have right after you present an algorithm.   Why not have the current text here and open an issue with the discussion.   You can point to that here.

docs/design-document.md
```diff
@@ -572,62 +548,78 @@ The workflow is as follows:
...
-         1. Set `parent` to `current`, `current` to delegatee role.
-         1. Continue to next iteration.
-      1. Else:
-         1. Set `K` to keys authorized in the delegations entry.
+This is achieved by performing pre-ordered depth first search over the
+delegations graph in a gittuf policy state. Assume the relevant policy state
+entry is `P` and the namespace being checked is `N`. Then:
+
+1. Validate `P`'s root metadata using the TUF workflow starting from the initial
+   root of trust metadata, ignore expiration date checks.
```
JustinCappos, line 594: I guess this isn't the place, but the rationale for ignoring this should be explained.

docs/design-document.md
```diff
@@ -798,40 +758,29 @@ examples of such invalid entries are:
...
+buggy client.
+
+##### A gittuf Access Control Policy is Violated
+
+An actor, Bob, creates an RSL entry for a branch he's not authorized for by
+gittuf policy. He pushes a change to that branch. Another actor, Alice, notices
+this when her gittuf client indicates a failure in the verification workflow.
+Alice creates an RSL annotation marking Bob's entry as one to be skipped. Alice
+also reverses Bob's change, creating a new RSL entry reflecting that. Thus, this
+demonstrates gittuf's recovery workflow in action.
```
JustinCappos, line 781: what does this last sentence really mean?

docs/extensions/hash-algorithm-agility.md
```diff
@@ -116,4 +118,4 @@ of the correct SHA-256 hash.
 
 **T:** We'd have to pass around a prior calculated SHA-256 hash via the
 translation mapping. However, if that must be trusted, we'd also have to ensure
-it wasn't tampered with. TUF semantics can help here.
\ No newline at end of file
+it wasn't tampered with. TUF semantics can help here.
```
JustinCappos, line 121: Diff is????

JustinCappos (approved):
Looks basically good.   A few comments / questions / wording things.

docs/design-document.md
adityasaky: We don't use it at all at the moment, not even for whether it matches for a user. We don't have a "source of truth" for that information. We could infer in some cases if a person uses sigstore to sign by looking at the email + provider but there's no reason for the oidc identity used there to match the committer email.

I've updated the text to say "rely on" rather than "use".

docs/design-document.md
adityasaky: actor is generic, can be person or bot. This shouldn't really say "person" here though.

adityasaky: rephrased to avoid confusion.

docs/design-document.md
adityasaky: it's unneeded given changes to the recovery flows, taken out.

docs/extensions/hash-algorithm-agility.md
adityasaky: probably a newline at the end.

docs/design-document.md
adityasaky: #280 I'll add a pointer to this

docs/design-document.md
adityasaky: https://github.com/gittuf/gittuf/issues/708

adityasaky: Added inline.

PR #712, opened 2025-01-02
Maintain persistent cache

adityasaky:
This is a continuation of the persistent cache work.


adityasaky:
This is far from complete but I think it's reasonable to merge this in as scaffolding for actually using (and adding other necessary APIs).

PR #724, opened 2025-01-08
policy: Streamline verification helper

adityasaky:
We have two very similar helpers: verifyGitObjectAndAttestations and verifyGitObjectAndAttestationsUsingVerifiers. The first one calls the second one. However, there were numerous cases of direct invocation of the second one. This commit makes the first helper the one that's used consistently. This ensures that finding which verifiers to use and so on are consistent everywhere.


adityasaky:
This really came up while I was continuing work on #699.

PR #729, opened 2025-01-09
*: Support multiple types of global rules

adityasaky:
This commit adds support for different types of global rules (i.e., non threshold constraints). This commit also adds a global rule to block force pushes to one or more specified refs.


docs/cli/gittuf_rsl_remote_check.md
adityasaky, line 1: looks like we didn't remove this when the command was removed...

adityasaky: make generate should probably delete docs/cli/*.md and then generate for the CI to catch it. CI will catch lack of doc or changes, but not deletions since regenerating won't affect the one that must be deleted.

experimental/gittuf/root.go
```diff
@@ -493,12 +493,56 @@ func (r *Repository) AddGlobalRule(ctx context.Context, signer sslibdsse.SignerV
 		return err
 	}
 
-	slog.Debug("Adding global rule...")
-	if err := rootMetadata.AddGlobalRule(name, patterns, threshold); err != nil {
+	slog.Debug("Adding threshold global rule...")
+	if err := rootMetadata.AddGlobalRule(tufv01.NewGlobalRuleThreshold(name, patterns, threshold)); err != nil {
```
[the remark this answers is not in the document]
adityasaky: it's really just a question of where we've defined it. they're identical for tufv02 as well, but I want to wait and see if any future constraints can only be added to a specific policy version or so

Issue #744, opened 2025-01-17
Update gittuf's extensions process to be formalized


adityasaky:
### Add a description

gittuf's design document can be enhanced using extension documents. These are ad-hoc with no clear format or language guidelines (I'm at fault here as the author of most of them). They are also not numbered, making referring to them more difficult as more are proposed (we have three open PRs right now that collectively add five extension documents).

Despite gittuf not being a specification-driven project like in-toto and TUF, I think a formal augmentation proposal process modeled on ITEs or TAPs is a good idea. @patzielinski suggested gittuf Augmentation Proposal (GAP), pun intended as a GAP can address a gap in gittuf's design.

### Relevant log output if the discussion pertains to existing gittuf functionality

```Shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #745, opened 2025-01-17
docs: Add GAPs

adityasaky:
Addresses #744


adityasaky:
If the maintainers / gittuf TSC members agree with this broadly (i.e., having such a process), we can separately move over existing "extension" docs to be GAPs. We also have about 5 more proposed as open PRs, which is really why it's probably a good idea to make this a proper process for how they're reviewed and implemented.

JustinCappos (commented):
My thinking is that this may be overly bureaucratic given the code first nature of the project.   I think this is a good idea if we ever become a spec first project with ideas being upstreamed into other systems, but it feels like overkill now.   I'd recommend tabling this until there is more of a need.

For now, I think we could have a design document with a defined format, but it feels like some of the effort here is too much for now...

adityasaky:
That's a good point, I definitely don't want us to get bogged down in the "process". Maybe whittling this down to essentially saying "here's everything a GAP / extension must have and address" without going into statuses and transitions is enough, as you suggest ("defined format").

More than anything, I want to ensure "big" changes or feature additions are designed systematically with sufficient reasoning about security impacts, in a way that survives changes in maintainership. I also do have half an eye on how our ideas may make their way into other projects (for e.g., I think there's interest in extending the work in https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3trNYnLWS11cJWC6BbxDs5niGo82/remotes/z6MkireRatUThvd3qzfKht1S44wpm4FEWSSa4PRMTSQZ3voM/tree/patches/1d1ce874f7c39ecdcd8c318bbae46ffd02fe1ea8/0004-canonical-references.md to support security controls / rules that are similar to gittuf).

mlieberman85:
I agree with @JustinCappos. There's always room to add more bureaucracy but if the main need is just making sure that big changes aren't just pushed through quickly I think just putting something simple around that in a contributing.md would be fine. Just my two cents but I found these sorts of governance changes are useful after it's clear something has gone wrong or someone pushes something they're not supposed to.

adityasaky:
@JustinCappos I've scaled this back to focus on the process and to make them very explicitly living documents. The emphasis now is on the format to ensure a proposal reasons about the right things. Can you take a look?

JustinCappos (approved):
Definitely this is better.   I could quibble more about even less process, but in the interest in moving ahead, I'm fine merging it.

adityasaky:
@mlieberman85 apologies, I didn't see your message before (github hadn't refreshed and shown it).

I agree with your point, making it part of contributing.md is also a good idea. I've just pushed a commit with that change. While things haven't already gone wrong, I decided to introduce something a bit more structured because we do have a number of "extension" documents in the repository and as pull requests that don't follow a strict structure and do not sufficiently reason about the changes. I'm guilty of this (all of those docs have been submitted by me so far) and this is to correct my own workflow and set a good baseline going forward.

When you get a chance, can you take a quick look at my changes from today? I've tried to scale it back so the only real prescribed thing here is the format of the document for a change and that these proposals are merged into the repository quickly even if they end up not getting implemented so that they're all tracked. Thanks!

mlieberman85 (approved):

docs/gaps/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+## GAP Responsibilities
+
+The participants in a GAP have specific responsibilities.
+
+### Sponsor
+
+Every GAP must have at least one sponsor. There is no limit to the number of
+sponsors a GAP may have. Each sponsor must be listed in the GAP's metadata
+section.
```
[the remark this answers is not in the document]
adityasaky: Updated

docs/gaps/README.md
```diff
@@ -0,0 +1,190 @@
...
+* Engaging in discussions of problems to determine if a GAP is necessary
+* Reviewing and providing feedback on a GAP in a timely manner with a focus on
+  the impact of the proposed changes on gittuf's security model
+* Ensuring the GAP follows the prescribed format
+* Updating merged GAPs (if the sponsors do not) to reflect the state of their
+  implementation; for example, if a GAP is implemented, if an implemented GAP
+  feature is removed, or a GAP feature is updated significantly, the maintainers
+  must ensure this is reflected in the living GAP document
+
+Changes to GAP documents are subject to approval from the same threshold of
```
reza-curtmola, line 118: Considering that a GAP can potentially have a bigger impact, are we comfortable with the same threshold required for a (potentially much more minor) implementation change?

docs/gaps/README.md
```diff
@@ -0,0 +1,190 @@
...
+regardless of their implementation status. This increases the visibility of each
+GAP, thus making it easier for other interested parties to discover the GAP,
+propose further changes, and contribute to the implementation of the GAP.
+
+The sponsors of a GAP may choose to withdraw their proposal or the maintainers
+may choose to reject the proposed changes after assessing the GAP. Even in these
+cases, the document must be merged into the repository with the status indicated
+in the document's metadata section. Note that a GAP must only be withdrawn or
+rejected on the basis of technical reasons (e.g., a better solution is proposed
+or a security issue is discovered a consequence of the proposal). The reasoning
```
reza-curtmola, line 159: typo: should be "is discovered as a consequence of the proposal".

docs/gaps/README.md
```diff
@@ -0,0 +1,190 @@
...
+The changes proposed in a GAP may be implemented via patches to the gittuf
+implementation. The changes to the implementation need not be submitted by the
+sponsors of the GAP. When the gittuf maintainers think a GAP has been
+implemented, they can propose an update to the document reflecting this. The
+sponsors of a GAP may also propose marking a GAP as implemented, which is
+subject to approval from the gittuf maintainers.
+
+### Removing a GAP's implementation
+
+If a GAP's implementation is removed, then the GAP must be updated to indicate
```
reza-curtmola, line 173: It would be nice to give a rationale as to why a GAP's implementation may be removed (maybe give an example).

reza-curtmola (commented):
Added some comments

docs/gaps/README.md
adityasaky: This is one of the processes I scaled back (earlier there were stricter review requirements for GAPs).

I think the same threshold is fine because ultimately, the big changes can be merged into the implementation without any reasoned documentation and discussion (what a GAP provides us) by a "small" threshold of reviews. So requiring GAPs at all is already an improvement I think. If we have an explosion in number of GAPs and feature additions, we can revise this at a later date.

docs/gaps/README.md
adityasaky: fixed!

docs/gaps/README.md
adityasaky: Added some example scenarios. I think this is likely to fall into two categories:
a) the feature is a security nightmare and we are willing to take the backwards compatibility hit to remove it
b) the feature is made redundant due to other GAPs or more fundamental changes to gittuf's design

reza-curtmola (approved):
Not sure we need the "remove a GAP" phase, as a GAP can also be removed by simply creating another GAP to remove it :) But, fine!

neilnaveen (approved):
This looks good to me, having fewer restrictions for pushing features sounds like a good idea for gittuf since we are still in an alpha/beta stage

wlynch (approved):

adityasaky:
Thanks all!

PR #766, opened 2025-01-28
policy: Add VerifyMergeableForCommit

adityasaky:
This is a stab at not using the RSL for the feature ref when checking for mergeability. We need this for the github app case.


adityasaky:
I'm also evaluating other cases where this may be needed.

adityasaky:
oh rip, i forgot the experimental/gittuf piece.

PR #770, opened 2025-01-29
Adding Transport Tests

zsun6:
Tests gittuf remote with http requests

### DIFF
--- a/.github/workflows/remote-tests.yml
+++ b/.github/workflows/remote-tests.yml
@@ -1,0 +1,41 @@
+name: remote tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        git-version: [2.34.1, 2.40.1]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Set up Git
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y git
+
+    - name: Set up GitHub SSH key (if using SSH)
+      run: |
+        mkdir -p ~/.ssh
+        echo "${{ secrets.GITHUB_DEPLOY_KEY }}" > ~/.ssh/id_rsa
+        chmod 600 ~/.ssh/id_rsa
+        ssh-keyscan github.com >> ~/.ssh/known_hosts
+
+    - name: Build and install git-remote-gittuf
+      run: |
+        go build -o /usr/local/bin/git-remote-gittuf ./cmd/git-remote-gittuf
+
+    - name: Run integration tests
+      run: |
+        ./scripts/run-tests.sh
--- a/runtests.py
+++ b/runtests.py
@@ -1,0 +1,101 @@
+import os
+import subprocess
+import shlex
+
+def run_command(cmd, expected_retcode=0):
+    """Runs the supplied command and checks for the expected return code."""
+    retcode = subprocess.call(shlex.split(cmd))
+    if retcode != expected_retcode:
+        raise Exception(f"Expected {expected_retcode} from process but it exited with {retcode}.")
+
+# Constants
+GITTUF_BINARY = "git-remote-gittuf"
+TEST_REPO_URL_SSH = "gittuf::git@github.com:gittuf/zsun6"
+TEST_REPO_URL_HTTPS = "gittuf::https://github.com/zsun6/gittuf-remote-test_repo"
+TEST_CLONE_DIR = "test_clone"
+TEST_REMOTE_NAME = "origin"
+GITTUF_TEST_BRANCH = "test-branch"
+
+# Verify if the gittuf binary is installed
+try:
+    run_command(f"command -v {GITTUF_BINARY}")
+    print("Git-Remote-Gittuf binary is installed.")
+except Exception:
+    print(f"Error: {GITTUF_BINARY} is not installed or not in PATH.")
+    exit(1)
+
+# Clean up from previous runs
+if os.path.isdir(TEST_CLONE_DIR):
+    subprocess.call(["rm", "-rf", TEST_CLONE_DIR])
+    print("Cleaned up previous test environment.")
+
+REPO_URL = TEST_REPO_URL_HTTPS
+
+# Test cloning the repository
+print(f"Testing clone using {REPO_URL}...")
+run_command(f"git clone {REPO_URL} {TEST_CLONE_DIR}")
+os.chdir(TEST_CLONE_DIR)
+
+if not os.path.isdir(".git"):
+    print("Error: Failed to clone repository.")
+    exit(1)
+print("Successfully cloned repository.")
+
+# Test setting remote to HTTPS
+print("Testing setting remote URL to HTTPS...")
+run_command(f"git remote set-url {TEST_REMOTE_NAME} {TEST_REPO_URL_HTTPS}")
+
+remote_url = subprocess.check_output(shlex.split(f"git remote get-url {TEST_REMOTE_NAME}")).strip().decode()
+if remote_url != TEST_REPO_URL_HTTPS:
+    print("Error: Failed to set remote URL to HTTPS.")
+    exit(1)
+print("Successfully set remote URL to HTTPS.")
+
+# Test first fetch
+print("Testing fetch operation...")
+run_command(f"git fetch {TEST_REMOTE_NAME}")
+
+# Check if the repository folder exists
+repo_dir = "repo"
+if os.path.isdir(repo_dir):
+    print(f"Repository folder '{repo_dir}' found.")
+    readme_file = os.path.join(repo_dir, "README.md")
+    if os.path.isfile(readme_file):
+        print("README file found. Printing contents:")
+        with open(readme_file, "r") as file:
+            print(file.read())
+    else:
+        print(f"README file not found in '{repo_dir}'.")
+else:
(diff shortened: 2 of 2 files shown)


.github/workflows/runtests3.py
patzielinski, line 1: I presume this is the file that contains the latest edits? If so, please remove the other two .py scripts from the PR and rename this one something like `run-transport-tests.py`.

.github/workflows/runtests3.py
```diff
@@ -0,0 +1,120 @@
...
+    subprocess.run(['git', '-C', bare_repo_path, 'update-server-info'], check=True)
+    
+    # Generate SSL certificate
+    keyfile = os.path.join(server_dir, 'key.pem')
+    certfile = os.path.join(server_dir, 'cert.pem')
+    subprocess.run([
+        'openssl', 'req', '-x509', '-newkey', 'rsa:4096',
+        '-keyout', keyfile, '-out', certfile, '-days', '365', '-nodes',
+        '-subj', '/CN=localhost'
+    ], check=True)
```
patzielinski, line 59: We should use a pre-made certificate here - the [actions run](https://github.com/gittuf/gittuf/actions/runs/13819866927/job/38662313056) shows a timeout, likely due to SSL cert generation eating up most of the time. You can add the certificate material to this PR - it doesn't matter if it's public, much like the keys we use for the gittuf demo.

.github/workflows/runtests3.py
```diff
@@ -0,0 +1,120 @@
...
+    subprocess.run(['git', '-C', bare_repo_path, 'config', 'http.receivepack', 'true'], check=True)
+    subprocess.run(['git', '-C', bare_repo_path, 'config', 'http.uploadpack', 'true'], check=True)
+    
+    # Initialize test repository
+    subprocess.run(['git', 'init', local_repo_dir], check=True)
+    with open(os.path.join(local_repo_dir, 'README.md'), 'w') as f:
+        f.write('# Local Test Repository\n')
+    subprocess.run(['git', '-C', local_repo_dir, 'add', '.'], check=True)
+    subprocess.run(['git', 'config', '--global', 'user.email', "'you@example.com'"],check =True)
+    subprocess.run(['git', 'config', '--global', 'user.name', "'Your Name'"],check =True)
```
patzielinski, line 39: Let's make these local configuration options in case we need to try this out on a machine. This will mean the other repo setup commands need to be updated to include these as well.
```suggestion
    subprocess.run(['git', 'config', '-C', local_repo_dir, '--local', 'user.email', "'you@example.com'"],check =True)
    subprocess.run(['git', 'config', '-C', local_repo_dir, '--local', 'user.name', "'Your Name'"],check =True)
```

.github/workflows/runtests3.py
```diff
@@ -0,0 +1,120 @@
...
+        print("Testing fetch...")
+        run_command(f"git -c http.sslVerify=false fetch {TEST_REMOTE_NAME}")
+        
+        print("Testing push...")
+        run_command(f"git checkout -b {GITTUF_TEST_BRANCH}")
+        with open("testfile.txt", "w") as f:
+            f.write("test content")
+        run_command("git add testfile.txt")
+        run_command(f"git commit -m 'Add test file'")
+        run_command(f"git push -u {TEST_REMOTE_NAME} {GITTUF_TEST_BRANCH}", 128)
```
patzielinski, line 111: Likely needs the `sslVerify` option here as well.

```suggestion
        run_command(f"git push -c http.sslVerify=false -u {TEST_REMOTE_NAME} {GITTUF_TEST_BRANCH}", 128)
```

.github/workflows/runtests3.py
```diff
@@ -0,0 +1,120 @@
...
+from http.server import HTTPServer, SimpleHTTPRequestHandler
+
+def run_command(cmd, expected_retcode=0):
+    """Runs the supplied command and checks for the expected return code."""
+    retcode = subprocess.call(shlex.split(cmd))
+    if retcode != expected_retcode:
+        raise Exception(f"Expected {expected_retcode} from process but it exited with {retcode}.")
+
+# Cleanup previous runs
+TEST_CLONE_DIR = "test_clone"
```
patzielinski, line 17: I'd suggest putting globals at the top of the file below imports, e.g. line 9.

.github/workflows/runtests3.py
```diff
@@ -0,0 +1,120 @@
...
+        
+        print("Testing push...")
+        run_command(f"git checkout -b {GITTUF_TEST_BRANCH}")
+        with open("testfile.txt", "w") as f:
+            f.write("test content")
+        run_command("git add testfile.txt")
+        run_command(f"git commit -m 'Add test file'")
+        run_command(f"git push -u {TEST_REMOTE_NAME} {GITTUF_TEST_BRANCH}", 128)
+        
+        print("All tests passed!")
```
patzielinski, line 113: The process hangs using the transport, but even after removing the `gittuf::` prefix, the script fails with

```
127.0.0.1 - - [13/Mar/2025 22:42:36] "GET /repo.git/info/refs?service=git-receive-pack HTTP/1.1" 200 -
127.0.0.1 - - [13/Mar/2025 22:42:36] "GET /repo.git/HEAD HTTP/1.1" 200 -
127.0.0.1 - - [13/Mar/2025 22:42:36] code 501, message Unsupported method ('PROPFIND')
127.0.0.1 - - [13/Mar/2025 22:42:36] "PROPFIND /repo.git/ HTTP/1.1" 501 -
error: Cannot access URL https://localhost:8000/repo.git/, return code 22
fatal: git-http-push failed
error: failed to push some refs to 'https://localhost:8000/repo.git'
```

It looks like you need to implement `PROPFIND` as was done with `do_GET` on line 66?

patzielinski (commented):
Could you also test this out with the transport and using the `GITTUF_LOG_FILE` env var to produce a logfile? That way we can narrow down in the transport where the issue may be.

Thanks!

.github/workflows/Git-transport-tests.py
```diff
@@ -0,0 +1,119 @@
...
+        content = f.read().strip()
+    if content == 'hello world':
+        print(f"[+] Push successful! README.md contains: '{content}'")
+    else:
+        print(f"[-] Push failed. README.md contains: '{content}'")
+
+
+if __name__ == '__main__':
+    # Ensure git-http-backend exists
+    if not shutil.which('git-http-backend'):
```
patzielinski, line 116: Is this referencing `git http-backend`, without the dash in between git and http?

.github/workflows/remote-tests-Gittuf.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Install Go
+      uses: actions/setup-go@3041bf56c941b39c61721a86cd11f3bb1338122a
+      with:
+        go-version: ${{ matrix.go-version }}
+
+    - name: Build gittuf
+      run: make just-install
+
+    - name: Expose git-http-backend
+      run: echo "$(git --exec-path)" >> $GITHUB_PATH
```
patzielinski, line 31: Is this just adding Git to the PATH?

.github/workflows/Git-transport-tests.py
```diff
@@ -0,0 +1,119 @@
...
+BARE_PARENT = os.path.join(BASE_DIR, 'repos')
+BARE_REPO = os.path.join(BARE_PARENT, 'repo.git')
+WORK_INITIAL = os.path.join(BASE_DIR, 'work1')
+WORK_HTTP = os.path.join(BASE_DIR, 'work2')
+WORK_VERIFY = os.path.join(BASE_DIR, 'verify')
+CGI_DIR = os.path.join(BASE_DIR, 'cgi-bin')
+
+# set a fallback identity so commits don’t fail in CI
+subprocess.check_call(["git", "config", "--global", "user.email", "action@github.com"])
+subprocess.check_call(["git", "config", "--global", "user.name",  "GitHub Action"])
```
patzielinski, line 34: Can these be done as needed on a `--local` basis? This would help with running/testing these scripts on local machines without overwriting the global configuration.

.github/workflows/remote-tests-Gittuf.yml
zsun6: yeah this is the fix for ubuntu, it doesn't recognize git without this

.github/workflows/Git-transport-tests.py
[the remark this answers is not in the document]
zsun6: Yes, the only difference between the files is that the gittuf one has `gittuf::` infront of the url on line 96

.github/workflows/Git-transport-tests.py
zsun6: Sure

.github/workflows/Git-transport-tests.py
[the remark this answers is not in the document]
patzielinski: In this case, let's remove this file; we want to test the gittuf transport with these tests.

patzielinski:
@zsun6 bump on this, are you still working on this?

.github/workflows/Git-transport-tests.py
```diff
@@ -0,0 +1,119 @@
...
+
+def run(cmd, cwd=None):
+    print(f"[*] Running: {' '.join(cmd)}")
+    subprocess.check_call(cmd, cwd=cwd)
+
+
+def run_http_server(repo_root, port):
+    # Prepare CGI directory and link git-http-backend
+    os.makedirs(CGI_DIR, exist_ok=True)
+    backend = shutil.which('git-http-backend') or shutil.which('git-http-backend.exe')
```
adityasaky, line 44: Do we need this check? Isn't this part of the git binary? `git http-backend` may work?

PR #778, opened 2025-02-02
*: Extend TreeBuilder to reuse existing trees

adityasaky:
If an intermediate tree already exists and we know this ahead of time, we should use it without recreating it with all of its contents resolved and written recursively. This commit adds support for such cases and renames TreeBuilder's APIs to generalize away from accepting blob IDs only.


adityasaky:
This enables optimizations in #758.

PR #782, opened 2025-02-04
gitinterface: Add subtree workflow

adityasaky:
This is a first pass at adding the propagation pieces to gitinterface from #758.


adityasaky:
Note: this is ready for a first review but I'm looking into adding some guardrails for how this might impact the worktree.

PR #795, opened 2025-02-05
gittuf: Remove GetRSLLog from API

adityasaky:
This isn't used anymore since we updated how gittuf rsl log works.


adityasaky:
hmm, looks like the propagation tests may be slightly flakey...

PR #800, opened 2025-02-05
gitinterface: Use commit-graph when available

adityasaky:
This commit allows use of git's commit-graph when it's available. It fails silently when something is unavailable in the commit-graph, falling back to the default mechanisms.


adityasaky:
closing, let's revisit this later

PR #808, opened 2025-02-07
Load controller metadata into policy and verify roots match

adityasaky:


adityasaky:
Everything here seems to have landed independently, closing...

PR #812, opened 2025-02-10
tuf: Propagate changes to both policy refs

adityasaky:
Part of #808


adityasaky:
This approach is going to be problematic because it'll cause policy and staging to diverge, making future Apply invocations fail. Working on an alternative solution, which I'll submit separately.

PR #813, opened 2025-02-10
[GUPPIEs] lua-sandbox: Add a lua sandbox for running GUPPIEs

patzielinski:
This PR adds the lua sandbox environment to gittuf, in order to be able to run GUPPIEs in an isolated environment.

Part of a series, more to come.

cc @fr0m-scratch

### DIFF
--- a/.test_ignore.txt
+++ b/.test_ignore.txt
@@ -2,6 +2,7 @@ github.com/gittuf/gittuf/docs/cli
 github.com/gittuf/gittuf/experimental/gittuf/options
 github.com/gittuf/gittuf/internal/cmd
 github.com/gittuf/gittuf/internal/dev
+github.com/gittuf/gittuf/internal/luasandbox
 github.com/gittuf/gittuf/internal/git-remote-gittuf
 github.com/gittuf/gittuf/internal/testartifacts
 github.com/gittuf/gittuf/internal/third_party
--- a/internal/luasandbox/main.go
+++ b/internal/luasandbox/main.go
@@ -1,0 +1,166 @@
+// Copyright The gittuf Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// This file contains modified code from the lua-sandbox project, available at
+// https://github.com/kikito/lua-sandbox/blob/master/sandbox.lua, and licensed
+// under the MIT License
+
+package luasandbox
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	lua "github.com/yuin/gopher-lua"
+)
+
+const (
+	// Quota from https://mozilla-services.github.io/lua_sandbox/sandbox.html
+	luaInstructionQuota = 1000000
+	luaTimeOut          = 100
+)
+
+// NewLuaEnvironment creates a new Lua state with the specified modules
+// enabled.
+func NewLuaEnvironment(ctx context.Context) (*lua.LState, error) {
+	// Create a new Lua state
+	lState := lua.NewState(lua.Options{SkipOpenLibs: true})
+
+	// Load default safe libraries
+	modules := []struct {
+		n string
+		f lua.LGFunction
+	}{
+		{lua.LoadLibName, lua.OpenPackage}, // Must be first
+		{lua.BaseLibName, lua.OpenBase},
+		{lua.TabLibName, lua.OpenTable},
+		{lua.StringLibName, lua.OpenString},
+		{lua.MathLibName, lua.OpenMath},
+		{lua.CoroutineLibName, lua.OpenCoroutine},
+	}
+
+	// Load the modules to the Lua state
+	for _, pair := range modules {
+		if err := lState.CallByParam(lua.P{
+			Fn:      lState.NewFunction(pair.f),
+			NRet:    0,
+			Protect: true,
+		}, lua.LString(pair.n)); err != nil {
+			panic(err)
+		}
+	}
+
+	// Enable only safe functions
+	enableOnlySafeFunctions(lState)
+
+	// Set the instruction quota and timeout
+	setTimeOut(ctx, lState, luaTimeOut)
+	if err := setInstructionQuota(lState, luaInstructionQuota); err != nil {
+		return nil, fmt.Errorf("error setting instruction quota: %w", err)
+	}
+
+	// TODO: Register API functions
+
+	return lState, nil
+}
+
+// enableOnlySafeFunctions disables all functions that are deemed to be unsafe.
+func enableOnlySafeFunctions(l *lua.LState) {
+	//-- List of unsafe packages/functions:
+	// -- * string.rep: can be used to allocate millions of bytes in 1 operation
+	// -- * {set|get}metatable: can be used to modify the metatable of global objects (strings, integers)
+	// -- * collectgarbage: can affect performance of other systems
+	// -- * dofile: can access the server filesystem
+	// -- * _G: It has access to everything. It can be mocked to other things though.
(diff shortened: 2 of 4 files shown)


internal/lua-sandbox/api/api.go
adityasaky, line 1: no hyphens in package names

internal/lua-sandbox/api/api.go
```diff
@@ -0,0 +1,575 @@
...
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	sandbox "github.com/gittuf/gittuf/internal/lua-sandbox/util"
```
adityasaky, line 24: sandbox or sandboxutil?

internal/lua-sandbox/api/api.go
```diff
@@ -0,0 +1,575 @@
...
+	lua "github.com/yuin/gopher-lua"
+)
+
+// Lua sandbox helper functions and wrapper functions of provided Go APIs
+var luaHelpers = `
```
adityasaky, line 29: maybe make this a map of helper name -> helper implementation, then range through them and register them? seems cleaner to extend over time.

internal/lua-sandbox/api/api.go
```diff
@@ -0,0 +1,575 @@
...
+	lState.SetGlobal("hookParameters", lua.LString(""))
+	lState.SetGlobal("hookExitCode", lua.LNumber(0))
+	lState.SetGlobal("allowedExecutables", lua.LString(strings.Join(allowedModules, ",")))
+
+	return lState, nil
+}
+
+// Return all filenames in the specified directory, take a second argument to
+// specify if the scan should be recursive
+func goScanDir(l *lua.LState) int {
```
adityasaky, line 102: move go implementations to a separate file?

internal/lua-sandbox/api/api.go
```diff
@@ -0,0 +1,575 @@
+// Copyright The gittuf Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// This file contains modified code from the lua-sandbox project, available at
+// https://github.com/kikito/lua-sandbox/blob/master/sandbox.lua, and licensed
+// under the MIT License
+
+package sandbox
```
adityasaky, line 8: package api?

internal/lua-sandbox/main.go
adityasaky, line 1: main.go usually implies a `main` package

adityasaky: should we consolidate this and the api package? unsure why they're different

internal/lua-sandbox/util/util.go
```diff
@@ -0,0 +1,129 @@
+// Copyright The gittuf Authors
```
adityasaky, line 1: same here, maybe a single luasandbox package is all we really need. Organization within a package in separate files is fine.

internal/lua-sandbox/api/api.go
```diff
@@ -0,0 +1,575 @@
...
+    debug.sethook(function()
+        count = count + 1
+        if count > quota then
+            error("Instruction quota exceeded (" .. quota .. " instructions).", 2)
+        end
+    end, "", 1)
+end
+
+function goLinter(path) return runFunc("goLinter", path, nil) end
+function removeFile(path) return deleteFile(path) end
```
adityasaky, line 55: we're effectively going to provide both `removeFile` and `deleteFile` to a user. Let's get rid of that duplication.

internal/lua-sandbox/api/api.go
adityasaky: actually, let's have this PR be simpler and add a very basic sandbox with next to no APIs. With all of these APIs in this PR, we're not going to do justice in our review

internal/lua-sandbox/api/api.go
```diff
@@ -0,0 +1,575 @@
...
+		l.Push(lua.LString(err.Error()))
+		return 1
+	}
+
+	l.Push(lua.LString(string(output)))
+	return 1
+}
+
+// goRegexMatch processes input text and searches for patterns
+func goRegexMatch(l *lua.LState) int {
```
adityasaky, line 187: we want to reason about all of these APIs carefully; they're a big part of what we allow and don't allow in the sandbox. I suggest having only the most basic ones in this PR like a split string, regex match, and move the others to a follow on PR.

internal/lua-sandbox/util/util.go
fr0m-scratch: will do!

internal/lua-sandbox/main.go
fr0m-scratch: My current plan is to have only sandbox initialization functions and security configuration helpers in `main`, and keep all API (either pure Lua helpers or Lua wrappers of go implementations) exposed to the users in `api`.

adityasaky: `package main` usually means it's an application package, for example, a command. We have a couple of those ourselves. All of these packages are purely library packages if I'm not wrong.

Also, I don't know that we need to make the API functions available via a separate package, ultimately, the users are not consuming them via the go package at all (in fact it's in `internal`)...

fr0m-scratch: Yes, totally agree. I already made them into the same package `luasandbox`. It's just haven't been linked to this PR yet. By that, I meant to keep initialization and API functions in different files. Sorry for the confusion. 

Can you help me check the latest commit here? https://github.com/fr0m-scratch/gittuf/commit/e0897bab39af764df26dfa6c5260df93d2ea9817

adityasaky: cc @patzielinski

internal/lua-sandbox/api.go
adityasaky, line 1: package name still contains a hyphen, please rename the directory

.test_ignore.txt
```diff
@@ -2,6 +2,7 @@ github.com/gittuf/gittuf/docs/cli
 github.com/gittuf/gittuf/experimental/gittuf/options
 github.com/gittuf/gittuf/internal/cmd
 github.com/gittuf/gittuf/internal/dev
+github.com/gittuf/gittuf/internal/lua-sandbox
```
adityasaky, line 5: and update this to match :)

adityasaky (changes requested):

internal/luasandbox/util.go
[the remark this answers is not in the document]
adityasaky: yes please

internal/luasandbox/main.go
```diff
@@ -0,0 +1,161 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	lua "github.com/yuin/gopher-lua"
+)
+
+// NewLuaEnvironment creates a new Lua  with the specified environment
+func NewLuaEnvironment(allowedModules []string) (*lua.LState, error) {
```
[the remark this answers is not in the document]
adityasaky: Let's start simple with a base set of modules.

internal/luasandbox/main.go
```diff
@@ -0,0 +1,161 @@
...
+			panic(err)
+		}
+	}
+
+	// Enable only safe functions
+	enableOnlySafeFunctions(lState)
+
+	// Set the instruction quota and timeout
+	setTimeOut(context.Background(), lState, 600)
+	if err := setInstructionQuota(lState, 300000000); err != nil {
```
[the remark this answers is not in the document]
adityasaky: yeah, it feels very magical rn

adityasaky (approved):

PR #815, opened 2025-02-10
[GUPPIEs] *: Add GUPPIEs to gittuf metadata

patzielinski:
This PR updates the gittuf metadata to include support for defining GUPPIEs.

Part of a series, more to come.

### DIFF
--- a/internal/tuf/tuf.go
+++ b/internal/tuf/tuf.go
@@ -4,6 +4,7 @@
 package tuf
 
 import (
+	"encoding/json"
 	"errors"
 
 	"github.com/gittuf/gittuf/internal/common/set"
@@ -29,6 +30,11 @@ const (
 	GlobalRuleThresholdType        = "threshold"
 	GlobalRuleBlockForcePushesType = "block-force-pushes"
 	RemoveGlobalRuleType           = "remove"
+
+	HookStagePreCommitString = "preCommit"
+	HookStagePrePushString   = "prePush"
+
+	HookEnvironmentLuaString = "lua"
 )
 
 var (
@@ -52,6 +58,11 @@ var (
 	ErrGlobalRuleAlreadyExists                         = errors.New("global rule already exists")
 	ErrPropagationDirectiveNotFound                    = errors.New("specified propagation directive not found")
 	ErrNotAControllerRepository                        = errors.New("current repository is not marked as a controller repository")
+	ErrDuplicatedHookName                              = errors.New("two hooks with same name found in policy")
+	ErrInvalidHookStage                                = errors.New("invalid stage for hook")
+	ErrInvalidHookEnvironment                          = errors.New("invalid environment for hook")
+	ErrHookNotFound                                    = errors.New("cannot find hook entry")
+	ErrNoHooksDefined                                  = errors.New("no hooks defined")
 )
 
 // Principal represents an entity that is granted trust by gittuf metadata. In
@@ -177,6 +188,18 @@ type RootMetadata interface {
 	// network for which the current repository is a controller.
 	// IsController must return true for this to be set.
 	GetNetworkRepositories() []OtherRepository
+
+	// AddHook adds to the metadata for the specified Git stage a hook named
+	// hookName to be run by the specified principals. For support of more than
+	// one hashing algorithm, providing multiple hashes is supported. The hook's
+	// environment (e.g. lua) and modules to expose to the hook (if using Lua)
+	// are also required.
+	AddHook(stage HookStage, hookName string, principalIDs []string, hashes map[string]string, environment HookEnvironment, modules []string) error
+	// RemoveHook removes the hook identified by hookName in the specified Git
+	// stage.
+	RemoveHook(stage HookStage, hookName string)
+	// GetHooks returns all hooks in the metadata for the specified Git stage.
+	GetHooks(stage HookStage) ([]Hook, error)
 }
 
 // TargetsMetadata represents gittuf's rule files. Its name is inspired by TUF.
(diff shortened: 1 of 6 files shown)


internal/tuf/tuf.go
```diff
@@ -177,6 +181,13 @@ type RootMetadata interface {
 	// network for which the current repository is a controller.
 	// IsController must return true for this to be set.
 	GetNetworkRepositories() []OtherRepository
+
+	// AddHook adds a hook to the metadata file.
+	AddHook(stage, hookName, env string, hashes map[string]gitinterface.Hash, modules, principalIDs []string) error
```
adityasaky, line 186: Can you add what the parameters do to the docstring? `env` for example isn't very self describing.

adityasaky: Are we going with hook? I wonder about using something else in the interface to avoid collision with git hooks.

internal/tuf/v01/root.go
```diff
@@ -29,6 +30,7 @@ type RootMetadata struct {
 	GlobalRules            []tuf.GlobalRule           `json:"globalRules,omitempty"`
 	Propagations           []tuf.PropagationDirective `json:"propagations,omitempty"`
 	MultiRepository        *MultiRepository           `json:"multiRepository,omitempty"`
+	Hooks                  Hooks                      `json:"hooks"`
```
adityasaky, line 33: omitempty should make the dsse test pass without an edit

internal/tuf/v01/root.go
```diff
@@ -784,3 +786,108 @@ func (o *OtherRepository) GetInitialRootPrincipals() []tuf.Principal {
 	}
 	return initialRootPrincipals
 }
+
+// Hooks defines the schema that groups hooks by stage.
+type Hooks struct {
+	PreCommit map[string]Hook `json:"pre-commit"`
```
adityasaky, line 792:
```suggestion
	PreCommit map[string]*Hook `json:"pre-commit"`
```

adityasaky: Also don't love the use of `-` in the json tag, maybe we can camelcase it `preCommit` for consistency?

internal/tuf/v01/root.go
```diff
@@ -784,3 +786,108 @@ func (o *OtherRepository) GetInitialRootPrincipals() []tuf.Principal {
...
+	newHook := Hook{
+		Name:         hookName,
+		PrincipalIDs: set.NewSetFromItems(principalIDs...),
+		Hashes:       hashes,
+		Environment:  env,
+		Modules:      modules,
+	}
+
+	switch stage {
+	case "pre-commit":
```
adityasaky, line 809: could you put this in `tuf` as consts?

adityasaky: `const PreCommitStageName = "pre-commit"` etc

internal/tuf/v01/root.go
```diff
@@ -784,3 +786,108 @@ func (o *OtherRepository) GetInitialRootPrincipals() []tuf.Principal {
...
+		originMap = r.Hooks.PreCommit
+	case "pre-push":
+		originMap = r.Hooks.PrePush
+	default:
+		return nil, tuf.ErrInvalidHookStage
+	}
+
+	var appletMap = make(map[string]tuf.Applet)
```
adityasaky, line 852:
```suggestion
	var appletMap = make(map[string]tuf.Applet, len(originMap)
```

internal/tuf/v01/root.go
```diff
@@ -784,3 +786,108 @@ func (o *OtherRepository) GetInitialRootPrincipals() []tuf.Principal {
...
+func (h *Hook) ID() string {
+	return h.Name
+}
+
+// GetPrincipalIDs returns the principals that must run this hook.
+func (h *Hook) GetPrincipalIDs() *set.Set[string] {
+	return h.PrincipalIDs
+}
+
+// GetHashes returns the hashes of the hook file.
```
adityasaky, line 880: Is this to enable hash agility for the hook file? Having `gitinterface.Hash` as a target type might be blocking if so.

adityasaky (changes requested):

internal/tuf/tuf.go
```diff
@@ -29,6 +30,9 @@ const (
 	GlobalRuleThresholdType        = "threshold"
 	GlobalRuleBlockForcePushesType = "block-force-pushes"
 	RemoveGlobalRuleType           = "remove"
+
+	PreCommitStageName = "pre-commit"
```
adityasaky, line 34: Thinking about it more, I'd setup a stage type so we can better validate what we see.

something like

```
type HookStage uint

const (
    PreCommit HookStage = iota
    ...
)

func (h HookStage) String() string {...}
```

internal/tuf/tuf.go
adityasaky: With my proposal for making stage a type, we could pass in `HookStage` type here and improve validation. The `Hooks` structs could also just use stage as the internal map key.

internal/tuf/tuf.go
```diff
@@ -338,3 +357,22 @@ type OtherRepository interface {
 	// sign the other repository's initial gittuf root of trust metadata.
 	GetInitialRootPrincipals() []Principal
 }
+
+// Applet represents a hook entry in a rule file ('TargetsMetadata').
+type Applet interface {
```
adityasaky, line 362: I feel like we should at least align the interface methods (AddHook) with this interface...

internal/tuf/v01/root.go
adityasaky: I would probably make it just a string, to store hex encoded representations. We should look to upstream TUF implementations here.

internal/tuf/v01/root.go
```diff
@@ -29,6 +29,7 @@ type RootMetadata struct {
 	GlobalRules            []tuf.GlobalRule           `json:"globalRules,omitempty"`
 	Propagations           []tuf.PropagationDirective `json:"propagations,omitempty"`
 	MultiRepository        *MultiRepository           `json:"multiRepository,omitempty"`
+	Hooks                  *Hooks                     `json:"hooks,omitempty"`
```
adityasaky, line 32: Can you add this to migrations too?

internal/tuf/v02/root.go
```diff
@@ -757,3 +758,112 @@ func (o *OtherRepository) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// Hooks defines the schema that groups hooks by stage.
+type Hooks struct {
```
adityasaky, line 763: Do we need Hooks to be a wrapper type? It could just be `map[tuf.HookStage]map[string]*Hook` right? Or even `map[tuf.HookStage][]*Hook` since each hook also has a notion of its name.

adityasaky: This way, we can also look up directly using a tuf.HookStage const without having to rely on `.String()` at lookup time. `tuf.HookStage` may need a custom json marshal / unmarshal though, if we want it to be readable in the metadata.

internal/tuf/tuf.go
```diff
@@ -338,3 +354,42 @@ type OtherRepository interface {
 	// sign the other repository's initial gittuf root of trust metadata.
 	GetInitialRootPrincipals() []Principal
 }
+
+// HookStage encodes the Git stage at which a hook is to run
+type HookStage uint
+
+const (
+	PreCommit HookStage = iota
```
adityasaky, line 362:
```suggestion
	HookStagePreCommit HookStage = iota
```

nit: helpful to avoid collisions imo

internal/tuf/v01/root.go
```diff
@@ -784,3 +788,96 @@ func (o *OtherRepository) GetInitialRootPrincipals() []tuf.Principal {
 	}
 	return initialRootPrincipals
 }
+
+// AddHook adds the specified hook to the metadata.
+func (r *RootMetadata) AddHook(stage tuf.HookStage, hookName string, principalIDs []string, hashes map[string]string, environment string, modules []string) error {
+	// TODO: Check if principal actually exists in TargetsMetadata?
```
adityasaky, line 794: yeah, this is pretty important...

internal/tuf/tuf.go
```diff
@@ -177,6 +182,18 @@ type RootMetadata interface {
 	// network for which the current repository is a controller.
 	// IsController must return true for this to be set.
 	GetNetworkRepositories() []OtherRepository
+
+	// AddHook adds to the metadata for the specified Git stage a hook named
+	// hookName to be run by the specified principals. For support of more than
+	// one hashing algorithm, providing multiple hashes is supported. The hook's
+	// environment (e.g. lua) and modules to expose to the hook (if using Lua)
+	// are also required.
+	AddHook(stage HookStage, hookName string, principalIDs []string, hashes map[string]string, environment string, modules []string) error
```
adityasaky, line 191: what environment options do we start with? do we want to enum these too?

adityasaky (commented):
We can likely merge this very soon.

internal/tuf/v01/root.go
adityasaky: shouldn't it be RootMetadata btw?

internal/tuf/tuf.go
adityasaky: Do we need the `local` to be merged / supported at all?

internal/tuf/v01/root.go
```diff
@@ -784,3 +788,96 @@ func (o *OtherRepository) GetInitialRootPrincipals() []tuf.Principal {
...
+		PrincipalIDs: set.NewSetFromItems(principalIDs...),
+		Hashes:       hashes,
+		Environment:  environment,
+		Modules:      modules,
+	}
+
+	if r.Hooks == nil {
+		r.Hooks = make(map[tuf.HookStage][]*Hook, 2)
+	}
+
```
adityasaky, line 807:
```
if r.Hooks[stage] == nil {
    r.Hooks[stage] = []*Hook{}
}
```

adityasaky: Do we need the switch here at all? You could also just append based on `stage`?

adityasaky (approved):

Issue #804, opened 2025-02-07
Add duplicate check for various multi-repository flows


patzielinski:
#802 adds support for multi-repository policies. Various flows for adding the required metadata should have checks for duplicates added to them, e.g. https://github.com/gittuf/gittuf/pull/802/files#diff-c1f6a308a1f9a2761d8dc740a22716239fbce5f6a6018ebb99b39c8b070046fbR445/

PR #820, opened 2025-02-11
added Duplication check

Horiodino:
fixes: https://github.com/gittuf/gittuf/issues/804#issuecomment-2648739175

### DIFF
--- a/internal/tuf/v01/root.go
+++ b/internal/tuf/v01/root.go
@@ -483,7 +483,35 @@ func (r *RootMetadata) AddControllerRepository(name, location string, initialRoo
 		r.MultiRepository = &MultiRepository{ControllerRepositories: []*OtherRepository{}}
 	}
 
-	// TODO: check for duplicates
+	for _, repo := range r.MultiRepository.ControllerRepositories {
+		if repo.Name == name || repo.Location == location {
+			return tuf.ErrDuplicateControllerRepository
+		}
+	}
+
+	newKeyIDs := make([]tuf.Principal, 0, len(initialRootPrincipals))
+	for _, principal := range initialRootPrincipals {
+		key, isKey := principal.(*Key)
+		if !isKey {
+			return tuf.ErrInvalidPrincipalType
+		}
+		newKeyIDs = append(newKeyIDs, key)
+	}
+
+	newKeyIDsSet := set.NewSet[string]()
+	for _, principal := range newKeyIDs {
+		newKeyIDsSet.Add(principal.ID())
+	}
+
+	for _, repo := range r.MultiRepository.ControllerRepositories {
+		existingKeyIDs := set.NewSet[string]()
+		for _, existingPrincipal := range repo.InitialRootPrincipals {
+			existingKeyIDs.Add(existingPrincipal.KeyID)
+		}
+		if existingKeyIDs.Equal(newKeyIDsSet) {
+			return tuf.ErrDuplicateControllerRepository
+		}
+	}
 
 	otherRepository := &OtherRepository{
 		Name:                  name,
@@ -492,11 +520,7 @@ func (r *RootMetadata) AddControllerRepository(name, location string, initialRoo
 	}
 
 	for _, principal := range initialRootPrincipals {
-		key, isKey := principal.(*Key)
-		if !isKey {
-			return tuf.ErrInvalidPrincipalType
-		}
-
+		key := principal.(*Key)
 		otherRepository.InitialRootPrincipals = append(otherRepository.InitialRootPrincipals, key)
 	}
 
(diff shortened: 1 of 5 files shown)


Horiodino:
cc @patzielinski

internal/tuf/v02/root_test.go
```diff
@@ -148,6 +152,24 @@ func TestRootMetadata(t *testing.T) {
 		networkRepositories = rootMetadata.GetNetworkRepositories()
 		assert.Nil(t, networkRepositories)
 	})
+
+	t.Run("test AddNetworkRepository with duplicates", func(t *testing.T) {
```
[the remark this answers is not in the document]
Horiodino: Okay, I've made the changes, but what's the point? It seems like it was the same as before.

internal/tuf/v01/root.go
```diff
@@ -457,7 +455,16 @@ func (r *RootMetadata) AddControllerRepository(name, location string, initialRoo
...
+		for _, repo := range r.MultiRepository.ControllerRepositories {
+			if repo.Name == name || repo.Location == location {
+				return tuf.ErrDuplicateControllerRepository
+			}
+			for _, existingPrincipal := range repo.InitialRootPrincipals {
+				if existingPrincipal.KeyID == key.KeyID {
+					return tuf.ErrDuplicateControllerRepository
+				}
+			}
+		}
```
patzielinski, line 467: This should be checking if all root principals are the same, not if any are the same.

internal/tuf/v01/root.go
```diff
@@ -497,6 +502,16 @@ func (r *RootMetadata) AddNetworkRepository(name, location string, initialRootPr
...
 
+		for _, repo := range r.MultiRepository.NetworkRepositories {
+			if repo.Name == name || repo.Location == location {
+				return tuf.ErrDuplicateNetworkRepository
+			}
+			for _, existingPrincipal := range repo.InitialRootPrincipals {
+				if existingPrincipal.KeyID == key.KeyID {
+					return tuf.ErrDuplicateNetworkRepository
+				}
+			}
```
patzielinski, line 513: Same as above.

internal/tuf/v01/root.go
```diff
@@ -444,27 +444,44 @@ func (r *RootMetadata) AddControllerRepository(name, location string, initialRoo
...
 		}
+		newKeyIDs[key.KeyID] = struct{}{}
+	}
+
+	for _, repo := range r.MultiRepository.ControllerRepositories {
+		existingKeyIDs := make(map[string]struct{}, len(repo.InitialRootPrincipals))
+		for _, existingPrincipal := range repo.InitialRootPrincipals {
+			existingKeyIDs[existingPrincipal.KeyID] = struct{}{}
+		}
+		if mapsEqual(newKeyIDs, existingKeyIDs) {
```
patzielinski, line 467: I think https://pkg.go.dev/reflect#DeepEqual can be used here?

internal/tuf/v01/root.go
```diff
@@ -444,27 +444,44 @@ func (r *RootMetadata) AddControllerRepository(name, location string, initialRoo
...
-		Name:                  name,
-		Location:              location,
-		InitialRootPrincipals: make([]*Key, 0, len(initialRootPrincipals)),
+	for _, repo := range r.MultiRepository.ControllerRepositories {
+		if repo.Name == name || repo.Location == location {
+			return tuf.ErrDuplicateControllerRepository
+		}
 	}
 
+	newKeyIDs := make(map[string]struct{}, len(initialRootPrincipals))
```
patzielinski, line 453: Could a slice of `tuf.Principal`s be used here instead of a `map[string]struct{}`?

internal/tuf/v01/root.go
```diff
@@ -444,27 +445,44 @@ func (r *RootMetadata) AddControllerRepository(name, location string, initialRoo
...
 	for _, principal := range initialRootPrincipals {
 		key, isKey := principal.(*Key)
 		if !isKey {
 			return tuf.ErrInvalidPrincipalType
 		}
+		newKeyIDs = append(newKeyIDs, key)
+	}
+
+	for _, repo := range r.MultiRepository.ControllerRepositories {
+		existingKeyIDs := make(map[string]struct{}, len(repo.InitialRootPrincipals))
```
adityasaky, line 464: can we just use a `set` here? We have one included and I think it even may handle set equal comparison...

adityasaky: https://github.com/gittuf/gittuf/blob/main/internal/common/set/set.go#L14-L16

internal/tuf/v01/root.go
```diff
@@ -445,37 +445,52 @@ func (r *RootMetadata) AddControllerRepository(name, location string, initialRoo
...
+		key := principal.(*Key)
+		otherRepository.InitialRootPrincipals = append(otherRepository.InitialRootPrincipals, key)
 	}
-	return r.AddPropagationDirective(NewPropagationDirective(policyPropagationName, location, "refs/gittuf/policy", "refs/gittuf/policy", policyPropagationLocation))
+
+	r.MultiRepository.ControllerRepositories = append(r.MultiRepository.ControllerRepositories, otherRepository)
+
+	propagationName := fmt.Sprintf("%s-%s", tuf.GittufControllerPrefix, name)
+	propagationLocation := path.Join(tuf.GittufControllerPrefix, name)
+	return r.AddPropagationDirective(NewPropagationDirective(propagationName, location, "refs/gittuf/policy", "refs/gittuf/policy", propagationLocation))
```
patzielinski, line 493: I think the change to how propagation directives are added (https://github.com/gittuf/gittuf/blob/main/internal/tuf/v01/root.go#L469-L478) isn't reflected here, which is probably why tests are failing. Can you see if adding the new set of propagation directives fixes this?

patzielinski (commented):
Apologies for the delay. I've flagged what I think is causing tests to fail. Once this is fixed, I _think_ we'll be able to merge this.

Thanks again!

internal/tuf/v01/root.go
Horiodino: ok

internal/tuf/v01/root.go
```diff
@@ -483,27 +483,49 @@ func (r *RootMetadata) AddControllerRepository(name, location string, initialRoo
...
+
+	for _, principal := range initialRootPrincipals {
+		key := principal.(*Key)
 		otherRepository.InitialRootPrincipals = append(otherRepository.InitialRootPrincipals, key)
 	}
 
 	r.MultiRepository.ControllerRepositories = append(r.MultiRepository.ControllerRepositories, otherRepository)
 
-	// Add the controller as a repository whose policy contents must be
-	// propagated into this repository
```
patzielinski, line 506: Let's restore these comment lines if possible.

patzielinski (commented):
LGTM. Please trim the commit message to just one blurb (e.g. consolidate the multiple messages into one. Thanks!

patzielinski (approved):
Thanks, @Horiodino!

PR #830, opened 2025-02-17
Hook attestations v0.01

Yasho-Bapat:
This commit adds a skeleton of a basic hook attestations implementation. The logs and basic println debugging shows that it reaches the stage of committing an attestation, however, upon checking `refs/gittuf/attestations`, it is empty. 

Attempting to solve this atm, but also @patzielinski look into this pls considering our timeline.


adityasaky:
It's https://github.com/gittuf/gittuf/pull/830/files#diff-f2cb870f063f11bd2506a144d24854d610fbc91d0b4ea49a19dfa9152694937fR181 I think? You need to add it to the attestations tree when committing.

Yasho-Bapat:
Fixed it now. This should be a minimal working hooks attestation flow

PR #832, opened 2025-02-17
ci: Add go 1.24

patzielinski:
https://go.dev/blog/go1.24


adityasaky:
So why are we dropping coverage on a go version bump? :thinking:

adityasaky (approved):

adityasaky:
@patzielinski should the ignores file be updated?

PR #839, opened 2025-02-18
[GUPPIEs] *: Add management functionality for GUPPIEs

patzielinski:
This PR adds functionality to manage GUPPIEs.

~~Blocked by #815.~~

### DIFF
--- a/experimental/gittuf/root.go
+++ b/experimental/gittuf/root.go
@@ -5,14 +5,18 @@ package gittuf
 
 import (
 	"context"
+	"crypto/sha256"
 	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log/slog"
 
 	"github.com/gittuf/gittuf/experimental/gittuf/options/root"
 	trustpolicyopts "github.com/gittuf/gittuf/experimental/gittuf/options/trustpolicy"
 	"github.com/gittuf/gittuf/internal/dev"
+	"github.com/gittuf/gittuf/internal/gitinterface"
 	"github.com/gittuf/gittuf/internal/policy"
 	policyopts "github.com/gittuf/gittuf/internal/policy/options/policy"
 	"github.com/gittuf/gittuf/internal/signerverifier/common"
@@ -24,6 +28,10 @@ import (
 	tufv01 "github.com/gittuf/gittuf/internal/tuf/v01"
 	tufv02 "github.com/gittuf/gittuf/internal/tuf/v02"
 	"github.com/secure-systems-lab/go-securesystemslib/signerverifier"
+)
+
+var (
+	ErrNoHookName = errors.New("hook name not provided")
 )
 
 // InitializeRoot is the interface for the user to create the repository's root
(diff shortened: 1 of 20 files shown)


internal/cmd/trust/trust.go
```diff
@@ -47,6 +48,7 @@ func New() *cobra.Command {
 	cmd.AddCommand(apply.New())
 	cmd.AddCommand(disablegithubappapprovals.New(o))
 	cmd.AddCommand(enablegithubappapprovals.New(o))
+	cmd.AddCommand(hooks.New(o))
```
adityasaky, line 51: for consistency, this should probably be `gittuf trust add-hook` etc instead of `gittuf trust hooks ...`.

experimental/gittuf/root.go
```diff
@@ -776,6 +782,166 @@ func (r *Repository) RemovePropagationDirective(ctx context.Context, signer ssli
...
+
+	rootMetadata, err := r.loadRootMetadata(state, rootKeyID)
+	if err != nil {
+		return err
+	}
+
+	slog.Debug("Checking if hook with specified name exists...")
+	hooks := []tuf.Hook{}
+	found := false
+	for _, hook := range state.Hooks[stage] {
```
adityasaky, line 921: this can panic if invoked on a stage that doesn't exist, unless you're serializing an empty list for each stage?

adityasaky: fwiw, we shouldn't be serializing empty lists.

experimental/gittuf/root.go
```diff
@@ -776,6 +782,166 @@ func (r *Repository) RemovePropagationDirective(ctx context.Context, signer ssli
 	return r.updateRootMetadata(ctx, state, signer, rootMetadata, commitMessage, options.CreateRSLEntry, signCommit)
 }
 
+// AddHook defines the workflow for adding a file to be executed as a hook. It
+// writes the hook file, populates all fields in the hooks metadata associated
+// with this file and commits it to the root of trust metadata.
+func (r *Repository) AddHook(ctx context.Context, signer sslibdsse.SignerVerifier, stage tuf.HookStage, hookName, filePath string, environment tuf.HookEnvironment, modules, principalIDs []string, signCommit bool, opts ...trustpolicyopts.Option) error {
```
adityasaky, line 788: I don't love that this is in a random file on disk somewhere...

adityasaky: I wonder if the API should take an io.Reader or []byte so we can build better UX later. For now, maybe the cmd package passes an os.File or its contents, but later, we could do neat things like dropping them into a temporary worktree where they can edit the programs with their editor of choice, like any other source file.

adityasaky: This should also make testing easier...

internal/policy/list.go
```diff
@@ -102,3 +102,14 @@ func ListPrincipals(ctx context.Context, repo *gitinterface.Repository, targetRe
 
 	return metadata.GetPrincipals(), nil
 }
+
+// ListHooks returns the hooks present in the specified rule file.
+// `targetRef` can be used to control which policy reference is used.
+func ListHooks(ctx context.Context, repo *gitinterface.Repository) (map[tuf.HookStage][]tuf.Hook, error) {
```
adityasaky, line 108: Do we need this to live in `policy`? We're directly interacting with `tuf` in `experimental/gittuf` now, so maybe this logic doesn't need to live here? Same for the other `List...` APIs.

adityasaky (changes requested):

internal/policy/list.go
patzielinski: https://github.com/gittuf/gittuf/pull/865

internal/cmd/trust/addhook/addhook.go
```diff
@@ -0,0 +1,144 @@
...
+		return tuf.ErrInvalidHookEnvironment
+	}
+
+	if (strings.ToLower(o.env) == "lua") && (o.modules == nil) {
+		return ErrLuaNoModules
+	}
+
+	var stage tuf.HookStage
+	switch strings.ToLower(o.stage) {
+	case "pre-commit":
```
adityasaky, line 100: do we have this in consts smewhere?

internal/cmd/trust/addhook/addhook.go
```diff
@@ -0,0 +1,144 @@
...
+		"",
+		"filepath of the script to be run as a hook",
+	)
+	cmd.MarkFlagRequired("file") //nolint:errcheck
+
+	cmd.Flags().StringVarP(
+		&o.stage,
+		"stage",
+		"s",
+		"",
```
adityasaky, line 46: Can we have a custom enum-esque type for this? I also wonder if a flag per stage is too extreme; e.g., `--is-pre-commit --is-pre-push` or so

internal/cmd/trust/addhook/addhook.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	return repo.AddHook(cmd.Context(), signer, stage, o.hookName, hookBytes, environment, o.modules, o.principalIDs, true, opts...)
+}
+
+func New(persistent *persistent.Options) *cobra.Command {
+	o := &options{p: persistent}
+	cmd := &cobra.Command{
+		Use:               "add-hook",
+		Short:             "Add a script to be run as a gittuf hook, specify when and where to run it.",
```
adityasaky, line 135: note dev mode pls

internal/tuf/tuf.go
```diff
@@ -381,8 +383,32 @@ func (h HookStage) String() string {
 	}
 }
 
+// MarshalText is used to convert the instance of HookStage into text. Needed
+// for proper marshalling into JSON as HookStage is a key in a map.
+func (h HookStage) MarshalText() (text []byte, err error) {
```
adityasaky, line 388: can you explain this vs MarshalJSON() pls?

patzielinski: This is apparently needed to properly marshal the keys of the map as text (e.g. `preCommit`, `prePush`) instead of the underlying uint enum values. Took some trial and error and looking at the metadata introduced by #815.

See https://stackoverflow.com/questions/52161555/how-to-custom-marshal-map-keys-in-json

experimental/gittuf/root.go
```diff
@@ -776,6 +785,153 @@ func (r *Repository) RemovePropagationDirective(ctx context.Context, signer ssli
...
+
+	rootMetadata, err := r.loadRootMetadata(state, rootKeyID)
+	if err != nil {
+		return err
+	}
+
+	slog.Debug("Checking if hook with specified name exists...")
+	hooks := []tuf.Hook{}
+	found := false
+	for _, hook := range state.Hooks[stage] {
```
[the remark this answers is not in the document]
patzielinski: Updated this to use `rootMetadata.GetHooks(stage)`.

internal/cmd/trust/addhook/addhook.go
patzielinski: A flag per stage should be fine, at least for now. We can always update this later if we need to...

patzielinski: I've added the flags that each control a bool that's then used in a `switch`, I think this is fine? Didn't see a way to enum-ify the command flag, unless I misunderstood.

internal/tuf/tuf.go
patzielinski: I don't think the marshaller needs a pointer receiver? Metadata looks to be correct after running this on a test repo... Also see linked example.

internal/cmd/trust/addhook/addhook.go
fr0m-scratch: I saw in https://github.com/gittuf/gittuf/pull/861 we added all hook stages to the tuf metedata. I wonder can we move those changes here and instead of using a bool flag, can we use the string flag?

adityasaky: Between the two, I prefer the current approach just because of ease of use. Having them be bool flags is better for tab autocompletion, for example.

Issue #792, opened 2025-02-05
Drop support for online verification for sigstore signatures


adityasaky:
### What happened?

In gittuf, DSSE signatures using sigstore rely on online rekor lookups for private sigstore instances, which the sigstore community is planning to drop support for. We're currently frozen on older versions of sigstore-go and cosign until we can resolve this.

This likely does affect git commit and tag signatures via gitsign too, but we can incorporate those changes as gitsign is updated upstream.

### Relevant log output

```Shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Horiodino:
@adityasaky, I’d like to move forward with this and may need your assistance if I run into any obstacles.

PR #840, opened 2025-02-19
fetch trustedRoots from json

Horiodino:
fixes: https://github.com/gittuf/gittuf/issues/792#event-16312411272


internal/signerverifier/sigstore/sigstore.go
```diff
@@ -175,80 +163,23 @@ func (v *Verifier) ExpectedExtensionKind() string {
...
-			keyID: rekorTransparencyLog,
-		}
-
-		// TODO: CT Log
-		// TODO TSA
-
-		trustedRoot, err := root.NewTrustedRoot(root.TrustedRootMediaType01, fulcioCertAuthorities, nil, nil, rekorTransparencyLogs)
-		return trustedRoot, true, err
+func (v *Verifier) getTUFRoot() (root.TrustedMaterial, error) {
+	trustedRootFilePath := os.Getenv(EnvSigstoreRootFile)
```
adityasaky, line 167: we don't need this anymore either, no? this is still using an env var that we likely don't want to support...

Horiodino: its for the trusted_root.json file path, if you would like we can just set it default for pwd, but it would be better to give an option i think.

adityasaky:
fyi: I'm at a conference, so my review will be slightly delayed, apologies.

internal/signerverifier/sigstore/sigstore.go
adityasaky: yes, my point is, doesn't sigstore-go's tuf client handle this for us now?

Horiodino: sigstore-go says `Load the tuf root.json if provided, if not use public good` , 

```
func DefaultOptions() *Options {
	var opts Options
	var err error

	opts.Root = DefaultRoot()
	home, err := os.UserHomeDir()
	if err != nil {
		// Fall back to using a TUF repository in the temp location
		home = os.TempDir()
	}
	opts.CachePath = filepath.Join(home, ".sigstore", "root")
	opts.RepositoryBaseURL = DefaultMirror

	return &opts
}

// DefaultRoot returns the root.json for the public good instance
func DefaultRoot() []byte {
	// The embed file system always uses forward slashes as path separators,
	// even on Windows
	p := "repository/root.json"

	b, err := embeddedRepo.ReadFile(p)
	if err != nil {
		// This should never happen.
		// ReadFile from an embedded FS will never fail as long as
		// the path is correct. If it fails, it would mean
		// that the binary is not assembled as it should, and there
		// is no way to recover from that.
		panic(err)
	}

	return b
}
``` 
at the same time it have flag for parsing the path so i think its better if we use both method , wdyt ?

adityasaky: To confirm, this works fine and loads the root.json found in the `~/.sigstore` directory if said directory already exists right? In other words, I wonder if what we have to communicate to private sigstore users is that they must use sigstore-specific tooling like `cosign` or `gitsign` to initialize the `~/.sigstore` directory with their root metadata. From then on, gittuf can take over? WDYT?

go.mod
```diff
@@ -15,8 +15,8 @@ require (
 	github.com/sigstore/cosign/v2 v2.4.1
 	github.com/sigstore/gitsign v0.12.0
 	github.com/sigstore/protobuf-specs v0.4.0
-	github.com/sigstore/sigstore v1.9.1
-	github.com/sigstore/sigstore-go v0.6.2
+	github.com/sigstore/sigstore v1.8.15
```
adityasaky, line 18: was this an accidental downgrade? :)

adityasaky:
I'll review this with a private sigstore instance, but eyeballing it, it looks good. :)

go.mod
Horiodino: yeah

internal/signerverifier/sigstore/sigstore.go
adityasaky: mea culpa, you were right @Horiodino :)

I misunderstood the structure of ~/.sigstore, and thought it'd a) include a root.json and b) use it if it found it, so sigstore initialization (out of band) would result in private sigstore's root.json being loaded and used rather than the embedded public good instance's root.json. Given that's not the case, I suggest rolling this change back to load the root.json from a path in the environment variable for now. Apologies!

Horiodino: np :+1:

internal/signerverifier/sigstore/sigstore.go
```diff
@@ -175,80 +163,22 @@ func (v *Verifier) ExpectedExtensionKind() string {
...
-
-		// TODO: CT Log
-		// TODO TSA
-
-		trustedRoot, err := root.NewTrustedRoot(root.TrustedRootMediaType01, fulcioCertAuthorities, nil, nil, rekorTransparencyLogs)
-		return trustedRoot, true, err
+func (v *Verifier) getTUFRoot() (root.TrustedMaterial, error) {
+	rootFilePath := os.Getenv("SIGSTORE_TRUSTED_ROOT")
+	if rootFilePath == "" {
+		return nil, tuf.ErrNoTrustedRootFound
```
adityasaky, line 169: we should default to the public good instance in this case, since the private sigstore root isn't declared.

adityasaky: I would also use https://github.com/sigstore/sigstore-go/blob/main/pkg/tuf/options.go#L85-L88 so we're still using tuf semantics to bootstrap trust for the root file.

Horiodino: I don't think that's necessary—the `DefaultOptions` already takes care of that. [[Reference](https://github.com/sigstore/sigstore-go/blob/main/pkg/tuf/options.go#L125)]

> I would also use https://github.com/sigstore/sigstore-go/blob/main/pkg/tuf/options.go#L85-L88 so we're still using tuf semantics to bootstrap trust for the root file.

adityasaky: Yes, but you return an error before the client is hit, no? In the public good case, we don't need the env var set.

Horiodino: is it correct now ?

internal/signerverifier/sigstore/sigstore.go
```diff
@@ -175,80 +163,29 @@ func (v *Verifier) ExpectedExtensionKind() string {
...
-		trustedRoot, err := root.NewTrustedRoot(root.TrustedRootMediaType01, fulcioCertAuthorities, nil, nil, rekorTransparencyLogs)
-		return trustedRoot, true, err
+		return root.GetTrustedRoot(c)
 	}
 
-	// Use the TUF flow
-	// TODO: support custom sigstore TUF root URL
-
-	tufClient, err := sigstoretuf.New(sigstoretuf.DefaultOptions())
+	trustedRootBytes, err := os.ReadFile(rootFilePath)
```
adityasaky, line 179: This is a bit off. Unfortunately, "root" is a bit overloaded here. We want to load the root.json bytes for the TUF repository, that's what's embedded in the sigstore-go library as well for the public good instance. Then, using that TUF root and the TUF client, we want to separately fetch `trusted_root.json` which is not the same as the TUF root.

TUF root -> https://github.com/sigstore/sigstore-go/blob/main/pkg/tuf/repository/root.json
trusted_root.json -> https://tuf-repo-cdn.sigstore.dev/targets/f44a1b88128e55ebfb62189becbc0fa48d4ec9915c65ac54ba0e46a008b12d5b.trusted_root.json on the public good instance

adityasaky: I suggest you load the TUF root from the env var if it's set. Then override the tuf client's options to set the bytes you loaded instead of the default, embedded tuf root file.

internal/signerverifier/sigstore/sigstore.go
```diff
@@ -175,80 +163,37 @@ func (v *Verifier) ExpectedExtensionKind() string {
...
-		rekorTransparencyLog := &root.TransparencyLog{
-			BaseURL:           v.rekorURL,
-			HashFunc:          crypto.SHA256,
-			ID:                keyHash[:],
-			PublicKey:         rekorKey,
-			SignatureHashFunc: crypto.SHA256,
+		opts := &sigstoretuf.Options{
+			Root:              tufRootBytes,
+			CacheValidity:     sigstoretuf.DefaultOptions().CacheValidity,
+			RepositoryBaseURL: sigstoretuf.DefaultOptions().RepositoryBaseURL,
```
adityasaky, line 187: 😬 we need to establish this too, let me see if sigstore-go / initialization puts it someplace we can use in ~/.sigstore.

Horiodino: RepositoryBaseURL ? Yes cant be used try running without that cause errs

adityasaky: Yes, in a private instance, the repository URL will differ. Let me see if the sigstore-go folks have an idea.

adityasaky:
This **builds** with https://github.com/sigstore/gitsign/pull/733 applied locally. We'll have to see if that PR is right. More work is necessary in gitsign to fully remove online verification here, as commit signatures will need to be looked up against the tlog.

patzielinski (approved):
Signing and verification work for me locally.

adityasaky (approved):

PR #837, opened 2025-02-18
Change how gittuf ref updates are recorded in the RSL

adityasaky:

PR #845, opened 2025-02-20
*: Add gittuf sync

adityasaky:
Broken away from #837

### DIFF
--- a/internal/gitinterface/sync.go
+++ b/internal/gitinterface/sync.go
@@ -64,7 +64,7 @@ func (r *Repository) Fetch(remoteName string, refs []string, fastForwardOnly boo
 }
 
 func (r *Repository) FetchObject(remoteName string, objectID Hash) error {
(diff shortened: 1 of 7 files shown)


internal/gitinterface/sync.go
```diff
@@ -64,7 +64,7 @@ func (r *Repository) Fetch(remoteName string, refs []string, fastForwardOnly boo
 }
 
 func (r *Repository) FetchObject(remoteName string, objectID Hash) error {
-	args := []string{"fetch", "--depth=1", remoteName, objectID.String()}
+	args := []string{"fetch", remoteName, objectID.String()}
```
adityasaky, line 67: `depth=1` results in `objectID` being fetched but as a "graft", meaning we can't reason about whether it's ahead or behind other objects.

adityasaky: this was a real head scratcher when writing tests because merge-base --is-ancestor reported "false" when asked if the parent commit was the ancestor of the test commit.

PR #859, opened 2025-03-06
[GUPPIEs] luasandbox: Add sandbox apis

patzielinski:
This PR introduces a set of APIs to the lua sandbox introduced in #813.

~~Blocked by #813.~~
~~Stacked on #870.~~
~~Stacked on #872.~~


internal/luasandbox/api.go
adityasaky, line 1: Can you add a note that specifies what goes in this file vs in util.go?

internal/luasandbox/main.go
adityasaky, line 1: Can we have a README.md that lists every single API we provide within the sandbox environment? With API signature and description of use, please.

internal/luasandbox/api.go
```diff
@@ -0,0 +1,135 @@
...
+	// Split input into lines for processing
+	parseStart := time.Now()
+	lines := strings.Split(gitDiffOutput, "\n")
+	log.Printf("Parse: %v", time.Since(parseStart))
+
+	// Process each line of the diff
+	regexStart := time.Now()
+	for _, line := range lines {
+		switch {
+		case strings.HasPrefix(line, "+++"):
```
adityasaky, line 82: is this effectively parsing the git diff file / text? I wonder if this is too specific to parsing git diff to be usable for general purpose regex matching?

internal/luasandbox/util.go
```diff
@@ -0,0 +1,144 @@
...
+	output, err := cmd.CombinedOutput()
+	if err != nil || strings.TrimSpace(string(output)) != "true" {
+		// Get the path to the .git directory
+		cmd = exec.Command("git", "rev-parse", "--git-dir")
+		output, err = cmd.CombinedOutput()
+		if err != nil {
+			return ""
+		}
+		gitDir := strings.TrimSpace(string(output))
+		if gitDir == "." || gitDir == ".git" {
```
adityasaky, line 34: does this hold up on windows?

internal/luasandbox/main.go
fr0m-scratch: Will do!

internal/luasandbox/api.go
fr0m-scratch: Yes, this is for matching patterns against git diff contents only, targeting on text additions. I will go ahead and also add some general regex helpers.

internal/luasandbox/util.go
patzielinski: Just tried the command on a Windows box and it looks to hold up, given it does output `.` or `.git`.

internal/luasandbox/api.go
```diff
@@ -0,0 +1,389 @@
...
+	"time"
+
+	lua "github.com/yuin/gopher-lua"
+)
+
+// This file (api.go) contains functions that are exposed to lua programs.
+
+// Lua sandbox helper functions in pure Lua
+var pureLuaHelpers = `
+	function splitString(str, sep)
```
adityasaky, line 30: not in readme 👀

adityasaky: fyi i'm working on this PR

internal/luasandbox/apis.go
```diff
@@ -75,16 +77,16 @@ func (l *LuaEnvironment) apiMatchRegex() API {
 		Name:      "matchRegex",
 		Signature: "matchRegex(pattern, text) -> matched",
 		Help:      "Check if the regular expression pattern matches the provided text.",
-		Implementation: func(l *lua.LState) int {
-			pattern := l.ToString(1)
-			text := l.ToString(2)
+		Implementation: func(s *lua.LState) int {
```
adityasaky, line 80: is this just to avoid shadowing the `l` from the environment / receiver?

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
 		`,
 	}
 }
+
+func (l *LuaEnvironment) apiGitReadBlob() API {
+	return &GoAPI{
+		Name:      "gitReadBlob",
+		Signature: "gitReadBlob(hash) -> blob bytes",
```
adityasaky, line 123:
```suggestion
		Signature: "gitReadBlob(blobID) -> blob bytes",
```

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+		Help:      "Retrieve the Git blob in byte form from the repository with the provided hash.",
```
adityasaky, line 124:
```suggestion
		Help:      "Retrieve the bytes of the Git blob specified using its ID from the repository.",
```

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+		Examples: []string{
+			"gitReadBlob(abcdef) -> 'Hello, world!'",
+		},
+		Implementation: func(s *lua.LState) int {
+			object := s.ToString(1)
```
adityasaky, line 129:
```suggestion
			blobID := s.ToString(1)
```

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+				s.Push(lua.LString(err.Error()))
+				return 1
+			}
+
+			blob, err := l.repository.ReadBlob(hash)
+			if err != nil {
+				s.Push(lua.LString(err.Error()))
+				return 1
+			}
+			s.Push(lua.LString(blob))
```
adityasaky, line 141: what happens when we have binary blobs?

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+			return 0
+		},
+	}
+}
+
+func (l *LuaEnvironment) apiGitGetObjectSize() API {
+	return &GoAPI{
+		Name:      "gitGetObjectSize",
+		Signature: "gitGetObjectSize(hash) -> size",
```
adityasaky, line 150:
```suggestion
		Signature: "gitGetObjectSize(objectID) -> size",
```

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+		Help:      "Retrieve the Git object in byte form from the repository with the provided hash.",
```
adityasaky, line 151: copy error

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+		Examples: []string{
+			"gitReadBlob(abcdef) -> 13",
```
adityasaky, line 153: and here

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+		},
+		Implementation: func(s *lua.LState) int {
+			object := s.ToString(1)
```
adityasaky, line 156:
```suggestion
			objectID := s.ToString(1)
```

adityasaky: can this ever panic?

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+	}
+}
+
+func (l *LuaEnvironment) apiGitGetReference() API {
+	return &GoAPI{
+		Name:      "gitGetReference",
+		Signature: "gitGetReference(ref) -> hash",
+		Help:      "Retrieve the tip of the specified Git reference.",
+		Examples: []string{
+			"gitGetReference(main) -> e7fca95377c9bad2418c5df7ab3bab5d652a5309",
```
adityasaky, line 180:
```suggestion
			"gitGetReference("main") -> e7fca95377c9bad2418c5df7ab3bab5d652a5309",
```

internal/luasandbox/apis.go
adityasaky:
```suggestion
			"gitReadBlob("abcdef") -> 13",
```

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
 	}
 }
+
+func (l *LuaEnvironment) apiGitReadBlob() API {
+	return &GoAPI{
+		Name:      "gitReadBlob",
+		Signature: "gitReadBlob(hash) -> blob bytes",
+		Help:      "Retrieve the Git blob in byte form from the repository with the provided hash.",
+		Examples: []string{
+			"gitReadBlob(abcdef) -> 'Hello, world!'",
```
adityasaky, line 126:
```suggestion
			"gitReadBlob("abcdef") -> 'Hello, world!'",
```

adityasaky: can we use actual IDs here? we don't want to accidentally communicate support for shortened hashes and such...

internal/luasandbox/apis.go
adityasaky: Should help communicate that it handles absolute reference conversion? do we do it or do we let `git` handle it for us? I think it's the latter

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+				return 1
+			}
+
+			blob, err := l.repository.ReadBlob(hash)
+			if err != nil {
+				s.Push(lua.LString(err.Error()))
+				return 1
+			}
+			s.Push(lua.LString(blob))
+			return 0
```
adityasaky, line 142: shouldn't this be return 1? am I misunderstanding this? I assumed it was "pop n from stack"

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+			s.Push(lua.LString(message))
+			return 0
+		},
+	}
+}
+
+func (l *LuaEnvironment) apiGitGetFilePathsChangedByCommit() API {
+	return &GoAPI{
+		Name:      "gitGetFilePathsChangedByCommit",
+		Signature: "gitGetFilePathsChangedByCommit(hash) -> len(paths), paths",
```
adityasaky, line 274:
```suggestion
		Signature: "gitGetFilePathsChangedByCommit(commitID) -> len(paths), paths",
```

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+			s.Push(lua.LString(ref))
+			return 0
+		},
+	}
+}
+
+func (l *LuaEnvironment) apiGitGetCommitMessage() API {
+	return &GoAPI{
+		Name:      "gitGetCommitMessage",
+		Signature: "gitGetCommitMessage(hash) -> message",
```
adityasaky, line 247:
```suggestion
		Signature: "gitGetCommitMessage(commitID) -> message",
```

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,208 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+				s.Push(lua.LString(err.Error()))
+				return 1
+			}
+
+			message, err := l.repository.GetCommitMessage(hash)
+			if err != nil {
+				s.Push(lua.LString(err.Error()))
+				return 1
+			}
+			s.Push(lua.LString(message))
```
adityasaky, line 265: are inline `\n` etc handled correctly? what if the message has a `\0` in there for some bizarre reason?

internal/luasandbox/apis.go
patzielinski: Correct.

internal/luasandbox/apis.go
patzielinski: I don't think the conversion can panic but invalid hashes are handled below with `NewHash`...

internal/luasandbox/apis.go
patzielinski: Good point - I figure we can either detect if the blob is binary and return an error as not supported, or we can return it as a custom data type?

internal/luasandbox/apis.go
patzielinski: `\n` should be passed through but let me poke around at this?

internal/luasandbox/apis.go
patzielinski: I believe it's the latter. Does it make sense to still keep the API function below this one though? (`apiGitGetAbsoluteReference`)

internal/luasandbox/apis.go
patzielinski: Have functionality to this effect ready in https://github.com/gittuf/gittuf/pull/883, will add a check here when that's merged.

adityasaky: I'm not sure that's the right approach...

patzielinski: I'll see then if I can pass a custom or binary type.

adityasaky: is the conversion to string necessary? Can't we just pass the raw bytes back and let it be handled by the caller?

patzielinski: I think gopher-lua mandates the use of the custom types it provides. `LString` is what we've mostly been using for now, but there is a custom type that I can pass through. I don't see a raw byte type unfortunately.

patzielinski: So Lua strings are designed to hold binary data: https://www.lua.org/pil/21.2.2.html. I think at this point it's probably easier/better to just return the data as the string and also add the aforementioned binary file check to the API for the script to know what type of data it'll be working on...

internal/luasandbox/apis.go
patzielinski: Lua strings are safe with these values: https://www.lua.org/manual/5.3/manual.html

internal/luasandbox/apis.go
patzielinski: The test suite I added tests for both text and binary blobs, and both work properly.

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,237 @@ func (l *LuaEnvironment) apiStrSplit() API {
 		`,
 	}
 }
+
+func (l *LuaEnvironment) apiGitReadBlob() API {
+	return &GoAPI{
+		Name:      "gitReadBlob",
+		Signature: "gitReadBlob(blobID) -> blob bytes",
```
adityasaky, line 123:
```suggestion
		Signature: "gitReadBlob(blobID) -> blob",
```

adityasaky: not sure what signature semantics we should use, I wonder if we want to mirror how lua does it?

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,237 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+				s.Push(lua.LString(err.Error()))
+				return 1
+			}
+
+			targetID, err := l.repository.GetTagTarget(hash)
+			if err != nil {
+				s.Push(lua.LString(err.Error()))
+				return 1
+			}
+			s.Push(lua.LString(targetID))
```
adityasaky, line 195: this is probably wrong, we may need to targetID.String() it to get the hex encoding?

adityasaky: https://github.com/gittuf/gittuf/blob/main/internal/gitinterface/hash.go#L35-L37

internal/luasandbox/apis.go
```diff
@@ -114,3 +116,237 @@ func (l *LuaEnvironment) apiStrSplit() API {
...
+			s.Push(&resultTable)
+			return 1
+		},
+	}
+}
+
+func (l *LuaEnvironment) apiGitGetRemoteURL() API {
+	return &GoAPI{
+		Name:      "gitGetRemoteURL",
+		Signature: "gitGetRemoteURL(remote) -> url",
```
adityasaky, line 334:
```suggestion
		Signature: "gitGetRemoteURL(remoteName) -> remoteURL",
```

internal/luasandbox/luasandbox_test.go
```diff
@@ -24,3 +28,641 @@ func TestNewLuaEnvironment(t *testing.T) {
...
+		err = environment.lState.DoString(testScript)
+		assert.Nil(t, err)
+
+		result := environment.lState.Get(-1)
+		environment.lState.Pop(1)
+
+		assert.Equal(t, lua.LString("Error: error parsing regexp: missing argument to repetition operator: `*`"), result)
+	})
+
+	environment.Cleanup()
```
adityasaky, line 101: maybe defer these after creating the environment?

adityasaky (changes requested):
looks a lot better!

internal/luasandbox/apis.go
patzielinski: ah, good catch!

internal/luasandbox/apis.go
patzielinski: I think these are fine for right now. Lua uses something similar: https://www.lua.org/manual/5.2/manual.html#6.1

The only difference I really see is the space between the API name and the parameters, as well as describing output in a text block...

adityasaky (approved):

PR #861, opened 2025-03-07
[GUPPIEs] *: Add hook invocation

patzielinski:
This PR adds functionality to the gittuf API that enables the invocation of gittuf hooks encoded in the metadata. This is used in gittuf-git to invoke hooks at the proper stage.

~~Blocked by #813.~~
~~Blocked by #839.~~

### DIFF
--- a/experimental/gittuf/hook.go
+++ b/experimental/gittuf/hook.go
@@ -4,12 +4,23 @@
 package gittuf
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io/fs"
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strings"
+
+	hookopts "github.com/gittuf/gittuf/experimental/gittuf/options/hooks"
+	"github.com/gittuf/gittuf/internal/gitinterface"
+	"github.com/gittuf/gittuf/internal/luasandbox"
+	luasandboxopts "github.com/gittuf/gittuf/internal/luasandbox/options/luasandbox"
+	"github.com/gittuf/gittuf/internal/policy"
+	sslibdsse "github.com/gittuf/gittuf/internal/third_party/go-securesystemslib/dsse"
+	"github.com/gittuf/gittuf/internal/tuf"
+	lua "github.com/yuin/gopher-lua"
 )
 
 type ErrHookExists struct {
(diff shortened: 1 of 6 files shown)


experimental/gittuf/hook.go
```diff
@@ -22,8 +30,97 @@ func (e *ErrHookExists) Error() string {
...
+
 var HookPrePush = HookType("pre-push")
 
+// InvokeHook runs the hooks defined in the specified stage for the user defined
+// by the supplied signer. A check is performed that the user holds the private
+// key necessary for signing, to support the generation of attestations. Upon
+// successful completion of all hooks for the stage for the user, the map of
+// hook names to exit codes is returned.
+func (r *Repository) InvokeHook(ctx context.Context, stage tuf.HookStage, signer sslibdsse.Signer, targetsRoleName string, attest bool, parameters ...string) (map[string]int, error) {
+	keyID, err := signer.KeyID()
```
adityasaky, line 45: will we always have a signer for this?

adityasaky: what about cases where execution is because of a git action where the signer is derived from git config?

patzielinski: fair point - I assume we can then just consume a key ID as a string and in the case that there is none supplied, attempt to load something from the Git configuration?

experimental/gittuf/hook.go
```diff
@@ -22,8 +28,77 @@ func (e *ErrHookExists) Error() string {
...
+	ErrNoHooksFoundForPrincipal = errors.New("no hooks found for the specified principal")
+)
+
 var HookPrePush = HookType("pre-push")
 
+// InvokeHook runs the hooks defined in the specified stage for the user defined
+// by principalID. Upon successful completion of all hooks for the stage for the
+// user, the map of hook names to exit codes is returned.
+// TODO: Add attestations workflow
+func (r *Repository) InvokeHook(ctx context.Context, stage tuf.HookStage, principalID, targetsRoleName string, parameters ...string) (map[string]int, error) {
```
adityasaky, line 41: nit: the name of this API indicates that a specific hook will be invoked, when this actually identifies what hook(s) to execute.

adityasaky: Have we thought through how the delegations graph is parsed to identify what programs to load and execute?

edit: ah they only live in the root metadata

adityasaky (approved):

experimental/gittuf/hook.go
patzielinski, line 41: pass in signer and use it to identify the principalID

experimental/gittuf/hook.go
```diff
@@ -22,8 +33,148 @@ func (e *ErrHookExists) Error() string {
...
+
+		remoteObjects := make(map[string][]gitinterface.Hash, len(options.RefSpecs))
+
+		for _, refSpec := range options.RefSpecs {
+			splitRefSpec := strings.Split(refSpec, ":")
+			localRef, remoteRef := splitRefSpec[0], splitRefSpec[1]
+
+			var remoteHash gitinterface.Hash
+
+			err = r.r.Fetch(options.RemoteName, []string{remoteRef}, true)
```
adityasaky, line 134: isn't this going to try and fetch into a branch we're pushing? don't we want this fetch to be into remote tracker for the branch?

adityasaky (approved):

PR #886, opened 2025-03-21
policy: Verify controller metadata with upstream

adityasaky:
This adds controller metadata verification to state.Verify.

### DIFF
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"os"
 	"strings"
 
 	"github.com/gittuf/gittuf/internal/common/set"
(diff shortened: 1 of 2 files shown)


internal/policy/policy.go
```diff
@@ -551,6 +553,61 @@ func (s *State) Verify(ctx context.Context) error {
...
+		return nil
+	}
+
+	// Check controller root metadata
+	if len(s.ControllerMetadata) != 0 {
+		controllerRepositories := rootMetadata.GetControllerRepositories()
+		for _, controllerRepositoryDetail := range controllerRepositories {
+			controllerName := controllerRepositoryDetail.GetName()
+
+			tmpDir, err := os.MkdirTemp("", "gittuf-controller-")
```
patzielinski, line 567: Should this string include the controller name?

patzielinski (approved):

PR #910, opened 2025-04-03
*: Remove hook modules field and add timeout

patzielinski:
This PR updates the hooks metadata to remove the `modules` field and add the `timeout` field. After the edits to the PR that added the Lua Sandbox (#859), it doesn't make too much sense to keep the `modules` field and allow arbitrary module loading. We should make the `timeout` configurable for users though, I think.

While this is a breaking change, we haven't released gittuf with these changes so I think we should be OK here.

### DIFF
--- a/internal/cmd/trust/addhook/addhook.go
+++ b/internal/cmd/trust/addhook/addhook.go
@@ -4,7 +4,6 @@
 package addhook
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -14,19 +13,18 @@ import (
 	"github.com/gittuf/gittuf/internal/cmd/common"
 	"github.com/gittuf/gittuf/internal/cmd/trust/persistent"
 	"github.com/gittuf/gittuf/internal/dev"
+	"github.com/gittuf/gittuf/internal/luasandbox"
 	"github.com/gittuf/gittuf/internal/tuf"
 	"github.com/spf13/cobra"
 )
-
-var ErrLuaNoModules = errors.New("must specify modules using --modules flag for Lua sandbox environment")
 
 type options struct {
 	p            *persistent.Options
 	filePath     string
 	hookName     string
 	env          string
-	modules      []string
 	principalIDs []string
+	timeout      int
 
 	isPreCommit bool
 	isPrePush   bool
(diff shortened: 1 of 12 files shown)


internal/cmd/trust/addhook/addhook.go
```diff
@@ -75,20 +73,20 @@ func (o *options) AddFlags(cmd *cobra.Command) {
...
 		&o.principalIDs,
 		"principal-ID",
 		nil,
 		"principal IDs which must run this hook",
 	)
 	cmd.MarkFlagRequired("principal-ID") //nolint:errcheck
+
+	cmd.Flags().IntVar(
+		&o.timeout,
+		"modules",
```
adityasaky, line 86:
```suggestion
		"timeout",
```

internal/cmd/trust/addhook/addhook.go
```diff
@@ -75,20 +73,20 @@ func (o *options) AddFlags(cmd *cobra.Command) {
...
+		luasandbox.LuaTimeOut,
+		"modules which the Lua hook must run",
```
adityasaky, line 88:
```suggestion
		"timeout for hook execution",
```

internal/luasandbox/luasandbox.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	luaTimeOut = 100
+	LuaTimeOut = 100
```
adityasaky, line 21: is this seconds?

internal/luasandbox/luasandbox.go
```diff
@@ -32,9 +32,8 @@ type LuaEnvironment struct {
 	allAPIs       []API
 }
 
-// NewLuaEnvironment creates a new Lua state with the specified modules
-// enabled.
-func NewLuaEnvironment(ctx context.Context, repository *gitinterface.Repository) (*LuaEnvironment, error) {
+// NewLuaEnvironment creates a new Lua state with the specified timeout.
+func NewLuaEnvironment(ctx context.Context, repository *gitinterface.Repository, timeout int) (*LuaEnvironment, error) {
```
adityasaky, line 36: wonder if we should opts... this.

Issue #698, opened 2024-12-17
Add / improve `Long` docstrings for gittuf commands.


patzielinski:
Many commands in gittuf do not have a [`Long`](https://pkg.go.dev/github.com/spf13/cobra#Command) docstring defined, or may not fully explain what a command does. We should add or improve these as needed so that it's clearer for the end-user.

rahul3002:
I would like to work on this issue

patzielinski:
@rahul3002 quick ping, are you still working on this issue?

yongjae354:
Hey @patzielinski I'd like to take a stab at this issue, if that's ok!

Sylani-55:
Hi @patzielinski ,
I'm Syed Mohammed Sylani, and I'm new to Open Source Development. I’ve forked the repository and successfully set up the required environment on my local system.
While reviewing the code, I came across the following snippet:
go
var (
	dir string
	cmd = &cobra.Command{
		Use:   "gendoc",
		Short: "Generate help docs",
		Args:  cobra.NoArgs,
		RunE: func(*cobra.Command, []string) error {
			return doc.GenMarkdownTree(root.New(), dir)
		},
	}
)
I noticed that the `Long` field is not included in the `cobra.Command` definition. Would it be appropriate or recommended to add a `Long: "..."` description to provide more detailed context about what the `gendoc` command does?
Looking forward to your guidance. Thank you!
Best regards,
Syed Mohammed Sylani

Sylani-55:
Hi @patzielinski, @adityasaky 

I'd like to work on this issue. As suggested, I plan to add a Long description to the cobra.Command definition for gendoc to provide more detailed context about its functionality.

I'll open a pull request with the update shortly. Please let me know if there are any specific guidelines or suggestions you'd like me to follow.

Best regards,
Syed Mohammed Sylani.

patzielinski:
@Sylani-55 sounds good! Please note that this issue is scoped to all gittuf commands, and just updating the `Long` description for `gendoc` won't close the issue.

Feel free though to open the PR!

Sylani-55:
Hi @patzielinski,

Thank you for the clarification!

Since this issue applies to all gittuf commands, would you prefer that I update the Long descriptions for all commands as part of a single pull request? Also, could you please assign the issue to me so I can start working on it formally?

Looking forward to your guidance.

Best regards,
Syed Mohammed Sylani.

patzielinski:
[...]

PR #912, opened 2025-04-04
[WIP] Add/Improve Long docstrings for gittuf commands

yongjae354:


patzielinski (commented):
These look good so far at first glance. Quick heads-up that you'll need to run `make generate` before we can merge since the command `Long` string updates also update the command reference docs (in `docs/cli`).

patzielinski (approved):

VijetaPriya47:
Wouldn't it close #698 ?

patzielinski:
This is just a subset of `Long` strings, there are still a number that need to be added.

PR #860, opened 2025-03-06
[GUPPIEs] gittuf-git: Add git command compatibility layer

patzielinski:
This PR adds the gittuf-git command compatibility layer that enables gittuf hooks to be run as part of an otherwise regular Git workflow.

This adds the functionality suggested by https://github.com/gittuf/gittuf/issues/220, but now into the main branch.

~Blocked by #839 and #861.~

PR #915, opened 2025-04-08
*: Add a path parameter to LoadRepository

### LINKED ISSUE #620: Add a path parameter to `LoadRepository`
The `LoadRepository` function, as part of the new gittuf API, loads a repository from the current working directory or the environment var `GIT_DIR`. We should add a parameter specifying the location of the repository to this function for API calls, while preserving or adapting the existing internal usage of `LoadRepository`.

Raised here: https://github.com/gittuf/gittuf/pull/618#discussion_r1813392251

patzielinski:
This PR adds a path parameter to `LoadRepository`, allowing loading of repositories outside of the current working directory. This is something that will make life on #860 easier, as well as something we should really add.

Closes #620.


experimental/gittuf/repository.go
```diff
@@ -24,10 +26,15 @@ func (r *Repository) GetGitRepository() *gitinterface.Repository {
 	return r.r
 }
 
-func LoadRepository() (*Repository, error) {
+func LoadRepository(opts ...repositoryopts.Option) (*Repository, error) {
```
patzielinski, line 29: I'm thinking an option is better here than a string, but can switch to string if we decide that's better for API consumers.

experimental/gittuf/repository_test.go
```diff
@@ -45,6 +48,43 @@ func TestLoadRepository(t *testing.T) {
...
+
+		gitinterface.CreateTestGitRepository(t, tmpDir, false)
+		repository, err := LoadRepository(repositoryopts.WithRepositoryPath(tmpDir))
+		assert.Nil(t, err)
+
+		expectedPath, err := filepath.Abs(filepath.Join(tmpDir, ".git"))
+		require.Nil(t, err)
+		actualPath, err := filepath.Abs(repository.r.GetGitDir())
+		require.Nil(t, err)
+		assert.Equal(t, expectedPath, actualPath)
```
patzielinski, line 86: `filepath.Abs` on macOS isn't resolving the whole path, see https://github.com/gittuf/gittuf/actions/runs/14341969327/job/40203160496. Since this test passes on Ubuntu and Windows, I wonder what the issue is on mac...

adityasaky: fixed

experimental/gittuf/repository.go
adityasaky: made this param compulsory

experimental/gittuf/repository.go
```diff
@@ -24,10 +35,22 @@ func (r *Repository) GetGitRepository() *gitinterface.Repository {
...
+	}
+	slog.SetDefault(slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
+		Level: level,
+	})))
+
+	slog.Debug(fmt.Sprintf("Loading Git repository from '%s'...", repositoryPath))
+
+	if repositoryPath == "" {
+		return nil, gitinterface.ErrRepositoryPathNotSpecified
+	}
```
patzielinski, line 51: might want to swap these around?

adityasaky (approved):

internal/cmd/addhooks/addhooks.go
```diff
@@ -26,7 +26,7 @@ func (o *options) AddFlags(cmd *cobra.Command) {
 }
 
 func (o *options) Run(cmd *cobra.Command, _ []string) error {
-	repo, err := gittuf.LoadRepository()
+	repo, err := gittuf.LoadRepository(".")
```
patzielinski, line 29: With this, I wonder if we'd want to add support for selecting a different directory for the gittuf tool to work on, e.g. --repo-dir as an option added at the same level as --verbose? Can ticketize if we like the idea.

patzielinski (commented):
Two thoughts, but otherwise LGTM!

internal/cmd/addhooks/addhooks.go
adityasaky: let's see if someone wants it...

PR #921, opened 2025-04-18
[WIP] Add gpg signerverifier

yongjae354:
@patzielinski


internal/signerverifier/gpg/gpg.go
```diff
@@ -14,6 +16,50 @@ import (
...
+// NewVerifierFromKey creates a new Verifier from SSlibKey of type GPG.
+func NewVerifierFromKey(key *signerverifier.SSLibKey) (*Verifier, error) {
+	keyring, err := openpgp.ReadArmoredKeyRing(bytes.NewReader([]byte(key.KeyVal.Public)))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse GPG key: %w", err)
+	}
+	// entity := keyring[0]
+	return &Verifier{
+		keyID:   key.KeyID,
+		gpgKey:  entity.PrimaryKey, // TODO: not sure how to get the openpgp.Key instance here, or if it should be a required field for Verifier?
```
patzielinski, line 35: We should require a verifier to be based on one key, IIUC, but maybe this can be an entity and we always use the primary key? See https://pkg.go.dev/github.com/ProtonMail/go-crypto/openpgp#Entity.SigningKey

internal/signerverifier/gpg/gpg.go
```diff
@@ -14,6 +16,50 @@ import (
...
+		keyring: keyring,
+	}, nil
+}
+
+// Verify implements the dsse.Verifier.Verify interface for GPG keys.
+func (v *Verifier) Verify(_ context.Context, data []byte, sig []byte) error {
+	sigReader := bytes.NewReader(sig)
+	// TODO: can we assume sig is in armored format?
```
patzielinski, line 43: I believe the answer is yes. We can also just require this for now and expand support later.

internal/signerverifier/gpg/gpg.go
```diff
@@ -38,3 +188,65 @@ func LoadGPGKeyFromBytes(contents []byte) (*signerverifier.SSLibKey, error) {
 
 	return gpgKey, nil
 }
+
+// FIXME: trying to import gitinterface causes circular dependency
+func getKeyIDFromGitConfig(r *Repository) (string, error) {
```
yongjae354: I'm trying to use code from gitinterface here, but it doesn't allow it because gitinterface already depends on signerverifier.

adityasaky:
Reference for how git handles gpg signing: https://github.com/git/git/blob/68cb7f9e92a5d8e9824f5b52ac3d0a9d8f653dbe/gpg-interface.c#L1010-L1016

experimental/gittuf/root_test.go
```diff
@@ -85,6 +86,50 @@ func TestInitializeRoot(t *testing.T) {
...
+
+		if err := repo.SetGitConfig("gpg.format", "openpgp"); err != nil {
+			t.Fatal(err)
+		}
+		if err := repo.SetGitConfig("user.signingkey", "157507bbe151e378ce8126c1dcfe043cdd2db96e"); err != nil {
+			t.Fatal(err)
+		}
+
+		signer, err := gpg.NewSignerFromKeyID("157507bbe151e378ce8126c1dcfe043cdd2db96e")
+		require.Nil(t, err)
```
yongjae354, line 109:
```suggestion
		// Make a test GPG keyring in tempdir to use for tests
		gpg.SetupTestGPGHomeDir(t, artifacts.GPGKey1Private)
		fingerprintGPG1 := "157507bbe151e378ce8126c1dcfe043cdd2db96e"

		if err := repo.SetGitConfig("gpg.format", "openpgp"); err != nil {
			t.Fatal(err)
		}
		if err := repo.SetGitConfig("user.signingkey", fingerprintGPG1); err != nil {
			t.Fatal(err)
		}

		signer, err := gpg.NewSignerFromKeyID(fingerprintGPG1)
		require.Nil(t, err)
```
I think it might be worth adding a variable here?

internal/signerverifier/git/git.go
yongjae354, line 99: I think we can remove the comment here now.

internal/signerverifier/git/git.go
```diff
@@ -56,7 +55,7 @@ func LoadSignerFromGitConfig(repo *gitinterface.Repository) (sslibdsse.SignerVer
 
 	// Get the path to the signing key, required if using an SSH key
 	signingKey := config["user.signingkey"]
-	if keyType == SigningMethodSSH && signingKey == "" {
+	if signingKey == "" && (keyType == SigningMethodSSH || keyType == SigningMethodGPG) {
```
patzielinski, line 58: I presume we always require that the user define their GPG key? (i.e. no support for the automatic key selection behavior as with Git itself)

adityasaky: Can you explore how git handles it if user.signingkey isn't specified in the config but gpg.format is set? Does git infer which key to use? In the git source code, it passes an explicit key to the gpg program via `-u`, so either it requires explicit configuring or it infers it elsewhere. If it doesn't infer it, we match git behavior here.

yongjae354: > Can you explore how git handles it if user.signingkey isn't specified in the config but gpg.format is set? Does git infer which key to use? In the git source code, it passes an explicit key to the gpg program via `-u`, so either it requires explicit configuring or it infers it elsewhere. If it doesn't infer it, we match git behavior here.


https://github.com/git/git/blob/e0bfec3dfc356f7d808eb5ee546a54116b794397/gpg-interface.c#L949-L960

I found the relevant snippet in the git source code. It calls a `get_default_key` hook if it exists (which seems to be the case when gpg.format is ssh), and otherwise passes in the committer info (name + email) to the gpg program via `-u`. 
Seems like the gpg program resolves the rest, inferring the correct key with what Git provides.

patzielinski: For the time being, I think we can require that the GPG key be specified in the Git config and ticketize this issue.

PR #946, opened 2025-05-05
[GUPPIEs]*: Add update flow for GUPPIEs

fr0m-scratch:
This PR adds an update flow including CLI command for GUPPIES

### DIFF
--- a/experimental/gittuf/root.go
+++ b/experimental/gittuf/root.go
@@ -1042,6 +1042,80 @@ func (r *Repository) RemoveHook(ctx context.Context, signer sslibdsse.SignerVeri
 	return r.updateRootMetadata(ctx, state, signer, rootMetadata, commitMessage, options.CreateRSLEntry, signCommit)
 }
 
+// UpdateHook updates the hook specified by stage and hookName with the new
+// principalIDs, hashes, environment, and timeout.
+func (r *Repository) UpdateHook(ctx context.Context, signer sslibdsse.SignerVerifier, stages []tuf.HookStage, hookName string, hookBytes []byte, environment tuf.HookEnvironment, principalIDs []string, timeout int, signCommit bool, opts ...trustpolicyopts.Option) error {
+	if !dev.InDevMode() {
+		return dev.ErrNotInDevMode
+	}
+
+	if signCommit {
+		slog.Debug("Checking if Git signing is configured...")
+		err := r.r.CanSign()
+		if err != nil {
+			return err
+		}
+	}
+
+	if hookName == "" {
+		return ErrNoHookName
+	}
+
+	if timeout < 1 {
+		return ErrInvalidHookTimeout
+	}
+
+	options := &trustpolicyopts.Options{}
+	for _, fn := range opts {
+		fn(options)
+	}
+
+	rootKeyID, err := signer.KeyID()
+	if err != nil {
+		return err
+	}
+
+	slog.Debug("Loading current policy...")
+	state, err := policy.LoadCurrentState(ctx, r.r, policy.PolicyStagingRef, policyopts.BypassRSL())
+	if err != nil {
+		return err
+	}
+
+	rootMetadata, err := r.loadRootMetadata(state, rootKeyID)
+	if err != nil {
+		return err
+	}
+
+	var hashes = make(map[string]string, 2)
+	blobID, err := r.r.WriteBlob(hookBytes)
+	if err != nil {
+		return err
+	}
+	hashes[gitinterface.GitBlobHashName] = blobID.String()
+
+	sha256Hash := sha256.New()
+	sha256Hash.Write(hookBytes)
+	hashes[gitinterface.SHA256HashName] = hex.EncodeToString(sha256Hash.Sum(nil))
+
+	slog.Debug("Updating hook in rule file...")
+	err = rootMetadata.UpdateHook(stages, hookName, principalIDs, hashes, environment, timeout)
+	if err != nil {
+		return err
+	}
+
+	for _, stage := range stages {
+		updatedHooks, err := rootMetadata.GetHooks(stage)
+		if err != nil {
+			return err
+		}
+
+		state.Hooks[stage] = updatedHooks
+	}
+
+	commitMessage := fmt.Sprintf("Update hook '%s' in root metadata", hookName)
+	return r.updateRootMetadata(ctx, state, signer, rootMetadata, commitMessage, options.CreateRSLEntry, signCommit)
+}
+
 // EnableController makes the current repository a "controller" repository used
 // to specify gittuf policies for other repositories.
 func (r *Repository) EnableController(ctx context.Context, signer sslibdsse.SignerVerifier, signCommit bool, opts ...trustpolicyopts.Option) error {
--- a/experimental/gittuf/root_test.go
+++ b/experimental/gittuf/root_test.go
@@ -1727,6 +1727,195 @@ func TestRemoveHook(t *testing.T) {
 	})
 }
 
+func TestUpdateHook(t *testing.T) {
+	t.Setenv(dev.DevModeKey, "1")
+
+	rootSigner := setupSSHKeysForSigning(t, rootKeyBytes, rootPubKeyBytes)
+	rootPubKey := tufv01.NewKeyFromSSLibKey(rootSigner.MetadataKey())
+
+	t.Run("update pre-commit hook", func(t *testing.T) {
+		r := createTestRepositoryWithRoot(t, "")
+
+		hookStage := tuf.HookStagePreCommit
+		hookName := "test-hook"
+		environment := tuf.HookEnvironmentLua
+		timeout := 100
+		principals := []string{rootPubKey.KeyID}
+
+		state, err := policy.LoadCurrentState(testCtx, r.r, policy.PolicyStagingRef)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		rootMetadata, err := state.GetRootMetadata(false)
+		assert.Nil(t, err)
+		_, err = rootMetadata.GetHooks(tuf.HookStagePreCommit)
+		assert.ErrorIs(t, err, tuf.ErrNoHooksDefined)
+
+		err = r.AddHook(testCtx, rootSigner, []tuf.HookStage{hookStage}, hookName, hookBytes, environment, principals, timeout, true, trustpolicyopts.WithRSLEntry())
+		assert.Nil(t, err)
+
+		err = r.StagePolicy(testCtx, "", true, false)
+		require.Nil(t, err)
+
+		state, err = policy.LoadCurrentState(testCtx, r.r, policy.PolicyStagingRef)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		rootMetadata, err = state.GetRootMetadata(false)
+		assert.Nil(t, err)
+		hooks, err := rootMetadata.GetHooks(tuf.HookStagePreCommit)
+		assert.Nil(t, err)
+
+		hookHash, err := r.r.WriteBlob(hookBytes)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		sha256Hash := sha256.New()
+		sha256Hash.Write(hookBytes)
+		sha256HashSum := hex.EncodeToString(sha256Hash.Sum(nil))
+
+		preCommitHook := tufv01.Hook{
+			Name:         hookName,
+			PrincipalIDs: set.NewSetFromItems(rootPubKey.KeyID),
+			Hashes:       map[string]string{gitinterface.GitBlobHashName: hookHash.String(), gitinterface.SHA256HashName: sha256HashSum},
+			Environment:  environment,
+			Timeout:      timeout,
+		}
+		preCommitHooks := []tuf.Hook{&preCommitHook}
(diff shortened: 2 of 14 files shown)


internal/tuf/v01/root.go
```diff
@@ -961,6 +961,33 @@ func (r *RootMetadata) RemoveHook(stages []tuf.HookStage, hookName string) error
...
+
+	for _, stage := range stages {
+		hookFound := false
+		for i, hook := range r.Hooks[stage] {
+			if hook.Name == hookName {
+				r.Hooks[stage][i].PrincipalIDs = set.NewSetFromItems(principalIDs...)
+				r.Hooks[stage][i].Hashes = hashes
+				r.Hooks[stage][i].Environment = environment
+				r.Hooks[stage][i].Timeout = timeout
+				return r.Hooks[stage][i], nil
```
patzielinski, line 979: Does this return after updating only one stage if the hook is found in that stage?

internal/cmd/trust/updatehook/updatehook.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	return repo.UpdateHook(cmd.Context(), signer, stages, o.hookName, hookBytes, environment, o.principalIDs, o.timeout, true, opts...)
+}
+
+func New(persistent *persistent.Options) *cobra.Command {
+	o := &options{p: persistent}
+	cmd := &cobra.Command{
+		Use:               "update-hook",
+		Short:             fmt.Sprintf("Modify the parameters of an existing gittuf hook, specify the hookname and the parameters to be updated. (developer mode only, set %s=1)", dev.DevModeKey),
```
patzielinski, line 144: We should explicitly define the behavior of this operation with respect to stages.

For example, if I supply all the stages that a hook is defined in, this is fine, but what if I update the hook only in one stage? What if I specify stages that the hook isn't defined in?

patzielinski (commented):
Looks good, a few items that we should think about.

patzielinski:
@fr0m-scratch quick bump on this, did you still want to continue on this?

internal/cmd/trust/updatehook/updatehook.go
fr0m-scratch: I suppose If a hook exists in multiple stages, only the specified stage will be updated. If a hook does not exist in the specified stage, it won't be updated. In which case, user should add a hook to that stage instead. Otherwise, the semantic to duplicate/move hook stages would be too complicated.

internal/tuf/v01/root.go
fr0m-scratch: I made some change that it will update the hook in all specified stages.

experimental/gittuf/root.go
```diff
@@ -991,6 +991,73 @@ func (r *Repository) RemoveHook(ctx context.Context, signer sslibdsse.SignerVeri
 	return r.updateRootMetadata(ctx, state, signer, rootMetadata, commitMessage, options.CreateRSLEntry, signCommit)
 }
 
+func (r *Repository) UpdateHook(ctx context.Context, signer sslibdsse.SignerVerifier, stages []tuf.HookStage, hookName string, hookBytes []byte, environment tuf.HookEnvironment, principalIDs []string, timeout int, signCommit bool, opts ...trustpolicyopts.Option) error {
```
patzielinski: Please add a comment above the function describing its purpose as with the other functions in this file.

experimental/gittuf/root_test.go
```diff
@@ -1633,6 +1633,188 @@ func TestRemoveHook(t *testing.T) {
...
+
+		err = r.UpdateHook(testCtx, rootSigner, []tuf.HookStage{hookStage}, hookName, newHookBytes, environment, principals, timeout,
+			true, trustpolicyopts.WithRSLEntry())
+		assert.Nil(t, err)
+		err = r.StagePolicy(testCtx, "", true, false)
+		require.Nil(t, err)
+		state, err = policy.LoadCurrentState(testCtx, r.r, policy.PolicyStagingRef)
+		if err != nil {
+			t.Fatal(err)
+		}
```
patzielinski:
```suggestion
		}

```

experimental/gittuf/root_test.go
```diff
@@ -1633,6 +1633,188 @@ func TestRemoveHook(t *testing.T) {
...
+		newSha256Hash := sha256.New()
+		newSha256Hash.Write(newHookBytes)
+		newSha256HashSum := hex.EncodeToString(newSha256Hash.Sum(nil))
+
+		err = r.UpdateHook(testCtx, rootSigner, []tuf.HookStage{hookStage}, hookName, newHookBytes, environment, principals, timeout,
+			true, trustpolicyopts.WithRSLEntry())
+		assert.Nil(t, err)
+		err = r.StagePolicy(testCtx, "", true, false)
+		require.Nil(t, err)
+		state, err = policy.LoadCurrentState(testCtx, r.r, policy.PolicyStagingRef)
```
patzielinski:
```suggestion

		state, err = policy.LoadCurrentState(testCtx, r.r, policy.PolicyStagingRef)
```

experimental/gittuf/root_test.go
```diff
@@ -1633,6 +1633,188 @@ func TestRemoveHook(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		rootMetadata, err = state.GetRootMetadata(false)
+		assert.Nil(t, err)
+		hooks, err = rootMetadata.GetHooks(tuf.HookStagePreCommit)
+		assert.Nil(t, err)
+		preCommitHook = tufv01.Hook{
```
patzielinski:
```suggestion

		preCommitHook = tufv01.Hook{
```

experimental/gittuf/root_test.go
```diff
@@ -1633,6 +1633,188 @@ func TestRemoveHook(t *testing.T) {
...
+			Name:         hookName,
+			PrincipalIDs: set.NewSetFromItems(rootPubKey.KeyID),
+			Hashes:       map[string]string{gitinterface.GitBlobHashName: newHookHash.String(), gitinterface.SHA256HashName: newSha256HashSum},
+			Environment:  environment,
+			Timeout:      timeout,
+		}
+		preCommitHooks = []tuf.Hook{&preCommitHook}
+		assert.Equal(t, preCommitHooks, hooks)
+	})
```
patzielinski:
```suggestion
	})

```

experimental/gittuf/root_test.go
```diff
@@ -1633,6 +1633,188 @@ func TestRemoveHook(t *testing.T) {
...
+		newSha256Hash := sha256.New()
+		newSha256Hash.Write(newHookBytes)
+		newSha256HashSum := hex.EncodeToString(newSha256Hash.Sum(nil))
+
+		err = r.UpdateHook(testCtx, rootSigner, []tuf.HookStage{hookStage}, hookName, newHookBytes, environment, principals, timeout,
+			true, trustpolicyopts.WithRSLEntry())
+		assert.Nil(t, err)
+		err = r.StagePolicy(testCtx, "", true, false)
+		require.Nil(t, err)
+		state, err = policy.LoadCurrentState(testCtx, r.r, policy.PolicyStagingRef)
```
patzielinski:
```suggestion

		state, err = policy.LoadCurrentState(testCtx, r.r, policy.PolicyStagingRef)
```

experimental/gittuf/root_test.go
```diff
@@ -1633,6 +1633,188 @@ func TestRemoveHook(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
```
patzielinski:
```suggestion
		}

```

experimental/gittuf/root_test.go
```diff
@@ -1633,6 +1633,188 @@ func TestRemoveHook(t *testing.T) {
...
+		rootMetadata, err = state.GetRootMetadata(false)
+		assert.Nil(t, err)
+		hooks, err = rootMetadata.GetHooks(tuf.HookStagePrePush)
+		assert.Nil(t, err)
+		prePushHook = tufv01.Hook{
```
patzielinski:
```suggestion

		prePushHook = tufv01.Hook{
```

internal/cmd/trust/updatehook/updatehook.go
```diff
@@ -0,0 +1,153 @@
...
+
+	return repo.UpdateHook(cmd.Context(), signer, stages, o.hookName, hookBytes, environment, o.principalIDs, o.timeout, true, opts...)
+}
+
+func New(persistent *persistent.Options) *cobra.Command {
+	o := &options{p: persistent}
+	cmd := &cobra.Command{
+		Use:               "update-hook",
+		Short:             fmt.Sprintf("Modify the parameters of an existing gittuf hook (developer mode only, set %s=1)", dev.DevModeKey),
+		Long:              fmt.Sprintf("Modify the parameters of an existing gittuf hook. Specify the name of the hook to update and provide all parameters with their updated values. You can specify multiple stages where the hook is defined. Note that all parameters required to add the hook must also be provided. If a hook exists in multiple stages, only the specified stage will be updated. If a hook does not exist in the specified stage, it won't be updated; you should add a new hook to that stage instead. Currently, only the 'lua' environment is supported (developer mode only, set %s=1)", dev.DevModeKey),
```
patzielinski:
```suggestion
		Long:              fmt.Sprintf("Modify the parameters of an existing gittuf hook. Specify the name of the hook to update and provide all parameters with their updated values. You can specify multiple stages where the hook is defined. Note that all parameters required to add the hook must also be provided. If a hook exists in multiple stages, only the specified stage(s) will be updated. If a hook does not exist in the specified stage, it won't be updated; you should add a new hook to that stage instead. Currently, only the 'lua' environment is supported (developer mode only, set %s=1)", dev.DevModeKey),
```

We should make this clearer if the update function supports updating a hook in multiple stages.

internal/tuf/v01/root.go
```diff
@@ -962,6 +962,34 @@ func (r *RootMetadata) RemoveHook(stages []tuf.HookStage, hookName string) error
 	return nil
 }
 
+// UpdateHook updates the hook specified by stage and hookName with the new
+// principalIDs, hashes, environment, and timeout.
+func (r *RootMetadata) UpdateHook(stages []tuf.HookStage, hookName string, principalIDs []string, hashes map[string]string, environment tuf.HookEnvironment, timeout int) (tuf.Hook, error) {
```
patzielinski: I think we can just have this return a `nil` if the operation completed successfully without also returning the hook.

internal/tuf/v02/root.go
```diff
@@ -953,6 +953,36 @@ func (r *RootMetadata) RemoveHook(stages []tuf.HookStage, hookName string) error
 	return nil
 }
 
+// UpdateHook updates the hook specified by stage and hookName with the new
+// principalIDs, hashes, environment, and timeout.
+func (r *RootMetadata) UpdateHook(stages []tuf.HookStage, hookName string, principalIDs []string, hashes map[string]string, environment tuf.HookEnvironment, timeout int) (tuf.Hook, error) {
```
patzielinski: Same here as with the function in `v01`.

go.mod
patzielinski, line 1: Looks like this was added by way of the merge commit. Please make sure to remove this change from the PR.

go.sum
patzielinski, line 1: Same as with `go.mod`, please also remove this change from the PR.

internal/tuf/tuf.go
```diff
@@ -217,6 +217,10 @@ type RootMetadata interface {
 	// RemoveHook removes the hook identified by hookName in the specified Git
 	// stage.
 	RemoveHook(stages []HookStage, hookName string) error
+	// UpdateHook updates the hook identified by hookName in the specified Git
+	// stage with the provided parameters. The parameters are the same as those
+	// used in AddHook.
+	UpdateHook(stages []HookStage, hookName string, principalIDs []string, hashes map[string]string, environment HookEnvironment, timeout int) error
```
patzielinski: I suggest moving this above `RemoveHook`.

internal/tuf/v02/root.go
```diff
@@ -981,6 +981,36 @@ func (r *RootMetadata) RemoveHook(stages []tuf.HookStage, hookName string) error
 	return nil
 }
 
+// UpdateHook updates the hook specified by stage and hookName with the new
+// principalIDs, hashes, environment, and timeout.
+func (r *RootMetadata) UpdateHook(stages []tuf.HookStage, hookName string, principalIDs []string, hashes map[string]string, environment tuf.HookEnvironment, timeout int) error {
```
patzielinski: I suggest moving this above `RemoveHook`.

internal/tuf/v01/root.go
```diff
@@ -990,6 +990,34 @@ func (r *RootMetadata) RemoveHook(stages []tuf.HookStage, hookName string) error
 	return nil
 }
 
+// UpdateHook updates the hook specified by stage and hookName with the new
+// principalIDs, hashes, environment, and timeout.
+func (r *RootMetadata) UpdateHook(stages []tuf.HookStage, hookName string, principalIDs []string, hashes map[string]string, environment tuf.HookEnvironment, timeout int) error {
```
patzielinski: Same as with `v02` with moving this above `RemoveHook`.

PR #947, opened 2025-05-05
luasandbox: added two apis for `luasandbox`

fr0m-scratch:
This PR added two additional apis for the lua sandbox:
1. `gitGetBlobID` to retrive blobID given refName and filePath.
2. `gitGetStagedFilePaths` to retrive paths of all staged filesthat yet committed.


internal/gitinterface/blob.go
```diff
@@ -42,3 +43,25 @@ func (r *Repository) WriteBlob(contents []byte) (Hash, error) {
...
+// GetBlobID returns the ID of the blob at the specified path in the given
+// reference. If the reference is ":", it will look for the blob at the path
+// in the current working directory of the repository.
+func (r *Repository) GetBlobID(ref, path string) (Hash, error) {
+	var fullRef string
+	if ref == ":" {
+		fullRef = ":" + path
+	} else {
+		fullRef = ref + ":" + path
+	}
```
patzielinski, line 56: Is this behavior adhering to some kind of standard (i.e. does Git support using the `:` for this?)

internal/gitinterface/status.go
```diff
@@ -98,16 +97,9 @@ func (r *Repository) Status() (map[string]FileStatus, error) {
...
-	if err != nil {
-		return nil, err
-	}
-	if err := os.Chdir(worktree); err != nil {
-		return nil, err
-	}
-	defer os.Chdir(cwd) //nolint:errcheck
 
-	output, err := r.executor("status", "--porcelain=1", "-z", "--untracked-files=all", "--ignored").executeString()
+	output, err := r.executor("-C", worktree, "status", "--porcelain=1", "-z", "--untracked-files=all", "--ignored").executeString()
```
patzielinski, line 101: Do you know why this change is needed?

internal/luasandbox/apis.go
```diff
@@ -350,3 +350,58 @@ func (l *LuaEnvironment) apiGitGetRemoteURL() API {
...
+			statuses, err := l.repository.Status()
+			if err != nil {
+				s.Push(lua.LString(err.Error()))
+				return 1
+			}
+
+			resultTable := s.NewTable()
+			localIndex := 1
+			for filePath, fileStatus := range statuses {
+				if fileStatus.X != gitinterface.StatusCodeUnmodified && fileStatus.X != gitinterface.StatusCodeIgnored {
```
patzielinski, line 372: Do we need to include `StatusCodeUntracked`?

internal/gitinterface/blob.go
fr0m-scratch: I think this usage is specific for `git rev-parse`, referencing here https://git-scm.com/docs/git-rev-parse#_specifying_revisions:~:text=%3A%5B%3Cn%3E%3A%5D%3Cpath%3E,is%20being%20merged.

internal/gitinterface/status.go
fr0m-scratch: The `os.Chdir` call can be pointing to the wrong dir for linked/detached worktrees, so it sometimes misses staged files. Change to `-C worktree` is basically the same thing but works for detached git dir. And it also avoids process-wide `chdir` and racing, so we do not need to do `defer` any more.

patzielinski:
@fr0m-scratch quick bump on this, did you still want to continue on this?

internal/gitinterface/status.go
patzielinski: Let's move this to its own PR, if possible. I'd like to have input from the others on this before merging this bit.

internal/luasandbox/luasandbox_test.go
```diff
@@ -656,3 +659,88 @@ func TestAPIGitGetRemoteURL(t *testing.T) {
...
+func TestAPIGitGetStagedFilePaths(t *testing.T) {
+	tmpDir := t.TempDir()
+	repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
+
+	treeBuilder := gitinterface.NewTreeBuilder(repo)
+	emptyTreeID, _ := treeBuilder.WriteTreeFromEntries(nil)
+	_, _ = repo.Commit(emptyTreeID, "refs/heads/main", "init\n", true)
+
+	_ = os.MkdirAll(filepath.Join(tmpDir, "foo"), 0o755)
+	_ = os.WriteFile(filepath.Join(tmpDir, "foo", "bar.txt"), []byte("hello"), 0o644)
```
patzielinski: Please check that the permissions here are appropriate. If they are OK, then add the following comment to ignore the linter here.
```suggestion
	_ = os.WriteFile(filepath.Join(tmpDir, "foo", "bar.txt"), []byte("hello"), 0o644) //nolint:gosec
```

internal/luasandbox/luasandbox_test.go
```diff
@@ -656,3 +659,88 @@ func TestAPIGitGetRemoteURL(t *testing.T) {
...
+func TestAPIGitGetBlobID(t *testing.T) {
+	tmpDir := t.TempDir()
+	repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
+
+	blobV1, _ := repo.WriteBlob([]byte("v1"))
+	treeBuilder := gitinterface.NewTreeBuilder(repo)
+	treeID, _ := treeBuilder.WriteTreeFromEntries([]gitinterface.TreeEntry{gitinterface.NewEntryBlob("s.txt", blobV1)})
+	_, _ = repo.Commit(treeID, "refs/heads/main", "init\n", true)
+
+	_ = os.WriteFile(filepath.Join(tmpDir, "s.txt"), []byte("v2"), 0o644)
```
patzielinski: Same as above.

internal/gitinterface/blob.go
```diff
@@ -42,3 +43,25 @@ func (r *Repository) WriteBlob(contents []byte) (Hash, error) {
...
+	stdout, err := r.executor("rev-parse", fullRef).executeString()
+	if err != nil {
+		return ZeroHash, fmt.Errorf("unable to resolve blobID for %s in %s: %w", path, ref, err)
+	}
+	blobID, err := NewHash(strings.TrimSpace(stdout))
+	if err != nil {
+		return ZeroHash, fmt.Errorf("invalid blob id: %w", err)
+	}
+	return blobID, nil
+}
```
patzielinski: Let's move this to its own PR so we can iterate on this there. We'll merge the new one before merging this one.

internal/gitinterface/blob.go
```diff
@@ -42,3 +43,25 @@ func (r *Repository) WriteBlob(contents []byte) (Hash, error) {
 
 	return hash, nil
 }
+
+// GetBlobID returns the ID of the blob at the specified path in the given
+// reference. If the reference is ":", it will look for the blob at the path
+// in the current working directory of the repository.
```
patzielinski: I presume you based this off of how Git handles this with respect to the ":"? If so, I'd add a comment inside the function linking to the behavior.

patzielinski (approved):

Issue #953, opened 2025-05-09
Enable inspection of all of gittuf root metadata via the CLI


patzielinski:
There are a few commands that can be used to inspect parts of gittuf root metadata via the CLI, e.g. `list-global-rules` and `list-hooks`. There are also other issues open to inspect other parts of the root metadata, e.g. #879 for propagation directives, but we should also enable inspection of parts such as the root of trust threshold, policy threshold, whether GitHub app approvals are trusted, and so on. (See the available commands under `gittuf trust` to see what is currently available and what is missing).

Below is a (non-exhaustive) list of parts of the root metadata we should enable inspection of:
- [ ] The GitHub app key
- [ ] Whether GitHub app approvals are trusted
- [ ] The policy key
- [ ] Repository location
- [ ] Root threshold
- [ ] Policy threshold

karankoder:
Hello @patzielinski,
Can i work on this issue?

patzielinski:
Sure! Please feel free to reach out on the OpenSSF Slack if you have questions.

MI2007:
@patzielinski Request to work on this issue

R4R3D1FF:
Since the person this issue is assigned to has not made any activity in a long time, I'd like to work on it. Can this be assigned to me?

patzielinski:
This is an issue we have to discuss how we want to proceed, since it has implication on the size of our CLI and compatibility. Let's hold off on this for now.

patzielinski:
Closing this as `inspect-root` is probably sufficient at this time.

PR #957, opened 2025-05-13
feat(trust): add inspect-root command to display root metadata

Abhinav00077:
This PR adds a new command 'gittuf trust inspect-root' that displays the root metadata in a human-readable JSON format. This command is useful for debugging and inspecting the current state of the gittuf root of trust.

Example Output:
{

  "type": "root",

  "schemaVersion": "https://gittuf.dev/policy/root/v0.2",

  "expires": "2026-05-13T23:00:44+05:30",

  "principals": {

    "SHA256:mHq2yTWNE12txaCzmr1cFRsAakDMWIHYruoAszBmxmU": {

      "keyid_hash_algorithms": null,

      "keytype": "ssh",

      "keyval": {

        "public": "AAAAC3NzaC1lZDI1NTE5AAAAIMGNc6wybQYOkaZxSiFuDy651/wLMmG708HYroQst76w"

      },

      "scheme": "ssh-ed25519",

      "keyid": "SHA256:mHq2yTWNE12txaCzmr1cFRsAakDMWIHYruoAszBmxmU"

    }

  },

  "roles": {

    "root": {

      "principalIDs": [

        "SHA256:mHq2yTWNE12txaCzmr1cFRsAakDMWIHYruoAszBmxmU"

      ],

      "threshold": 1

    }

  }

}

Changes:
1.Added a new command 'inspect-root' under the 'trust' subcommand.

2.The command loads the repository, retrieves the current policy state, and pretty-prints the root metadata as JSON.

3.The command is registered in 'internal/cmd/trust/trust.go'.

Testing:
1.Built and tested the command in a gittuf-initialized repository.

2.Verified that the output is correctly formatted and matches the expected root metadata.

Closes: #953

### DIFF
--- a/internal/cmd/trust/inspectroot/inspectroot.go
+++ b/internal/cmd/trust/inspectroot/inspectroot.go
@@ -13,9 +13,18 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type options struct{}
+type options struct {
+	policyRef string
+}
 
-func (o *options) AddFlags(_ *cobra.Command) {}
+func (o *options) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(
+		&o.policyRef,
+		"policy-ref",
+		policy.PolicyStagingRef,
+		"policy reference to inspect",
+	)
+}
 
 func (o *options) Run(cmd *cobra.Command, _ []string) error {
 	repo, err := gittuf.LoadRepository(".")
@@ -23,7 +32,7 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	state, err := policy.LoadCurrentState(cmd.Context(), repo.GetGitRepository(), policy.PolicyStagingRef, policyopts.BypassRSL())
+	state, err := policy.LoadCurrentState(cmd.Context(), repo.GetGitRepository(), o.policyRef, policyopts.BypassRSL())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


Abhinav00077:
Please Review @adityasaky

patzielinski (changes requested):
Hi @Abhinav00077, thanks for the PR! I've made a few suggestions. I don't think this closes the linked issue, as this is just a complete printout of the root metadata; we should add individual commands that are a bit more user-friendly to close that issue, in my opinion.

Please note that all commits submitted for inclusion in gittuf must include a DCO signoff, achievable by using `-s` with `git commit` (see [here](https://wiki.linuxfoundation.org/dco)). You'll need to redo the commit with the signoff and force push.

Abhinav00077:
@patzielinski  i made changes do review

patzielinski:
This still needs a DCO signoff and `make generate` to be run, please do these and I can take another look, thanks!

Abhinav00077:
> This still needs a DCO signoff and `make generate` to be run, please do these and I can take another look, thanks! (Note you'll need to squash commits to ensure that there aren't any commits that don't have a DCO signoff for that check to pass).

@patzielinski i tried again 1.Added DCO signoff to the commit
2.Squashed all commits into a single commit
3.Ran make generate to ensure documentation is up to date

adityasaky:
I wonder if `gittuf trust list` or something like that should pretty-print the different components. Similar to how we have `gittuf policy list-rules` and such?

Abhinav00077:
> I wonder if `gittuf trust list` or something like that should pretty-print the different components. Similar to how we have `gittuf policy list-rules` and such?

On it, it would be more user friendly.

Abhinav00077:
> I agree, that would be a helpful command. I do wonder though if having a command that prints out the JSON metadata for the current state would be helpful to avoid needing to always use manual inspection methods, or if it would be redundant.

agreed i think it would be a good way to access the data programmatically

Abhinav00077:
@patzielinski i was looking into why dco check is failing i realized that The current version of my PR has the correct, signed-off commit (63d405e), but GitHub is still referencing the old commit for the DCO check which is d46462f, i tried closing th PR and reopening it but it still checks for older

patzielinski:
Please take a look at https://stackoverflow.com/questions/5189560/how-do-i-squash-my-last-n-commits-together; after that you should only have one commit on this PR, that has the DCO signoff.

PR #962, opened 2025-05-14
Fix/sigstore offline verification

Abhinav00077:
This PR addresses issue #792 by adding support for offline verification of DSSE signatures using sigstore.

## Changes
1. Added OfflineMode option to sigstore verifier
2. Enabled offline mode by default for policy metadata verification
3. Enabled offline mode for git commit and tag signature verification
4. Removed dependency on online rekor lookups

## Why
The sigstore community is planning to drop support for online rekor lookups for private sigstore instances. This change prepares gittuf for this transition while maintaining security through certificate chain verification.

## Testing
1.Verify that policy metadata can be verified offline
2. Verify that git commit and tag signatures can be verified offline
3. Verify that online verification still works when explicitly enabled

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 dist/*
 *.prof
 vendor*
+gittuf-test/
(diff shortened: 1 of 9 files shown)


Abhinav00077:
@adityasaky please review

adityasaky:
Hi @Abhinav00077, we have a fix for this pending in #840. That's just waiting on me to confirm some details upstream. Thanks for taking a stab at this, though!

PR #968, opened 2025-05-14
feat: add Root and Targets metadata display to policy TUI

### LINKED ISSUE #916: Update the TUI to display more information about Root and Targets metadata (e.g. Keys/Principals)
Currently, the TUI only supports displaying rules. We should update it to support displaying more information about both Root and Targets metadata, such as viewing what the defined keys/principals are, and other information that may be present that is not currently exposed in the TUI, aside from that which other issues are open for.

VijetaPriya47:
This PR adds functionality to view Root and Targets metadata information in the policy TUI. Users can now access detailed information about Root principals, keys, and Targets rules directly from the TUI interface.

## Changes
- Added new screen types `screenViewRootMetadata` and `screenViewTargetsMetadata`
- Added new menu items in the main list for viewing Root and Targets metadata
- Implemented metadata display functionality:
  - Root Metadata view shows:
    - Schema version
    - Root principals with their keys and custom metadata
  - Targets Metadata view shows:
    - Schema version
    - Rules with their patterns, principals, and thrsholds
- Uses `policy.LoadCurrentState()` to load the current policy state

closes #916


internal/cmd/policy/tui/tui.go
```diff
@@ -592,6 +602,85 @@ func (m model) View() string {
...
+			for _, key := range principal.Keys() {
+				sb.WriteString(fmt.Sprintf("        %s (%s)\n", key.KeyID, key.KeyType))
+			}
+			if metadata := principal.CustomMetadata(); len(metadata) > 0 {
+				sb.WriteString("    Custom Metadata:\n")
+				for key, value := range metadata {
+					sb.WriteString(fmt.Sprintf("        %s: %s\n", key, value))
+				}
+			}
+		}
```
patzielinski, line 645: The root metadata currently doesn't support principals, we just have key IDs. This is something we've been trying to address, but I wonder how we should treat it in this PR.

Maybe we can check the type of prinicpal it is (i.e. key or person) and then load the data as appropriate?

patzielinski (changes requested):
Thanks for the PR, @VijetaPriya47! A few suggestions. I think we'll need a bit more functionality in this PR for it to fully close the linked issue.

Please also see if you can resolve the linter issues.

Thanks!

internal/cmd/policy/tui/tui.go
```diff
@@ -592,6 +602,85 @@ func (m model) View() string {
...
+			for _, rule := range rules {
+				sb.WriteString(fmt.Sprintf("\nRule %s:\n", rule.ID()))
+				sb.WriteString(fmt.Sprintf("    Patterns: %s\n", strings.Join(rule.GetProtectedNamespaces(), ", ")))
+				sb.WriteString(fmt.Sprintf("    Principals: %s\n", strings.Join(rule.GetPrincipalIDs().Contents(), ", ")))
+				sb.WriteString(fmt.Sprintf("    Threshold: %d\n", rule.GetThreshold()))
+			}
+		}
+
+		sb.WriteString("\nPress Left Arrow to go back")
+		return lipgloss.NewStyle().Margin(1, 2).Render(sb.String())
```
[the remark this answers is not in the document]
VijetaPriya47: You're right that they seem similar at first glance, but they work for different purposes. “List Rules" gives an overview of the current rules but the "view targets metadata" shows the more detailed view, raw structure including schema versions and internal details, etc.

internal/cmd/policy/tui/tui.go
VijetaPriya47: For this PR, we can update the code to check the type of each “principal” and display the available information (just key ID for now). I’ll also add a code comment referencing the design doc and noting this as a future improvement area. what do you think?

internal/cmd/policy/tui/tui.go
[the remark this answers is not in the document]
patzielinski: There shouldn't be any overlap in this functionality then; just displaying the principals declared in the targets metadata should be sufficient, and we can name that "List Policy Principals" or something similar.

internal/cmd/policy/tui/tui.go
```diff
@@ -592,6 +602,116 @@ func (m model) View() string {
...
+			}
+			// Check if principal has custom metadata (richer object)
+			if metadata := principal.CustomMetadata(); len(metadata) > 0 {
+				sb.WriteString("    Custom Metadata:\n")
+				for key, value := range metadata {
+					sb.WriteString(fmt.Sprintf("        %s: %s\n", key, value))
+				}
+			} else {
+				// Principal is just a key ID
+				sb.WriteString("    Type: Key ID\n")
```
patzielinski, line 650: Isn't this a duplicate of the key display up on line 640?

internal/cmd/policy/tui/tui.go
```diff
@@ -592,6 +602,99 @@ func (m model) View() string {
...
+			return lipgloss.NewStyle().Margin(1, 2).Render(sb.String())
+		}
+
+		// Display schema version
+		sb.WriteString(fmt.Sprintf("Schema Version: %s\n\n", targetsMetadata.SchemaVersion()))
+
+		// Display policy principals
+		principals := targetsMetadata.GetPrincipals()
+		if len(principals) > 0 {
+			sb.WriteString("List Policy Principals:\n")
```
patzielinski, line 681:
```suggestion
			sb.WriteString("Policy Principals:\n")
```

internal/cmd/policy/tui/tui.go
patzielinski: bump on this

VijetaPriya47:
hyy @patzielinski can you review once again?

internal/cmd/policy/tui/tui.go
```diff
@@ -592,6 +602,98 @@ func (m model) View() string {
...
+			sb.WriteString(lipgloss.NewStyle().Foreground(lipgloss.Color(colorFooter)).Render(m.footer))
+			return lipgloss.NewStyle().Margin(1, 2).Render(sb.String())
+		}
+
+		rootMetadata, err := state.GetRootMetadata(false)
+		if err != nil {
+			m.footer = fmt.Sprintf("Error loading root metadata: %v", err)
+			sb.WriteString(lipgloss.NewStyle().Foreground(lipgloss.Color(colorFooter)).Render(m.footer))
+			return lipgloss.NewStyle().Margin(1, 2).Render(sb.String())
+		}
```
patzielinski, line 622: I think we can move this code to a new function similar to `getCurrRules` or the other helpers below.

internal/cmd/policy/tui/tui.go
```diff
@@ -592,6 +602,98 @@ func (m model) View() string {
...
+
+		targetsMetadata, err := state.GetTargetsMetadata(m.policyName, false)
+		if err != nil {
+			m.footer = fmt.Sprintf("Error loading targets metadata: %v", err)
+			sb.WriteString(lipgloss.NewStyle().Foreground(lipgloss.Color(colorFooter)).Render(m.footer))
+			return lipgloss.NewStyle().Margin(1, 2).Render(sb.String())
+		}
```
patzielinski, line 672: Same as above for moving this into a helper.

internal/cmd/policy/tui/tui.go
```diff
@@ -592,6 +602,98 @@ func (m model) View() string {
...
+		sb.WriteString(titleStyle.Render("Root Metadata") + "\n\n")
+
+		state, err := policy.LoadCurrentState(context.Background(), m.repo.GetGitRepository(), m.options.targetRef)
+		if err != nil {
+			m.footer = fmt.Sprintf("Error loading state: %v", err)
+			sb.WriteString(lipgloss.NewStyle().Foreground(lipgloss.Color(colorFooter)).Render(m.footer))
+			return lipgloss.NewStyle().Margin(1, 2).Render(sb.String())
+		}
+
+		rootMetadata, err := state.GetRootMetadata(false)
```
patzielinski, line 617: Would it be better to move this to be a part of the model, like the slices of rules around line 150?

patzielinski (commented):
Hi @VijetaPriya47, did a bit more of an in-depth review, a few suggestions to match the style of the other parts of code in the TUI.

I also tried this on both the gittuf repo, as well as on a fresh repository with gittuf initialized using the gittuf demo instructions, and I couldn't get the root or targets metadata view to load; it always raised the error that it couldn't load the RSL entry. Can you try to reproduce the issue?

Thanks!

internal/policy/policy.go
```diff
@@ -279,6 +279,14 @@ func LoadCurrentState(ctx context.Context, repo *gitinterface.Repository, ref st
...
 	entry, _, err := rsl.GetLatestReferenceUpdaterEntry(repo, rsl.ForReference(ref))
 	if err != nil {
+		if errors.Is(err, rsl.ErrRSLEntryNotFound) {
+			// If no RSL entry exists, fall back to loading directly from the reference
+			commitID, err := repo.GetReference(ref)
+			if err != nil {
+				return nil, err
+			}
+			return loadStateFromCommit(repo, commitID)
+		}
```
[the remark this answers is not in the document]
VijetaPriya47: hi, can you check it again.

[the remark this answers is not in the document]
VijetaPriya47: can you check it now?

internal/cmd/policy/tui/tui.go
```diff
@@ -592,6 +602,84 @@ func (m model) View() string {
...
+		}
+
+		// Display schema version
+		sb.WriteString(fmt.Sprintf("Schema Version: %s\n\n", targetsMetadata.SchemaVersion()))
+
+		// Display policy principals
+		principals := targetsMetadata.GetPrincipals()
+		if len(principals) > 0 {
+			sb.WriteString("Policy Principals:\n")
+			for id, principal := range principals {
```
patzielinski, line 667: Can this also display a principal's keys in addition to their custom metadata?

internal/cmd/policy/tui/tui.go
```diff
@@ -852,3 +940,33 @@ func repoUpdateGlobalRule(o *options, gr globalRule) error {
 		return tuf.ErrUnknownGlobalRuleType
 	}
 }
+
+// getRootMetadata loads and returns the root metadata
+func (m *model) getRootMetadata() (tuf.RootMetadata, error) {
+	state, err := policy.LoadCurrentState(context.Background(), m.repo.GetGitRepository(), policy.PolicyStagingRef)
```
patzielinski, line 946: Instead of hardcoding `policy.PolicyStagingRef`, this should use the ref selected by the user (i.e. `o.targetRef`).

internal/cmd/policy/tui/tui.go
```diff
@@ -852,3 +940,33 @@ func repoUpdateGlobalRule(o *options, gr globalRule) error {
...
+	if err != nil {
+		return nil, fmt.Errorf("error loading root metadata: %w", err)
+	}
+
+	return rootMetadata, nil
+}
+
+// getTargetsMetadata loads and returns the targets metadata
+func (m *model) getTargetsMetadata() (tuf.TargetsMetadata, error) {
+	state, err := policy.LoadCurrentState(context.Background(), m.repo.GetGitRepository(), policy.PolicyStagingRef)
```
patzielinski, line 961: Same as above in `getRootMetadata()`.

patzielinski (commented):
Tested this and it looks to work now, awesome! A few suggestions, and I think we can merge this in soon, thanks.

PR #986, opened 2025-05-18
docs: add Long description to add-key commandDoc add key long desc

Sylani-55:
This PR enhances the 'add-key' command in the gittuf CLI by adding a detailed Long description.

The update clarifies the command’s purpose, supported key formats, required flags, and its integration with the trust policy framework and RSL system.

This update is part of the ongoing effort to fully document all Cobra commands.

Contributor: Syed Mohammed Sylani

### DIFF
--- a/internal/cmd/policy/addkey/addkey.go
+++ b/internal/cmd/policy/addkey/addkey.go
@@ -69,7 +69,7 @@ func New(persistent *persistent.Options) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:               "add-key",
 		Short:             "Add a trusted key to a policy file",
-		Long:              `This command allows users to add trusted keys to the specified policy file. By default, the main policy file is selected. Note that the keys can be specified from disk, from the GPG keyring using the "gpg:<fingerprint>" format, or as a Sigstore identity as "fulcio:<identity>::<issuer>".`,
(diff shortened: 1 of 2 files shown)


internal/cmd/policy/addkey/addkey.go
```diff
@@ -67,9 +67,26 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
...
+
+By default, the main policy file (targets) is used, but you can override this with the --policy-name flag.
+This command also supports generating Record of State Log (RSL) entries when the --rsl flag is enabled.
+
+Requirements:
+- A valid signing key must be provided via --signing-key
+- At least one public key must be supplied using --public-key
+
+Usage:
+  gittuf policy add-key --public-key <path|gpg:fingerprint|fulcio:identity::issuer> [--policy-name <name>] [--signing-key <path>]`,
```
patzielinski, line 89:
```suggestion
		Long: `The 'add-key' command adds one or more trusted public keys to a gittuf policy file. This command is used to define which keys are authorized to sign commits, references, or policy changes
according to the repository's trust model. Various key formats and sources are supported, including local PEM-encoded public key files, GPG keys from a local keyring using the "gpg:<fingerprint>" syntax, and Sigstore identities using the "fulcio:<identity>::<issuer>" syntax. By default, the main policy file (targets) is used, which can be overridden with the --policy-name flag.`,
```

patzielinski (changes requested):

adityasaky (approved):

PR #988, opened 2025-05-18
docs: add Long description to add-rule command Doc add rule long desc

Sylani-55:
This PR adds a comprehensive Long description to the 'add-rule' command in the gittuf CLI.

The updated documentation explains the purpose of the rule, expected input formats for principals and patterns,
how delegation works in policy files, and the significance of signature thresholds and RSL entries.

This is part of an ongoing effort to document all Cobra commands in detail.

Contributor: Syed Mohammed Sylani

### DIFF
--- a/internal/cmd/policy/addrule/addrule.go
+++ b/internal/cmd/policy/addrule/addrule.go
@@ -102,9 +102,33 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 func New(persistent *persistent.Options) *cobra.Command {
 	o := &options{p: persistent}
 	cmd := &cobra.Command{
-		Use:               "add-rule",
-		Short:             "Add a new rule to a policy file",
-		Long:              `This command allows users to add a new rule to the specified policy file. By default, the main policy file is selected. Note that authorized keys can be specified from disk, from the GPG keyring using the "gpg:<fingerprint>" format, or as a Sigstore identity as "fulcio:<identity>::<issuer>".`,
+		Use:   "add-rule",
+		Short: "Add a new rule to a policy file",
+		Long: `The 'add-rule' command adds a new delegation rule to a gittuf policy file, enabling fine-grained
+authorization based on path patterns and principals.
+
+Each rule defines:
+- A name (--rule-name)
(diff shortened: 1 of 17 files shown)


VijetaPriya47:
Hi @Sylani-55 I think you should change all the commands in one PR only instead of opening multiple PRs.

patzielinski:
@Sylani-55, to clarify on what we meant before, I suggest opening one PR per subcommand of commands that need to be updated, e.g. all commands under the `policy` subcommand (commands under `cmd/policy`), etc...

This should be a good balance between too many PRs and too much in one PR and should help us get these merged in quicker. Let me know if you have questions and thanks again!

internal/cmd/policy/addrule/addrule.go
```diff
@@ -102,9 +102,33 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
...
+By default, the rule is added to the main policy file unless --policy-name is specified. If the --rsl flag is passed,
+a Record of State Log (RSL) entry will be added to log this policy change.
+
+Requirements:
+- A valid signing key must be provided using --signing-key
+- At least one of --authorize or --authorize-key must be specified
+- --rule-name and --rule-pattern are required
+
+Usage:
+  gittuf policy add-rule --rule-name <name> --authorize <principalID> --rule-pattern <pattern> [flags]`,
```
patzielinski, line 131:
```suggestion
		Long: `The 'add-rule' command adds a new rule to a gittuf policy file. Each rule contains a name (--rule-name) for the rule, one or more principals (--authorize or --authorize-key) who are allowed to sign within the scope of the rule, a set of rule patterns (--rule-pattern) defining the namespaces or paths the rule governs, and a signature threshold (--threshold), which is the minimum number of valid signatures required to satisfy the rule. Principals can be specified by their principal IDs. By default, the rule is added to the main policy file unless --policy-name is specified.`,
```

patzielinski (changes requested):

adityasaky (approved):

patzielinski (approved):

Issue #466, opened 2024-06-30
Improve testing for GitHub PR attestation functionality


patzielinski:
#439 introduces attestations for GitHub pull request approvals. It's in need of tests, but as some functionality relies on the GitHub API, we'll need to mock the API.

DevAnuragT:
Hi! I'd like to work on this issue.

I'll expand the tests for GitHub pull request approval attestations by:
- Adding tests for edge cases like empty/missing approvers.
- Covering retrieval and indexing logic (e.g., by review ID).
- Validating the structure and contents of generated predicates.
- Mocking any external GitHub API calls if required.

Please let me know if there are any specific cases you'd like me to focus on. Thanks!

patzielinski:
Awesome, thanks @DevAnuragT. I suggest just seeing what you can do overall.

DevAnuragT:
Thanks for the confirmation! I’ll start exploring and expanding the test coverage for the GitHub PR approval attestations as discussed. I’ll aim to cover as many scenarios and edge cases as possible. Will keep you posted on the progress!

yuvraj-kolkar17:
Hi @patzielinski  
Just wanted to kindly follow up on this PR #1067 .
 Please let me know if there’s anything I need to improve or change. I'd love to contribute more to this project

R4R3D1FF:
Hello, can this be assigned to me?

patzielinski:
Sure, please reach out if you have questions.

PR #995, opened 2025-05-21
test: improve test coverage for PR approval attestation (#466)

DevAnuragT:
This PR addresses issue #466 by enhancing the test coverage for GitHub Pull Request approval attestations. The changes include:

- Adding tests for edge cases involving empty or missing approvers and dismissedApprovers fields, including cases where one or both are nil or empty.

- Covering retrieval and indexing logic by asserting that attestations fetched by review ID and related info match the stored data.

- Validating the overall structure of generated predicates by comparing envelopes to ensure correctness.

- No external GitHub API calls are mocked as they are not involved in the tested code paths.


internal/attestations/github_test.go
```diff
@@ -84,10 +90,153 @@ func TestGetGitHubPullRequestApprovalAttestation(t *testing.T) {
...
+			}
+		}
+		if at > 0 {
+			result[i] = email[:at]
+		} else {
+			result[i] = email
+		}
+	}
+	return result
+}
```
patzielinski, line 157: I don't think this is necessary, we should test for the exact data unless it's problematic to do so.

internal/attestations/github_test.go
```diff
@@ -84,10 +90,153 @@ func TestGetGitHubPullRequestApprovalAttestation(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Now unmarshal the Predicate from structpb.Struct into the custom Go struct
+	var parsed v01.PullRequestApprovalAttestation
+	err = common.UnmarshalPBStruct(statement.Predicate, &parsed)
+	require.NoError(t, err)
+
+	// Assertions
+	assert.ElementsMatch(t, []string{"alice", "bob"}, stripEmail(parsed.GetApprovers()))
+	assert.ElementsMatch(t, []string{"charlie"}, stripEmail(parsed.GetDismissedApprovers()))
```
patzielinski, line 136: I think an `assert.Equal` would be preferred here, asserting that each element is equal, as well as an assertion that the length is what is expected.

internal/attestations/github_test.go
```diff
@@ -84,10 +90,153 @@ func TestGetGitHubPullRequestApprovalAttestation(t *testing.T) {
...
+		}
+	}
+	return result
+}
+
+/*
+TestNewPullRequestApprovalAttestation_ErrorOnEmptyLists verifies that
+NewGitHubPullRequestApprovalAttestation returns an error when both
+approvers and dismissedApprovers are empty.
+*/
```
patzielinski, line 163: When these are moved into `t.Run` functions, you can copy the description inside the `t.Run` for this test.

internal/attestations/github_test.go
```diff
@@ -84,10 +90,153 @@ func TestGetGitHubPullRequestApprovalAttestation(t *testing.T) {
...
+	ref := "refs/heads/main"
+	fromID := gitinterface.ZeroHash.String()
+	treeID := gitinterface.ZeroHash.String()
+
+	// Create an attestation with no approvers or dismissed approvers
+	attestation, err := NewGitHubPullRequestApprovalAttestation(ref, fromID, treeID, nil, nil)
+
+	// The attestation should be nil and an error should be returned
+	assert.Nil(t, attestation)
+	assert.Error(t, err)
```
patzielinski, line 174:
```suggestion
	assert.ErrorIs(t, err)
```

internal/attestations/github_test.go
```diff
@@ -84,10 +90,153 @@ func TestGetGitHubPullRequestApprovalAttestation(t *testing.T) {
...
+	toID := gitinterface.ZeroHash.String()
+	approvers := []string{"alice@example.com"}
+	dismissedApprovers := []string{"bob@example.com"}
+
+	// Create an attestation envelope with both approvers and dismissedApprovers
+	env := createGitHubPullRequestApprovalAttestationEnvelope(t, ref, fromID, toID, approvers, dismissedApprovers)
+
+	// Verify that the envelope and payload are valid
+	assert.NotNil(t, env)
+	assert.NotEmpty(t, env.Payload)
```
patzielinski, line 232: These should be compared against an expected envelope.

internal/attestations/github_test.go
```diff
@@ -84,10 +90,153 @@ func TestGetGitHubPullRequestApprovalAttestation(t *testing.T) {
...
+}
+
+// createGitHubPullRequestApprovalAttestationEnvelope creates a GitHub Pull Request Approval attestation envelope
+func createGitHubPullRequestApprovalAttestationEnvelope(t *testing.T, refName, fromID, toID string, approvers, dismissedApprovers []string) *sslibdsse.Envelope {
```
patzielinski, line 236: Are there any other functions which need to change based on the addition of `dismissedApprovers` here?

internal/attestations/github_test.go
```diff
@@ -84,10 +90,153 @@ func TestGetGitHubPullRequestApprovalAttestation(t *testing.T) {
...
+	t.Helper()
```
patzielinski, line 237:
```suggestion
	t.Helper()
	
```

internal/attestations/common/common.go
```diff
@@ -22,3 +22,15 @@ func PredicateToPBStruct(predicate any) (*structpb.Struct, error) {
...
+func UnmarshalPBStruct(pbStruct *structpb.Struct, out any) error {
+	// Convert structpb.Struct to JSON bytes
+	jsonBytes, err := pbStruct.MarshalJSON()
+	if err != nil {
+		return err
+	}
+
+	// Unmarshal into the provided Go struct
+	return json.Unmarshal(jsonBytes, out)
+}
```
patzielinski, line 36: Would it be possible to use `AsMap` instead of having this unmarshaller?

e.g. https://github.com/gittuf/gittuf/blob/main/internal/attestations/github/v01/approval_test.go#L48

patzielinski (changes requested):
Hi @DevAnuragT, thanks for the PR. A few suggestions. I also see that the coverage went down after this PR, could you please look into that?

Thanks!

internal/attestations/github_test.go
```diff
@@ -84,10 +90,153 @@ func TestGetGitHubPullRequestApprovalAttestation(t *testing.T) {
...
 
-	authorization, err := NewGitHubPullRequestApprovalAttestation(refName, fromID, toID, approvers, nil)
+	// Retrieve the stored attestation
+	retrieved, err := attestations.GetGitHubPullRequestApprovalAttestationFor(repo, appName, testRef, testID, testID)
+	assert.Nil(t, err)
+
+	// Extract the statement from the envelope
+	var statement ita.Statement
+	payloadBytes, err := base64.StdEncoding.DecodeString(retrieved.Payload)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
DevAnuragT: Thanks for the suggestion! I wanted to mention that in Go testing, the idiomatic way to assert an error is nil is using require.NoError(t, err). It’s more descriptive and provides clearer failure messages compared to require.Nil(t, err), which is more generic.

For consistency and clarity, I’ve used require.NoError throughout the tests.

Let me know if you still want to change it.

internal/attestations/github_test.go
DevAnuragT: Yes I have already taken care of that.

internal/attestations/github_test.go
```diff
@@ -84,10 +90,153 @@ func TestGetGitHubPullRequestApprovalAttestation(t *testing.T) {
...
+	}
+	return result
+}
+
+/*
+TestNewPullRequestApprovalAttestation_ErrorOnEmptyLists verifies that
+NewGitHubPullRequestApprovalAttestation returns an error when both
+approvers and dismissedApprovers are empty.
+*/
+func TestNewPullRequestApprovalAttestation_ErrorOnEmptyLists(t *testing.T) {
```
adityasaky, line 164: I'd skip the `_...` bit and just make that a `t.Run` description.

DevAnuragT: Yes, I have changed it in my latest commit

internal/attestations/github_test.go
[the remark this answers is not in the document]
patzielinski: Thanks for the suggestion; I figure though given our project's standard of using `require.Nil` means that it's probably better for now to change this to prevent fragmentation in style.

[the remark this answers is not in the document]
DevAnuragT: Sure, adding a new commit.

DevAnuragT:
I have made the necessary changes, can this be merged now?

PR #1018, opened 2025-06-03
fix: set pager environment to support color output (fixes #911)

### LINKED ISSUE #911: RSL output coloring does not work (Fedora 41)
### What happened?

When running commands such as `gittuf rsl log`, it looks like the text coloring feature doesn't quite work properly (see log output below). Turning the feature off with `--no-color` makes the output legible (sans text coloring).

This could very well be an issue with my configuration but opening this here to track this and help others in case the solution is to update a configuration item here.

### Relevant log output

```Shell
ESC[33mentry 61af92b4aef5941156d7e2c3477712fbc0029b16ESC[0m

  Ref:    refs/heads/reconcile-policy-staging-propagation
  Target: 79bf8a22ad41c684147a16332781b699171ba807
  Number: 1361

ESC[33mentry 723975a4e76dcc81f7594514ceb51c7522c51ec4ESC[0m

  Ref:    refs/heads/reconcile-policy-staging-propagation
  Target: 2564c32c7bda9845903e1257b59d6282e8f7d5fa
  Number: 1360

ESC[33mentry fb004f0b4e782e1dcd4eb687083082ae0f4923d2ESC[0m

  Ref:    refs/heads/add-git-config-signer
  Target: cebcca397746c85170d690e94e6c31f3ee346c8a
  Number: 1359

ESC[33mentry 9e5f526d48f8af1db3223a5959fb01e059abfa7bESC[0m

  Ref:    refs/heads/add-git-config-signer
  Target: cb1c5e771c839bb8ca4674cdb675a48fd6316af4
  Number: 1358

ESC[33mentry 3bd46e105a10662eebd0ab1ebbb1ac62ea006febESC[0m

  Ref:    refs/heads/reconcile-policy-staging-propagation
  Target: edba7aad615fd7b23a4f2218e5db7709b6fc42ab
  Number: 1357
```

### Code of Conduct

shivpratikhande:
### Summary

This PR sets the `PAGER` and `LESS` environment variables in `main.go` to ensure ANSI color codes (used for colored output) are rendered properly through pagers like `less`.

Previously, on systems like Fedora and macOS, commands like `gittuf rsl log` displayed raw escape sequences (e.g., `ESC[33m`) instead of colored text. This was traced to the pager not interpreting color correctly.

By setting:

```go
os.Setenv("PAGER", "less -R")
os.Setenv("LESS", "-R")
```
By setting these environment variables in the main.go entry point, we ensure that `less` handles color output correctly, resolving this issue.

### Before
image

### After
image

### DIFF
--- a/.github/workflows/ci-locale.yml
+++ b/.github/workflows/ci-locale.yml
@@ -14,7 +14,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: ['1.23', '1.24']
+        go-version: ['1.24']
     runs-on: ubuntu-latest
     steps:
     - name: Set locale
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        go-version: ['1.23', '1.24']
+        go-version: ['1.24']
     runs-on: ${{ matrix.os }}
     steps:
     - name: Checkout code
--- a/.github/workflows/get-started-tests.yml
+++ b/.github/workflows/get-started-tests.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        go-version: ['1.23', '1.24']
+        go-version: ['1.24']
     runs-on: ${{ matrix.os }}
     steps:
     - name: Checkout code
--- a/.github/workflows/ubuntu-2204.yml
+++ b/.github/workflows/ubuntu-2204.yml
@@ -14,7 +14,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: ['1.23', '1.24']
+        go-version: ['1.24']
     runs-on: ubuntu-22.04
     steps:
       - name: Downgrade Git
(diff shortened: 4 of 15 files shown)


main.go
```diff
@@ -13,6 +13,10 @@ import (
 )
 
 func main() {
+
+	os.Setenv("PAGER", "less -R")
+	os.Setenv("LESS", "-R")
```
patzielinski: I'm not sure if just setting these here is the best option. Maybe we could have these set only if coloring is enabled?

I also wonder whether this works on Windows, and how we should handle that case.

patzielinski (changes requested):
Hi @shivpratikhande, thanks for the PR! I've left a comment on the change.

Please also note that all commits submitted for inclusion in gittuf must include a DCO signoff, achievable by using `-s` with `git commit` (see [here](https://wiki.linuxfoundation.org/dco)). You'll need to redo the commit with the signoff and force push.

main.go
shivpratikhande: Hi @patzielinski,  
I've updated the PR as per your suggestions:  
- Only set `PAGER`/`LESS` if color is enabled  
- Skip setting them on Windows  
These changes were made in the `root.go` file.  

Please let me know if anything else is needed. Thanks!

patzielinski (commented):
It looks like the linter has flagged some items, please make sure to make the suggested changes, thanks!

@adityasaky WDYT of this solution?

Edit: Please also make sure to perform the DCO signoff, you will need to squash commits as every commit in the PR must have the DCO signoff.

PR #1019, opened 2025-06-04
*: Move local cache out of devmode

patzielinski:
This PR moves gittuf's local persistent caching functionality out of devmode.


internal/cmd/cache/populatecache/populatecache.go
```diff
@@ -28,7 +25,7 @@ func New() *cobra.Command {
 	o := &options{}
 	cmd := &cobra.Command{
 		Use:   "populate-cache",
```
adityasaky: so does this become `gittuf cache populate-cache`? I wonder if we want it to be `gittuf cache init`

adityasaky:
also, please rebase, thanks!

adityasaky (approved):

PR #1020, opened 2025-06-05
fix: set pager environment to support color output (fixes #911)

### LINKED ISSUE #911: RSL output coloring does not work (Fedora 41)

shivpratikhande:
This PR replaces #1018 with a DCO-signed commit and required linting fixes.


patzielinski (commented):
This looks to fix the issue on my installation. Have you tried this though on a variety of platforms like macOS and some Linux distros?

You'll also need to squash commits to just one commit, as there are multiple commits for this change.

shivpratikhande:
I have tested this fix on Ubuntu and it works well. 

Regarding the commits, I checked my branch and the PR currently contains only one commit related to this change. If you see multiple commits or want me to squash or amend anything, please let me know!

internal/cmd/root/root.go
```diff
@@ -77,6 +78,9 @@ func (o *options) PreRunE(_ *cobra.Command, _ []string) error {
 	isTerminal := isatty.IsTerminal(output.Fd()) || isatty.IsCygwinTerminal(output.Fd())
 	if o.noColor || !isTerminal {
 		display.DisableColor()
+	} else if runtime.GOOS != "windows" {
+		os.Setenv("PAGER", "less -R")
```
adityasaky, line 83: Won't this override the user's chosen pager? Suppose I run `PAGER=cat gittuf rsl log`, isn't this going to go back to less?

adityasaky: How does Git ensure the right flags are set for colouring output? :thinking:

shivpratikhande: You're right unconditionally setting `PAGER=less -R` would override the user's intent, even if they explicitly set a custom pager (like `PAGER=cat`)

shivpratikhande: While going through the code, I ran the command `echo $PAGER` in the terminal and realized that no default pager is set in my environment. 

Diving into the codebase, I found that the problem stems from the **lack of proper fallback handling** when no pager is defined at all. In such cases, the output displays **raw ANSI escape codes** instead of the expected **colorized output**.

shivpratikhande: To address the issue, I have refined the solution in the following way:

1. **Custom user-defined PAGER**:  
   If the user has explicitly set a `PAGER`, we should avoid overriding it.

2. **Handle fallback when no pager is set**:  
   If `PAGER` is not set, we should assign a **default pager** (like `less -R`) that correctly handles **ANSI escape sequences** and preserves colored output.

3. **Fallback logic condition**:  
   This fallback should only trigger when `PAGER` is unset, ensuring users maintain full control over their pager preferences.

4. **Additional environment variable `LESS`**:  
   Simply setting `PAGER=less` is not sufficient, as `less` requires proper flags (like `-FRX`) to handle color and output correctly.  
   Therefore, we should also set the `LESS` environment variable to flags such as `FRX` (following Git’s fallback pattern).  
   This ensures colored output works as expected.

```go
if runtime.GOOS != "windows" {
    pager := os.Getenv("PAGER")

    if pager == "" {
        os.Setenv("PAGER", "less")
        os.Setenv("LESS", "FRX")
    } else if pager == "less" {
        if os.Getenv("LESS") == "" {
            os.Setenv("LESS", "FRX")
        }
    }
}

shivpratikhande: What do you think about this? @patzielinski @adityasaky

patzielinski: Is this based on how Git does it? Could you provide a link to the source code that does this if so?

shivpratikhande: yes
https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration

shivpratikhande: image

patzielinski: Ah, I meant the actual source code in Git that sets arguments as needed for the pager. i.e. does Git also use the `FRX` parameters you added in the code block?

shivpratikhande: https://git-scm.com/docs/git-config#:~:text=When%20the%20LESS%20environment%20variable%20is%20unset%2C%20Git%20sets%20it%20to%20FRX%20(if%20LESS%20environment

this is the official part from the git doc

image

patzielinski: I took a look through the docs and Git has an order of locations it searches through for the pager to use:

> `core.pager` 
Text viewer for use by Git commands (e.g., **less**). The value is meant to be interpreted by the shell. The order of preference is the `$GIT_PAGER` environment variable, then `core.pager` configuration, then `$PAGER`, and then the default chosen at compile time (usually **less**).

I figure we should mimic this behavior when determining what pager to use?

adityasaky: I opened #1048 to rework how we handle pager management @shivpratikhande. I think this allows us to avoid messy if-else ladders with the flags.

If that's merged, I think we can close this. I've used the configs proposed in this thread in that PR (and acked with Co-authored-by :)). @shivpratikhande, would you be interested in adding a pager handler that looks at git's core.pager config?

shivpratikhande: I’d be happy to add the pager handler similar to gits pager config

shivpratikhande: @adityasaky 
I’d be happy to add the pager handler similar to gits pager config

adityasaky:
Closing in favor of #1048.

PR #1033, opened 2025-06-12
docs: simplify Long descriptions for reconcile and add-github-app commands

Sylani-55:
Updated Long descriptions for:
- `reconcile`: Clarified its purpose as synchronizing the local RSL.
- `add-github-app`: Explained its role in adding a GitHub App key.

CLI documentation has been regenerated to reflect these updates. 
Signed-off-by: Mohammed Sylani sylanmohammed@gmail.com


patzielinski (changes requested):
@Sylani-55 can you take a look at some commands that don't have any `Long` docstrings? I think that the existing ones for the commands here are sufficient.

Thanks!

PR #1042, opened 2025-06-18
Add skeleton for `approve-temp` CLI command

BrawlerXull:
Adds a new `gittuf approve-temp` subcommand for generating temporary approval attestations. 

This pr implements only the CLI scaffolding with flags and stub output, without any signing or policy logic.

### DIFF
--- a/internal/cmd/approve_temp/approve_temp.go
+++ b/internal/cmd/approve_temp/approve_temp.go
@@ -1,0 +1,32 @@
+// Copyright The gittuf Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package approve_temp
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+// New returns the cobra.Command for the `approve-temp` subcommand.
+func New() *cobra.Command {
+	var action string
+	var duration string
+
+	cmd := &cobra.Command{
+		Use:   "approve-temp",
+		Short: "Generate a temporary approval attestation",
+		Long: `Creates a signed, time-limited approval attestation for a PR or commit.
+Currently this is just a stub; full implementation will be added in later phases.`,
+		Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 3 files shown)


BrawlerXull:
# Running `gittuf approve-temp --help`
Screenshot 2025-06-18 at 1 54 37 PM



# Running `gittuf --help`
Screenshot 2025-06-18 at 1 55 15 PM

internal/cmd/approve_temp/approve_temp.go
```diff
@@ -0,0 +1,32 @@
...
+// New returns the cobra.Command for the `approve-temp` subcommand.
+func New() *cobra.Command {
+	var action string
+	var duration string
+
+	cmd := &cobra.Command{
+		Use:   "approve-temp",
+		Short: "Generate a temporary approval attestation",
+		Long: `Creates a signed, time-limited approval attestation for a PR or commit.
+Currently this is just a stub; full implementation will be added in later phases.`,
```
patzielinski, line 21: We can omit the stub part, as we can simply place this command behind developer mode if we need to.
```suggestion
		Long: `Creates a signed, time-limited approval attestation for a PR or commit.`,
```

patzielinski (changes requested):
Hi @BrawlerXull, thanks for the PR! A few items:

- All commits submitted for inclusion in gittuf must have a [DCO signoff](https://wiki.linuxfoundation.org/dco). Please make sure to redo your commits with a DCO signoff.
- Can the format of the command definition be changed to something more along the lines of what's in https://github.com/gittuf/gittuf/blob/main/internal/cmd/attest/authorize/authorize.go?
- I think this command shouldn't be attached to the root, but rather the `attest` subcommand. On naming, maybe we can make this `authorize-temp` to more match the existing `authorize` naming? Unsure about that last point though.
- There's also a few items that the linter has flagged, please make sure to address them.

I think we'll likely have to hold off on merging this until we have something that it can be wired up to; I'm personally not a fan of having a command in the CLI that isn't wired up to anything in the API.

PR #1062, opened 2025-07-18
ci: fix Ubuntu 22.04 CI failure by installing Git from ppa:git-core/ppa

Sylani-55:
### What
This PR fixes CI test failures on Ubuntu 22.04 by installing a newer Git version (from `ppa:git-core/ppa`) in the CI workflow.

### Why
CI was failing with:

This happens because the default Git version (2.34.x) in Ubuntu 22.04 is old, and sometimes the package gets moved to security mirrors.

### How
- Added a CI step (guarded with `if: runner.os == 'Linux'`) to install Git from `ppa:git-core/ppa`.

### Additional notes
✅ Includes DCO sign-off  
📌 Keeps macOS and Windows runners unchanged

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,5 +25,11 @@ jobs:
       with:
         go-version: ${{ matrix.go-version }}
         cache: true
+    - name: Install updated Git from PPA
+      if: runner.os == 'Linux'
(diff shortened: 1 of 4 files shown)


adityasaky:
Fixed in #1059.

Issue #463, opened 2024-06-27
Testing Coverage


patzielinski:
### Add a description

Our testing coverage could use a boost, as indicated by our results when running `go test` in CI. See [this actions run](https://github.com/gittuf/gittuf/actions/runs/9702341498/job/26777969373) for an indication of where we're at.

Go supports generating an HTML report of what code needs tests with `go tool cover`, so that should be helpful in identifying where we need more tests.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Kevin-Aguirre:
Hi, I'm interested in working on this.

I'm unable to see the results of the action run but when I run `go test -timeout 20m -v ./...` I see missing test files for the `github.com/gittuf/gittuf/internal/cmd/`, I'm guessing that's what this is in reference to?

adityasaky:
> I'll be chipping at this one PR at a time (https://github.com/gittuf/gittuf/pull/934).

🙌

Aditya-132:
i want to work on this isuue

patzielinski:
@Aditya-132 sure! Please feel free to reach out on the OpenSSF Slack if you have questions.

eyeaadil:
@patzielinski , I am not able to Join Slack , can you please provide me the invitation link.

Aditya-132:
> [@Aditya-132](https://github.com/Aditya-132) sure! Please feel free to reach out on the OpenSSF Slack if you have questions.

sure

vivekbisen04:
Hi @patzielinski ,
I'm interested in working on this issue and have started writing some test cases. Could you please assign it to me so I can continue

vivekbisen04:
Hi @patzielinski , has this project been removed from the mentorship program?

ayuxsh009:
Hi @patzielinski if this issue is available can i work on this ?

patzielinski:
Sure, this is a general issue referring to our overall testing coverage amount; can you let us know what files or packages you'd like to work on first?

ayuxsh009:
Hi @patzielinski,

I’d like to start with the internal/cache package first:
**cache.go, policy.go, attestations.go, verification.go**

I will add unit tests for entry insertion/lookup edge cases, range selection behavior, last-verified tracking, and persistent cache load/delete behavior.

If that looks good, I can follow up with small utility packages like `version.go` and `common.go`.

patzielinski:
[...]

PR #1067, opened 2025-07-24
test: increase github.go test coverage from 49.2% to 83.1%

yuvraj-kolkar17:
What type of PR is this?

/kind test  
/area attestations

What this PR does / why we need it:  
This PR adds a complete unit test suite for the logic defined in `internal/attestations/github.go`.

### Coverage Improvement

| File      | Previous Coverage | New Coverage |
|-----------|-------------------|--------------|
| github.go | 49.2%             | 83.1%        |

#### Before vs After Coverage:

<p align="center">
  Before Coverage
  After Coverage
</p>

Which issue(s) this PR fixes:  
#463 

Does this PR introduce a user-facing change?  
None

### DIFF
--- a/internal/attestations/github_test.go
+++ b/internal/attestations/github_test.go
@@ -9,10 +9,12 @@ import (
 	"path"
 	"testing"
 
+	"github.com/gittuf/gittuf/internal/attestations/github"
 	"github.com/gittuf/gittuf/internal/gitinterface"
 	"github.com/gittuf/gittuf/internal/signerverifier/dsse"
 	sslibdsse "github.com/gittuf/gittuf/internal/third_party/go-securesystemslib/dsse"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestSetGitHubPullRequestApprovalAttestation(t *testing.T) {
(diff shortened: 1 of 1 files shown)


patzielinski:
This looks to affect the same code tests as #995. I'll take a look if there's anything that is still untested after that one is merged.

internal/attestations/github_test.go
```diff
@@ -84,6 +86,225 @@ func TestGetGitHubPullRequestApprovalAttestation(t *testing.T) {
...
+	attestations := &Attestations{}
+
+	// Add first attestation
+	err := attestations.SetGitHubPullRequestApprovalAttestation(repo, mainZeroZero, baseURL, 1, appName, testRef, testID, testID)
+	require.Nil(t, err)
+
+	// Add another attestation with same review ID and same index path but different app
+	// This should succeed because the same review ID can be observed by more than one app
+	err = attestations.SetGitHubPullRequestApprovalAttestation(repo, mainZeroZero, baseURL, 1, anotherAppName, testRef, testID, testID)
+	assert.Nil(t, err)
```
adityasaky: can we check the state of attestations after this?

patzielinski: Pushed a change to address this.

internal/attestations/github_test.go
```diff
@@ -56,32 +58,258 @@ func TestGetGitHubPullRequestApprovalAttestation(t *testing.T) {
...
+		// Create an invalid envelope (empty envelope)
+		invalidEnv := &sslibdsse.Envelope{}
+
+		tmpDir := t.TempDir()
+		repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
+
+		attestations := &Attestations{}
+
+		// This should fail validation
+		err := attestations.SetGitHubPullRequestApprovalAttestation(repo, invalidEnv, baseURL, 1, appName, testRef, testID, testID)
```
adityasaky: This case shouldn't be under `TestGetGitHubPullRequestApprovalAttestation` I think?

internal/attestations/github_test.go
```diff
@@ -56,32 +58,258 @@ func TestGetGitHubPullRequestApprovalAttestation(t *testing.T) {
...
+
+	t.Run("invalid URL", func(t *testing.T) {
+		tmpDir := t.TempDir()
+		repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
+
+		attestations := &Attestations{}
+
+		// Try to get attestation with invalid URL
+		_, err := attestations.GetGitHubPullRequestApprovalAttestationForReviewID(repo, "invalid-url", 123, "github")
+		assert.NotNil(t, err)
```
adityasaky: Do we have a specific error we can test for here?

internal/attestations/github_test.go
yongjae354: @adityasaky Seems like we already test for validation in a different test file:
https://github.com/gittuf/gittuf/blob/8f5837fba09da6024170a41fad1022f13a035506/internal/attestations/github/v01/approval_test.go#L56

Should we get rid of this particular test case here then? Should we factor it out to a different test function?

yongjae354: I think that we can probably put this case under `TestSetGitHubPullRequestApprovalAttestation`.

adityasaky: I could see it either way, do we see a coverage difference with and without this specific case?

yongjae354: Yup, we cover an addition line where validation errors. I'll go ahead and move the test case under `TestSetGitHubPullRequestApprovalAttestation`.

adityasaky:
@patzielinski can you address the linter?

PR #1080, opened 2025-09-05
ci: Add go 1.25 and Python 3.13

patzielinski:
This PR adds support for go 1.25 in CI,  as well as Python 3.13.


.github/workflows/get-started-tests.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        go-version: ['1.24']
+        go-version: ['1.24', '1.25']
```
yongjae354, line 22: Should we also update the python version here? (at line 35)

patzielinski: Good catch!

JustinCappos (approved):

PR #1097, opened 2025-09-22
[WIP] docs: Add GAP for supporting teams and hats

yongjae354:
This PR adds a GAP for supporting teams and hats in gittuf. 
Related issue: #231

### DIFF
--- a/docs/gaps/teams-proposal/README.md
+++ b/docs/gaps/teams-proposal/README.md
@@ -1,0 +1,50 @@
+# Supporting Teams in gittuf
+
+## Metadata
+
+* **Number:** TBD
+* **Title:** Supporting Teams in gittuf
+* **Implemented:** No
+* **Withdrawn/Rejected:** No
+* **Sponsors:** Yongjae Chung (yongjae354)
+* **Contributors:**
+* **Related GAPs:** [GAP-5](/docs/gaps/5/README.md)
+* **Last Modified:** September 22, 2025
+
+## Abstract
+Currently gittuf requires identities to be limited to a single key or person. In this GAP, we aim to increase the usability by introducing the idea of Teams, where multiple identities can be grouped together to be given equal privileges.
+
+## Motivation
+Often developers work in teams, and a repository administrator (or higher-level authorized principal) may want to assign multiple entities to a single group, delegating equal authorization to make changes on the repository. [GAP-5](/docs/gaps/5/README.md) introduces the notion of a principal, which allows multiple keys to be grouped to a single entity. However, the current implementation of principals does not support grouping multiple entities into a single entity. This GAP introduces the idea of a Team, where multiple principals can be grouped into a team, and a single principal can be associated with multiple teams. 
+
+In addition to the definition of teams, the idea of "hats" must be supported. Developers may be included in different teams, and must be able to select which team to represent when signing off a change. For example, a repository contributer Alice may be part of 2 different teams: 'security' and 'dev'. These two teams may have different privileges and thresholds.
+
+
+## Specification
+A Team type contains the following fields:
+* ID - A unique identifier of the team entity
+* Principals - A set of principals associated with this team
+* Threshold - A number of approvals required by principals associated to this team
+
+TODO: add schema, examples
+TODO: add hat attestation definition, schema, examples
+
+## Reasoning
+TODO:
+- Should a team further delegating authorization downstream be allowed?
+
+## Backwards Compatibility
+
+## Security
+TODO
+Allowing hat-based authorizations might increase security risk. For example, having a principal with a high-privilege security hat and a regular developer hat increases the key compromise risk for a single key. A potential solution for this is to require Sigstore integration for multi-factor and ephemeral authorization.
+
+## Prototype Implementation
+
+## Implementation
+
+## Changelog
+
+## Acknowledgements
+
+## References
--- a/docs/gaps/teams-proposal/README.md
+++ b/docs/gaps/teams-proposal/README.md
@@ -8,58 +8,64 @@
 * **Withdrawn/Rejected:** No
 * **Sponsors:** Yongjae Chung (yongjae354)
 * **Contributors:** Patrick Zielinski (patzielinski), Isabel Ayres (i-ayres),
-  Peter Ye (PeterJYE)
+  Peter Ye (PeterJYE), Reza Curtmola (reza-curtmola), Justin Cappos (
+  JustinCappos)
 * **Related GAPs:** [GAP-5](/docs/gaps/5/README.md)
-* **Last Modified:** October 16, 2025
+* **Last Modified:** November 8, 2025
 
 ## Abstract
-Currently gittuf doesn't have a convenient way to group multiple identities
-together. In this GAP, we aim to increase the usability and scalability of
-gittuf by introducing the idea of teams, where multiple identities can be
-grouped together to be given equal privileges.
+
+Currently, gittuf doesn't have a convenient way to group multiple persons
+together and represent them as a single entity. This GAP aims to enhance the
+usability of gittuf by introducing two new features. First is the new team
+principal type, which represents a group of persons in the gittuf policy
+metadata. Second is the support for a gittuf user to "wear a hat", allowing them
+to specify a team to sign off on behalf of.
 
 ## Motivation
+
 In certain scenarios, repository administrators may need to delegate
-authorization to a group of people. This group may be a team of developers,
-release managers, lawyers, QA, security, or etc., depending on the delegated
-namespace. [GAP-5](/docs/gaps/5/README.md) introduces the notion of a principal,
-which allows multiple keys to be grouped to a single entity. However, the
-current implementation of principals does not support grouping multiple entities
-into a single entity. This makes it difficult for a delegating entity to
-identify groups within a delegation (since they are implicit), as well as to
-modify all associated delegations when someone needs to be added to or removed
-from a group. With teams in gittuf, the delegating entity only needs to run an
-update to their team definition once, which will affect all delegations where
-the team is authorized.
+authorization to a group: a set of multiple users that serve the same role in a
+repository. This may be a team of developers, release managers, lawyers, QA,
+security, or etc. [GAP-5](/docs/gaps/5/README.md) introduces the notion of a
+person principal, which allows multiple keys to be associated with a single
+entity. However, the current implementation of principals does not support
+grouping multiple persons into a single entity. This makes it difficult for a
+delegating entity to identify a group in the system, since they are defined
+implicitly (see [Reasoning](#reasoning)). Also, it is inefficient and
+error-prone to modify all associated delegations when a person needs to be added
+to or removed from a group. With explicit team definitions, the delegating
+entity only needs to run an update to their team definition once, which will
+affect all delegations where the team is authorized.
 
-This GAP introduces the idea of a team in gittuf, in which multiple identities
-can be grouped together as part of a single team, and a single identity can
-be associated with multiple teams. 
-
-In addition to the definition of teams, the idea of wearing different "hats"
-must be supported. In some cases, people may be assigned to multiple teams, and
-must be able to select which team to represent when signing off a change using
-gittuf. For instance, a repository contributer Alice is a in 2 different teams:
-'security' and 'dev', where these two teams have different privileges and
-thresholds. In this scenario, Alice would wear her 'dev' hat during normal
-updates, and wear her 'security' hat only for emergency updates that require
-immediate action. As seen here, hats are a way for a person to select which team
-they wish to represent when signing off a change.
+Additionally in some cases, a gittuf user may need to escalate their own
+privileges under certain circumstances. For example, when a critical security
+vulnerability is discovered, an emergency patch may need to be pushed directly
+to the production branch although the usual process requires multiple signatures
+under the branch protection rule. For urgent updates as such, a project may
+benefit from a privileged 'security' team principal that can access the branch
+without further approvals. To support this behavior, as well as the general
+capability for a person to be assigned to multiple teams in different contexts,
+we define a "hat", which specifies the team principal that the gittuf user is
+signing on behalf of.
 
 ## Specification
+
 A team type contains the following fields:
-* ID - A unique identifier of the team entity
+
+* ID - A unique name or identifier of the team entity
 * Principals - A set of principals associated with this team
-* Threshold - A number of approvals required by principals associated to this
+* Threshold - A number of approvals required by principals associated with this
   team
 
 The schema of a team would look like the following.
+
 ```
 "<id>": {
     "teamID": "<id>"
     "principals": {
-        "<personID-1>": {
-            "personID": "<personID-1>",
+        "<personID1>": {
+            "personID": "<personID1>",
             "publicKeys": {
                 "<keyID1>": {
                     "keyid": "<keyID1>",
@@ -71,7 +77,7 @@ The schema of a team would look like the following.
                     },
                     "scheme": "fulcio"
                 },
-                "<keyid2>>": ...
+                "<keyID2>>": ...
             },
             "associatedIdentities": {
                 "<integrationID1>": "<usernameA>",
@@ -80,7 +86,7 @@ The schema of a team would look like the following.
             }
             "customMetadata": {...}
         },
-        "<personID-2>": {
+        "<personID2>": {
             ...
         },
         ...
@@ -89,15 +95,64 @@ The schema of a team would look like the following.
 }
 ```
 
-A hat attestation is used when a user wishes to approve a change by attesting on
-behalf of a team. A hat authorization contains the following fields:
-* TargetRef - The name of the Git reference that will be updated
-* FromID - The ID of the Git reference before the change is applied.
-* TargetID - The ID of the Git reference after the change is applied.
-* PrincipalID - The attesting user's principal ID.
-* Hat - The attesting user's team they are attesting on behalf of.
+The team ID can be any arbitrary string (likely the name of the team) that
+uniquely identifies the team within the policy. Note that a team threshold must
+be satisfied in order for the team to count towards a rule threshold during
+verification.
 
-Below is an example schema of a hat attestation.
+### Definition of a Hat
+
+A `hat` is a specifier that indicates the intended team ID of a person, when
+they sign on behalf of a team.
+
+### Updating a Team Definition
+
+When a team's membership or threshold must be updated, the delegating entity of
+the team must update their delegation. The members of the team are not trusted
+to update the team's definitions. Additionally, if a member's key must be
+rotated, the delegating entity must update their delegation with the member's
+key information.
+
+### Commiting to the repository on behalf of a team
+
+For users to indicate that they have made a change on the repository on behalf
+of a team's authorization, we have two options. The implementation approach is
+not decided as of now.
+
+#### Option 1 - Encode the hat information in the commit message
+
+When a gittuf user makes a commit to the repository on behalf of a team, they
+must encode their current hat information in their Git commit message. An
+example commit message may look like this:
+
+```
+Fix: Documentation
+This commit fixes typos in README
+
+hat: dev
+```
+
+This implies that during verification, gittuf must inspect the commit object
+pointed by the RSL and parse the commit message to check for hat information.
+
+#### Option 2 - Define a new attestation type for attesting a user's hat information for commits
+
+When a gittuf user makes a commit to the repository on behalf of a team, they
+must create an attestation for the hat they were wearing at the time of the
+commit. We call this new attestation type a `hat attestation`. A hat attestation
+structure contains the following fields.
+
+* TargetID - The ID of the Git reference pointing to the user's Git commit.
+* PrincipalID - The user's principal ID.
+* Hat - The user's team ID they are committing on behalf of.
+
+### Making an approval on behalf of a team
+
+To allow users to make an approval on behalf of a team, we introduce an optional
+field to the existing authorization attestation structure to account for hat
+information. An example of an authorization attestation with hat information
+looks like the following.
+
 ```
 "refs/heads/main": {
     "targetRef": "refs/heads/main",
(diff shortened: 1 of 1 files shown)


docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,104 @@
...
+* **Withdrawn/Rejected:** No
+* **Sponsors:** Yongjae Chung (yongjae354)
+* **Contributors:**
+* **Related GAPs:** [GAP-5](/docs/gaps/5/README.md)
+* **Last Modified:** September 22, 2025
+
+## Abstract
+Currently gittuf requires identities to be limited to a single key or person. In
+this GAP, we aim to increase the usability by introducing the idea of Teams,
+where multiple identities can be grouped together to be given equal privileges.
```
JustinCappos: The real motivation here is that many different folks can delegate to people within some sort of group (release managers, lawyers, QA, security team, etc.) and they do not need to all change their delegations when someone is added / removed from that group.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,104 @@
+# Supporting Teams in gittuf
```
JustinCappos: If you're supporting hats, mention this here too.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,175 @@
+# Supporting Teams and Hats in gittuf
+
+## Metadata
+
+* **Number:** TBD
+* **Title:** Supporting Teams and Hats in gittuf
+* **Implemented:** No
+* **Withdrawn/Rejected:** No
+* **Sponsors:** Yongjae Chung (yongjae354)
+* **Contributors:**
```
patzielinski: We might want to consult the others on the effort for whether they'd like to be listed as contributors.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,175 @@
...
+This GAP introduces the idea of a Team in gittuf, in which multiple principals
+can be grouped together as part of a single entity, and a single principal can
+be associated with multiple teams. In addition to the definition of teams, the
+idea of "Hats" must be supported. some people may be assigned to multiple teams,
+and must be able to select which team to represent when signing off a change
+using gittuf. For example, a repository contributer Alice may be part of 2
+different teams: 'security' and 'dev'. These two teams may have different
+privileges and thresholds.
+
+
```
patzielinski, line 51:
```suggestion
```

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,175 @@
...
+* **Title:** Supporting Teams and Hats in gittuf
+* **Implemented:** No
+* **Withdrawn/Rejected:** No
+* **Sponsors:** Yongjae Chung (yongjae354)
+* **Contributors:**
+* **Related GAPs:** [GAP-5](/docs/gaps/5/README.md)
+* **Last Modified:** October 3, 2025
+
+## Abstract
+Currently gittuf requires identities to be limited to a single key or person. In
```
patzielinski: I'd rephrase this to say that gittuf is unable to group identities together; I think it flows better with the idea introduced on line 17.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,175 @@
...
+this GAP, we aim to increase the usability and scalability of gittuf by
+introducing the idea of Teams, where multiple identities can be grouped together
+to be given equal privileges.
+
+## Motivation
+In some usecases of gittuf, repository administrators may need to delegate
```
patzielinski: I'd expand this to say that this applies in certain scenarios, rather than just usecases of gittuf.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+
+## Specification
+A Team type contains the following fields:
+* ID - A unique identifier of the team entity
+* Principals - A set of principals associated with this team
+* Threshold - A number of approvals required by principals associated to this
+  team
+
+The schema of a team would look like the following.
```
patzielinski, line 61: I'm fine with either capitalizing this or keeping it lowercase, but just make sure to keep it uniform across the document.

```suggestion
The schema of a Team would look like the following.
```

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,175 @@
...
+            ...
+        },
+        ...
+    }
+    "threshold": "<int>"
+}
+```
+
+When a user in a team wishes to approve a change, they attest using a
+`ReferenceAuthorizationWithHat`.
```
patzielinski: This needs some motivation, why is an authorization needed? I'd rework this to also stay away from the Go implementation, rather focus on what a hat authorization is, similar to https://github.com/gittuf/gittuf/blob/main/docs/gaps/3/README.md#specification.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,175 @@
...
+key is now higher for Alice, as obtaining Alice's private keys for developer hat
+access also implies access to the security hat. A potential solution for this
+problem is to require Sigstore integration for users that are members of teams,
+enforcing multi-factor and ephemeral authorization.
+
+### Counting a Team towards a Threshold
+Similar to what has been discussed in [GAP-5](/docs/gaps/5/README.md), adding
+teams introduces some additional complexity during verification. 
+
+> The workflow must now track whether a key was used for another previously
```
patzielinski: Is the `>` needed here?

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,175 @@
...
+user's team inclusion is approved latter than a user's hat attestation?)*
+
+### TODO: Using Teams to Delegate Authorization
+
+## Backwards Compatibility
+Introducing teams and hat selection in gittuf must be backwards compatible, as
+the previous metadata definitions and verification flows of keys and principals
+remain the same. It is up to the repository administrators to choose between
+using previous key-based or principal-based metadata for smaller projects, and
+incorporating teams in larger projects with groups of collaborators.
```
patzielinski: We must be careful here, administrators may choose to decide whether to use teams, but the gittuf project will likely advance its metadata like we did with the transition from keys to principals, where users may still use the original `add-key` function (albeit its use is discouraged).

docs/gaps/teams-proposal/README.md
yongjae354: I wanted to quote the text from GAP-5; should I rather put it around ""?

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,175 @@
...
+the team is authorized.
+
+This GAP introduces the idea of a Team in gittuf, in which multiple principals
+can be grouped together as part of a single entity, and a single principal can
+be associated with multiple teams. In addition to the definition of teams, the
+idea of "Hats" must be supported. some people may be assigned to multiple teams,
+and must be able to select which team to represent when signing off a change
+using gittuf. For example, a repository contributer Alice may be part of 2
+different teams: 'security' and 'dev'. These two teams may have different
+privileges and thresholds.
```
MI2007: To link the "some... thresholds"(Lines 37-41) sentence with the idea of Hats, you should add a sentence at the end like "Hats would allow for the person to select which team to represent when signing off a change". Although it is implied that the sentences are describing the function of Hats, it isn't explicitly stated, so I think adding that sentence can clarify that. (Not too necessary though)

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,175 @@
...
+identify groups within a delegation (since they are implicit), as well as to
+modify all associated delegations when someone needs to be added to or removed
+from a group. With teams in gittuf, the delegating entity only needs to run an
+update to their team definition once, which will affect all delegations where
+the team is authorized.
+
+This GAP introduces the idea of a Team in gittuf, in which multiple principals
+can be grouped together as part of a single entity, and a single principal can
+be associated with multiple teams. In addition to the definition of teams, the
+idea of "Hats" must be supported. some people may be assigned to multiple teams,
```
MI2007:
```suggestion
idea of "Hats" must be supported. Some people may be assigned to multiple teams,
```

MI2007: Also, starting the sentence like "In some cases, people may be assigned to multiple teams" may sound better

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,175 @@
...
+	TargetRef   string `json:"targetRef"`
+	FromID      string `json:"fromID"`
+	TargetID    string `json:"targetID"`
+	PrincipalID string `json:"principalID"`
+	Hat         string `json:"hat"`
+}
+```
+A ReferenceAuthorizationWithHat only counts towards a team threshold, not a
+  general rule threshold. If a user attests indicating a hat, gittuf assumes
+  they are attesting as part of a team, not as an individual principal.
```
MI2007: Should clarify why to use ReferenceAuthorizationWithHat

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+into a single entity. This makes it difficult for a delegating entity to
+identify groups within a delegation (since they are implicit), as well as to
+modify all associated delegations when someone needs to be added to or removed
+from a group. With teams in gittuf, the delegating entity only needs to run an
+update to their team definition once, which will affect all delegations where
+the team is authorized.
+
+This GAP introduces the idea of a team in gittuf, in which multiple identities
+can be grouped together as part of a single team, and a single identity can
+be associated with multiple teams.
```
MI2007: "and... teams" (Lines 37-38) can be removed  since you talk about hats and identities being associated with multiple teams in the next paragraph.

MI2007:
```suggestion

```

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+current implementation of principals does not support grouping multiple entities
+into a single entity. This makes it difficult for a delegating entity to
+identify groups within a delegation (since they are implicit), as well as to
+modify all associated delegations when someone needs to be added to or removed
+from a group. With teams in gittuf, the delegating entity only needs to run an
+update to their team definition once, which will affect all delegations where
+the team is authorized.
+
+This GAP introduces the idea of a team in gittuf, in which multiple identities
+can be grouped together as part of a single team, and a single identity can
```
MI2007:
```suggestion
can be grouped together as part of a single team.
```

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+be associated with multiple teams. 
+
+In addition to the definition of teams, the idea of wearing different "hats"
+must be supported. In some cases, people may be assigned to multiple teams, and
```
MI2007: I would rephrase the first sentence in this paragraph. Maybe something like "With the introduction of teams, the idea of wearing different "hats" must also be implemented."

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+must be able to select which team to represent when signing off a change using
+gittuf. For instance, a repository contributer Alice is a in 2 different teams:
+'security' and 'dev', where these two teams have different privileges and
+thresholds. In this scenario, Alice would wear her 'dev' hat during normal
+updates, and wear her 'security' hat only for emergency updates that require
+immediate action. As seen here, hats are a way for a person to select which team
+they wish to represent when signing off a change.
```
MI2007: You can add a sentence concluding how hats are important. "They are imperative for entities associated with multiple teams to select which team they want to represent.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Specification
+A team type contains the following fields:
+* ID - A unique identifier of the team entity
+* Principals - A set of principals associated with this team
+* Threshold - A number of approvals required by principals associated to this
+  team
```
MI2007:
```suggestion
  team to sign off on a change.
```

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Metadata
+
+* **Number:** TBD
+* **Title:** Supporting Teams and Hats in gittuf
+* **Implemented:** No
+* **Withdrawn/Rejected:** No
+* **Sponsors:** Yongjae Chung (yongjae354)
+* **Contributors:** Patrick Zielinski (patzielinski), Isabel Ayres (i-ayres),
+  Peter Ye (PeterJYE)
```
JustinCappos:
```suggestion
  Peter Ye (PeterJYE), Justin Cappos (JustinCappos)
```

I believe I had the original motivation and pitch for this as a feature.  I should probably be listed to show I'm somewhat conflicted when reviewing.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+* **Related GAPs:** [GAP-5](/docs/gaps/5/README.md)
+* **Last Modified:** October 16, 2025
+
+## Abstract
+Currently gittuf doesn't have a convenient way to group multiple identities
+together. In this GAP, we aim to increase the usability and scalability of
+gittuf by introducing the idea of teams, where multiple identities can be
+grouped together to be given equal privileges.
```
JustinCappos: There is no motivation for hats here.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Motivation
+In certain scenarios, repository administrators may need to delegate
+authorization to a group of people. This group may be a team of developers,
+release managers, lawyers, QA, security, or etc., depending on the delegated
+namespace. [GAP-5](/docs/gaps/5/README.md) introduces the notion of a principal,
+which allows multiple keys to be grouped to a single entity. However, the
+current implementation of principals does not support grouping multiple entities
+into a single entity. This makes it difficult for a delegating entity to
+identify groups within a delegation (since they are implicit), as well as to
```
JustinCappos: You're using the term groups without defining what you mean sufficiently.

docs/gaps/teams-proposal/README.md
JustinCappos: Is a team a group?  Are they different?  How?

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+the team is authorized.
+
+This GAP introduces the idea of a team in gittuf, in which multiple identities
+can be grouped together as part of a single team, and a single identity can
+be associated with multiple teams. 
+
+In addition to the definition of teams, the idea of wearing different "hats"
+must be supported. In some cases, people may be assigned to multiple teams, and
+must be able to select which team to represent when signing off a change using
+gittuf. For instance, a repository contributer Alice is a in 2 different teams:
```
JustinCappos: Is a contributor a developer?  How are they different?

(in other words, pick one technical term and repeat it.  Synonyms are really bad for technical documents.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+'security' and 'dev', where these two teams have different privileges and
+thresholds. In this scenario, Alice would wear her 'dev' hat during normal
```
JustinCappos: Why is it clear this is correct in this scenario?  Can you make the example more self-evident?

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+updates, and wear her 'security' hat only for emergency updates that require
+immediate action. As seen here, hats are a way for a person to select which team
```
JustinCappos: You should motivate this up front.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+}
+```
+
+A hat attestation is used when a user wishes to approve a change by attesting on
+behalf of a team. A hat authorization contains the following fields:
+* TargetRef - The name of the Git reference that will be updated
+* FromID - The ID of the Git reference before the change is applied.
+* TargetID - The ID of the Git reference after the change is applied.
+* PrincipalID - The attesting user's principal ID.
+* Hat - The attesting user's team they are attesting on behalf of.
```
JustinCappos: So, they do this after the fact?  Or would it make sense to have the hat as an optional field of their existing way to generate an attestation?  

What if someone makes an attestation without a hat?  What happens?  Is this all teams, no teams, a specific team, etc.?

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+<!-- 
+Example:
+Protect-main requires 2 signatures (threshold=2). 
+Authorized entities: {dev-team, Carol}
+- dev-team: Principals = {Alice, Bob}, threshold = 2
+
+Cases to consider:
+- Alice signs off on her commit on main, on behalf of dev-team.
+- Alice authorizes Bob's commit on main on behalf of dev-team.
+- Alice authorizes Carol's change on the branch; Carol is not on dev-team, is a separate identity defined on rule.
```
JustinCappos: I can't parse the last part of this line

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+gittuf. For instance, a repository contributer Alice is a in 2 different teams:
+'security' and 'dev', where these two teams have different privileges and
+thresholds. In this scenario, Alice would wear her 'dev' hat during normal
+updates, and wear her 'security' hat only for emergency updates that require
+immediate action. As seen here, hats are a way for a person to select which team
+they wish to represent when signing off a change.
+
+## Specification
+A team type contains the following fields:
+* ID - A unique identifier of the team entity
```
JustinCappos: What sort of ID is this?  Is it human readable?  Does it mean anything across repos?  Is it only for a single repo?

docs/gaps/teams-proposal/README.md
JustinCappos, line 59: How do you update the team membership?  This clearly shouldn't be forced to be the same threshold as signing on behalf of the team.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+When a user attests using a hat attestation, gittuf assumes they are attesting
+on behalf of a team, not as an individual person on the metadata. In other
+words, a hat attestation only counts toward a team threshold, not a rule
+threshold during verification. 
+
+## Reasoning
+
+### TODO: Verification Workflow with Teams & Hats
```
JustinCappos: I'm not really sure what is happening here.  Why the TODOs?  Is this ready for review or early stage?

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+## Security
+
+### Addressing Additional Key Compromise Risk
+Allowing hat-based authorizations may increase the risk of key compromise for a
+single user key. For example, assume a user Alice, with a high-privilege
+security hat and a regular developer hat. The key compromise risk for a single
+key is now higher for Alice, as obtaining Alice's private keys for developer hat
+access also implies access to the security hat. A potential solution for this
+problem is to require Sigstore integration for users that are members of teams,
+enforcing multi-factor and ephemeral authorization.
```
JustinCappos: I'm less sure this is true.  Alice was already trusted before in this case, almost certainly.  This just stops her signature from being always trusted.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+both have the same key associated with them.
+
+For team thresholds, since we don't allow for nested teams (a team consisting
+another team as a member), the above workflow still applies. That is, counting
+signatures towards team thresholds will still track keys and make sure they
+cannot be used for two different principals.
+
+For rule thresholds, the system must additionally track whether a key was used
+for a previously examined team, to avoid counting two entities that use the same
+underlying key, towards the threshold.
```
JustinCappos: Why?  Can you justify this and provide a realistic example?

JustinCappos (changes requested):
I think we need some edits / changes here to clarify and flesh this out

docs/gaps/teams-proposal/README.md
yongjae354: @JustinCappos My thought was that add/remove/updating a team should depend on the delegating entity's configuration. In this example in the design document, we have thresholds of 2 for the root of trust, 2 for the primary rule file,
So in this case initializing a team in the primary rule file should require 2 approvals from the root principals, initializing a team for a specific namespaced rule would require 2 approvals from primary rule file principals, ... and so on so forth? 

https://github.com/gittuf/gittuf/blob/8357a47c0aeccafd64711d4516eafe2736761ed5/docs/design-document.md?plain=1#L214-L245


I'll make this aspect more clear in the Specification as well!

docs/gaps/teams-proposal/README.md
yongjae354: Yes, I'm using the term "team" to represent a group of entities in the system. I'll make this a bit more clear in the motivation section.

docs/gaps/teams-proposal/README.md
yongjae354: I'm not too sure about this at the moment. For principals, it can be a name or an email address, so I was assuming it would be the same for teams. Would this have any implications though when we decide to use these IDs outside a single repo?

patzielinski: Principal identifiers are (in gittuf) only valid for the repository they are defined in. I figure a team identifier can be any arbitrary string, but the administrator adding the team definition will likely just use a name, e.g. `security-response-team`.

docs/gaps/teams-proposal/README.md
yongjae354: Do you mean that the Sigstore solution would not be feasible?

docs/gaps/teams-proposal/README.md
patzielinski: For existing gittuf attestations, it is trivial to utilize a new version of the attestation that will also encode hat data (i.e. a person approving a change to be merged into the `main` branch will simply just indicate in their approval invocation that they are doing it on behalf of some team).

The issue is when a user makes a commit to the repository. Commits can be signed by users (and we utilize this information in gittuf), but the question is where to encode hat information.

Do we do it in the commit as a trailer? e.g. add a line at the end that says `Hat-ID: security-response-team`

That could be a way to do it, but we need to think about if there are cases where this could be problematic...

docs/gaps/teams-proposal/README.md
patzielinski: I'd add a line that says something like "As stated in GAP-5" or similar.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,188 @@
...
+* **Title:** Supporting Teams and Hats in gittuf
+* **Implemented:** No
+* **Withdrawn/Rejected:** No
+* **Sponsors:** Yongjae Chung (yongjae354)
+* **Contributors:** Patrick Zielinski (patzielinski), Isabel Ayres (i-ayres),
+  Peter Ye (PeterJYE)
+* **Related GAPs:** [GAP-5](/docs/gaps/5/README.md)
+* **Last Modified:** October 16, 2025
+
+## Abstract
```
patzielinski, line 16:
```suggestion
## Abstract

```

Make sure to put a newline between section headers.

docs/gaps/teams-proposal/README.md
yongjae354: Sorry about the confusion;  this was meant to be early stage as I wasn't sure how the hats design would turn out. I'm going to update this once I get a better idea of the approach!

docs/gaps/teams-proposal/README.md
yongjae354: I also think it makes sense for the hat info to be an optional field in the existing approval attestation structure, so that attestations without that field will just act on the single-entity principal, and be backwards compatible that way.

Another option for user commits would be making a specific attestation type for only those made on behalf of a team, which is how we implemented it in the current prototype. It would be a simpler structure containing:
* TargetID - The ID of the Git reference pointing to the user's Git commit.
* PrincipalID - The user's principal ID.
* Hat - The user's team ID they are committing on behalf of.

docs/gaps/teams-proposal/README.md
JustinCappos: I might be confused here.  Please fix risk vs impact and I'll take a fresh look later.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,310 @@
...
+* **Related GAPs:** [GAP-5](/docs/gaps/5/README.md)
+* **Last Modified:** November 8, 2025
+
+## Abstract
+
+Currently, gittuf doesn't have a convenient way to group multiple persons
+together and represent them as a single entity. This GAP aims to enhance the
+usability of gittuf by introducing two new features. First is the new team
+principal type, which represents a group of persons in the gittuf policy
+metadata. Second is the support for a gittuf user to "wear a hat", allowing them
```
JustinCappos, line 22: a user who is part of multiple different teams needs to be able to indicate which "hat" they ware wearing when signing.  I.e. which team they intend their signature should be used for.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,310 @@
...
+to or removed from a group. With explicit team definitions, the delegating
+entity only needs to run an update to their team definition once, which will
+affect all delegations where the team is authorized.
+
+Additionally in some cases, a gittuf user may need to escalate their own
+privileges under certain circumstances. For example, when a critical security
+vulnerability is discovered, an emergency patch may need to be pushed directly
+to the production branch although the usual process requires multiple signatures
+under the branch protection rule. For urgent updates as such, a project may
+benefit from a privileged 'security' team principal that can access the branch
```
JustinCappos, line 46:
```suggestion
benefit from a privileged 'bypass' team principal that can access the branch
```

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,310 @@
...
+without further approvals. To support this behavior, as well as the general
+capability for a person to be assigned to multiple teams in different contexts,
+we define a "hat", which specifies the team principal that the gittuf user is
```
JustinCappos, line 49: This last sentence should be a separate paragraph that then motivates this sub problem.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,310 @@
...
+```
+
+The team ID can be any arbitrary string (likely the name of the team) that
+uniquely identifies the team within the policy. Note that a team threshold must
+be satisfied in order for the team to count towards a rule threshold during
+verification.
+
+### Definition of a Hat
+
+A `hat` is a specifier that indicates the intended team ID of a person, when
```
JustinCappos, line 105: You should explain the context, etc. here a bit more.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,310 @@
...
+they sign on behalf of a team.
+
+### Updating a Team Definition
+
+When a team's membership or threshold must be updated, the delegating entity of
```
JustinCappos, line 110: Note that we have a TAP that does support this ([key self-rotation](https://github.com/theupdateframework/taps/blob/master/tap8.md)).  Why not support this also?

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,310 @@
...
+the team must update their delegation. The members of the team are not trusted
+to update the team's definitions. Additionally, if a member's key must be
+rotated, the delegating entity must update their delegation with the member's
+key information.
+
+### Commiting to the repository on behalf of a team
+
+For users to indicate that they have made a change on the repository on behalf
+of a team's authorization, we have two options. The implementation approach is
+not decided as of now.
```
JustinCappos, line 120: what are the pros and cons?

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,310 @@
...
+information. An example of an authorization attestation with hat information
+looks like the following.
+
+```
+"refs/heads/main": {
+    "targetRef": "refs/heads/main",
+    "fromID": "<sha-1 hash>",
+    "targetID": "<sha-1 hash>",
+    "principalID": "<person-ID-1>",
+    "hat": "security",
```
JustinCappos, line 162: Is it a good idea to have the team namespace and user identity namespace be the same?  What if I make my identity name be "security" and someone creates a team "security" later?

Note, upon looking at this again, I think these namespaces may be separate already.  If so, this should be clarified including why this doesn't matter.

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,310 @@
...
+                        git:refs/heads/prod}
+        -> (2, {Alice, Bob, Carol})
+    protect-ios-app: {file:ios/*}
+        -> (1, {Alice})
+    protect-android-app: {file:android/*}
+    -> (1, {Bob})
+
+ruleFile: protect-ios-app
+principals: {{ios-team: 
+                principals={Dana, George},
```
JustinCappos, line 215: If you have to list all of the principals and the threshold, what is the point of naming the team ios-team?

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+As stated in [GAP-5](/docs/gaps/5/README.md): "The workflow must now track
+whether a key was used for another previously examined principal to avoid
+counting a single signature for two principals who both have the same key
+associated with them."
+
+For team thresholds, since gittuf does not support nested teams (a team
+consisted of another team as its member), the above workflow still applies. That
+is, counting signatures towards team thresholds must still track keys and make
+sure the same key cannot be used for more than 1 person on the team.
```
JustinCappos, line 273: Can multiple identities add the same key?  Is this the case you are concerned about?

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+For rule thresholds, the verification flow behaves a bit more leniently. This
+means that it does not explicitly track whether a person's signature was
+consumed for a previously examined team, allowing itself to count more than 1
+entity that use the same underlying key toward the rule threshold. Enabling this
+behavior can be useful in certain cases. For example, if Bob is responsible for
+both development and compliance on the branch, it may make sense to allow for
+Bob to sign on behalf of both teams. gittuf ultimately leaves this decision up
+to the delegating entity to decide.
+
```
JustinCappos, line 283: This question may be too general but, how does all of this relate to something like MFA where we may want to be able to have some MFA system also sign something / provide a temporal key?  So, if I want to bypass, what if it makes me use additional authentication mechanisms internally?  Can this be captured / used?

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,310 @@
...
+### Addressing the Additional Impact of Key Compromise
+
+Allowing hat-based authorizations may increase the impact of key compromise for
+a single user key. For example, assume a user Alice with a privileged security
+hat and a regular developer hat. The potential impact from a key compromise for
+a single key owned by Alice is now higher, as obtaining Alice's private keys for
+developer hat access also implies access to the security hat. A potential
+solution for this problem is to encourage Sigstore integration for users that
+are members of teams, enforcing multifactor and ephemeral authorization. The
```
JustinCappos, line 292: is this sufficient?  Does this just become a single point of failure?

docs/gaps/teams-proposal/README.md
```diff
@@ -0,0 +1,310 @@
...
+alternative is for a user to use different keys for each hat, which reduces the
```
JustinCappos, line 293: If you do this, do you need teams and hats?  Isn't the whole point of this is to let people use their key in different ways.  If I have a different key for each team, why do we use this?

JustinCappos (commented):
Wide ranging comments.  Much better overall

patzielinski:
Closing this for now due to possible future CEDAR work. We can revisit/reopen this depending on how that work lets us express teams/hats.

JustinCappos:
This is fine, but as this more directly impacts the TUF policy language, I
do expect this will be needed.

yongjae354:
I've been outside the conversations for a while so I'm not aware of the policy changes with cedar, but please let me know if you want any updates from me.

PR #1098, opened 2025-09-23
docs: Add document on gittuf recovery

patzielinski:
This PR adds a document describing the process for manually recovering a repository to a known-good state should a violation occur.

Automated repair is on the way, but I think it's good to have this documented nonetheless.

### DIFF
--- a/docs/development/gittuf-recovery.md
+++ b/docs/development/gittuf-recovery.md
@@ -1,0 +1,169 @@
+# Recovering from Verification Errors in gittuf
+
+From time to time, there may be a scenario where gittuf verification fails on
+the repository, either due to a malicious change or an issue with insufficient
+signatures on a legitimate change.
+
+In order to return the repository to a state where it may pass gittuf
+verification again, a recovery procedure must be performed. This procedure may
+be performed either automatically (to be added in a future update to gittuf), or
+manually, by following the steps detailed below.
+
+This guide assumes at least some familiarity with gittuf's design, with respect
+to components such as the Reference State Log (RSL). See the [gittuf Design
+Document](/docs/design-document.md) for more information.
+
+Manual recovery of a repository follows a three-step process. This process also
+assumes that your signing key is specified in the policy as authorized to make
+changes to the reference in question.
+
+## 1. Revoking the Problematic Entry (or Entries)
+
+First, you must identify the oldest RSL entry that causes gittuf verification to
+fail. To do this, you must run verification with the `--verbose` flag and then
+analyze the log output for the problematic entry.
+
+Take the example of running `gittuf verify-ref --verbose main` on the official
+gittuf repository:
+
+```bash
+time=2025-09-22T15:10:56.557-04:00 level=DEBUG msg="Verifying signature of Git object with ID '235312e7f06c75d4be0ffb3ff87388f10be9882f'..."
+time=2025-09-22T15:10:56.569-04:00 level=DEBUG msg="Proceeding with verification of attestations..."
+time=2025-09-22T15:10:56.569-04:00 level=DEBUG msg="Using approvers from code review tool attestations..."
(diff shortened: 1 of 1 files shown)


docs/development/gittuf-recovery.md
```diff
@@ -0,0 +1,155 @@
...
+
+This guide assumes at least some familiarity with gittuf's design, with respect
+to components such as the Reference State Log (RSL). See the [gittuf Design
+Document](/docs/design-document.md) for more information.
+
+Manual recovery of a repository follows a three-step process. This process also
+assumes that your signing key is specified in the policy as authorized to make
+changes to the reference in question.
+
+## 1. Revoking the Problematic Entry
```
adityasaky: Any subsequent entries to the branch in question also need to be revoked

docs/development/gittuf-recovery.md
```diff
@@ -0,0 +1,155 @@
...
+  Number: 458
+```
+
+To record an RSL entry for this entry again, you should rewind the branch in
+question until the latest commit on the branch is the one preceding any invalid
+ones. In our case, we must rewind the `main` branch by one commit. To do this,
+you may use the `update-ref` command from Git:
+
+```bash
+git update-ref main main~1 # Where the number after the ~ is the number of commits to rewind by.
```
adityasaky: Maybe best to update-ref to the commit in question

docs/development/gittuf-recovery.md
```diff
@@ -0,0 +1,155 @@
...
+
+```bash
+gittuf rsl record --local-only main
+```
+
+Finally, pull the invalid changes back onto the main branch with `git pull`.
+
+## 3. Recording the Latest State
+
+Once the previous-good commit has been recorded again in the RSL, you must
```
adityasaky, line 150: The changes are already in `main`. This is about recording the "actually good" changes back in the RSL. It's important to note that this subsequent RSL entry needs to be signed by someone who can meet the policy threshold. Since this is a document for gittuf/gittuf only, maybe some of those details should be added in?

adityasaky: Alternatively, we can do it "right" by actually:
a) reverting changes in a PR after the revocation, back to last good state
b) reverting the revert in a PR to apply changes -- this only works when the issue is due to API timing errors or similar as it assumes all changes will be reapplied

docs/development/gittuf-recovery.md
```diff
@@ -0,0 +1,159 @@
...
+  Number: 461
+
+    Annotation ID: 194200eb770c8025170dd9bcf2dc89f760da8017
+    Skip:          yes
+    Number:        474
+    Message:
+      GitHub API timing issues
+```
+
+You must also check for any later RSL entries for the branch in question, and
```
adityasaky: They can be part of the same annotation entry, just list all entry IDs in the command. I usually prefer this tbh.

docs/development/gittuf-recovery.md
```diff
@@ -0,0 +1,159 @@
...
+ # Where the hash is the hash of the last good entry on the branch.
+```
+
+Next, you must record the RSL entry with `gittuf rsl record`:
+
+```bash
+gittuf rsl record --local-only main
+```
+
+Finally, pull the invalid changes back onto the main branch with `git pull`.
```
adityasaky: I advice against this a this can sync the RSL when the transport is used no? We can just track the "invalid" change commit ID in an env var temporarily or so?

docs/development/gittuf-recovery.md
```diff
@@ -0,0 +1,159 @@
...
+RSL entry must be signed by a user whose signatures count towards the threshold
+required in gittuf policy for the branch in question.** For example, if Alice
+and Bob are both required to sign changes that appear on the main branch, either
+Alice or Bob must sign this RSL entry.
+
+To do this, record the changes again with `gittuf rsl record`. In our case, this
+will be the same command again as above:
+
+```bash
+gittuf rsl record --local-only main
```
adityasaky, line 161: Don't forget to sync the RSL after the fixes!

PR #1099, opened 2025-09-23
[WIP] *: Enable automated recovery of gittuf repositories

patzielinski:
This PR adds an automated recovery worklflow that allows users to restore their repository to a known-good state should verification fail.

**Work in progress.**


patzielinski:
@adityasaky, this is likely in need of improvements, but any thoughts would be welcome :slightly_smiling_face:

Issue #684, opened 2024-12-12
policy: minor spacing issue with list-rules and list-principals


haotran-california:
### What happened?

@patzielinski 

Not quite sure if this counts as a bug? But there's no spacing between rules/principals when running the `gittuf policy list-rules` and `gittuf policy list-principals` commands 

### Relevant log output

```Shell
gittuf policy list-rules
Rule protect-main:
    Refs affected:
        git:refs/heads/main
    Authorized keys:
        SHA256:P2ntfeC0xb9UNGqU/DKbkBdO1RcRygYn0SUQ4jh5Ncw
    Required valid signatures: 1
Rule protect-test-branch:
    Refs affected:
        git:refs/heads/test
    Authorized keys:
        SHA256:P2ntfeC0xb9UNGqU/DKbkBdO1RcRygYn0SUQ4jh5Ncw
    Required valid signatures: 1

gittuf policy list-principals
Principal SHA256:P2ntfeC0xb9UNGqU/DKbkBdO1RcRygYn0SUQ4jh5Ncw:
    Keys:
        SHA256:P2ntfeC0xb9UNGqU/DKbkBdO1RcRygYn0SUQ4jh5Ncw (ssh)
Principal SHA256:TH/MwKbgNuaEI5sfpMViFye2IrL7FgVFPL4kH2M5PJs:
    Keys:
        SHA256:TH/MwKbgNuaEI5sfpMViFye2IrL7FgVFPL4kH2M5PJs (ssh)
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

haotran-california:
@patzielinski 

Gotchu, I think I'll let the issue sit for a bit then and continue poking around.

adityasaky:
Filing a PR adding the spaces will be merged, fwiw :)

sathas1301:
Can I take this, not sure why its been open since long time for a simple new line fix.  Am I missing anything?

PR #1105, opened 2025-10-01
fix spacing issue

sathas1301:
Added new line spacing for list-rule and list-principal entries.

Files changed listprincipals.go and listrules.go.

Fixing issue #684

### DIFF
--- a/internal/cmd/policy/listprincipals/listprincipals.go
+++ b/internal/cmd/policy/listprincipals/listprincipals.go
@@ -46,6 +46,7 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
+	count := 0
 	for _, principal := range principals {
 		fmt.Printf("Principal %s:\n", principal.ID())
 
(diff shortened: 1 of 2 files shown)


internal/cmd/policy/listprincipals/listprincipals.go
```diff
@@ -61,6 +61,7 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 				fmt.Printf(strings.Repeat(indentString, 2)+"%s: %s\n", key, value)
 			}
 		}
+		fmt.Printf("\n\n")
```
patzielinski: Let's have this be just one newline. Please also make sure to not have a trailing newline added when the last item is printed out.

internal/cmd/policy/listrules/listrules.go
```diff
@@ -66,7 +66,7 @@ func (o *options) Run(cmd *cobra.Command, _ []string) error {
 			fmt.Printf(strings.Repeat("    ", curRule.Depth+2)+"%s\n", key)
 		}
 
-		fmt.Println(strings.Repeat("    ", curRule.Depth+1) + fmt.Sprintf("Required valid signatures: %d", curRule.Delegation.GetThreshold()))
+		fmt.Println(strings.Repeat("    ", curRule.Depth+1) + fmt.Sprintf("Required valid signatures: %d\n", curRule.Delegation.GetThreshold()))
```
patzielinski: Similar item here, let's omit the newline when printing out the last item in the list.

Issue #1006, opened 2025-05-22
Fix worktree path resolution for repositories with detached git directories


fahdfady:
### Add a description

I noticed several TODO comments in the codebase related to handling detached git directories:

`worktree = strings.TrimSuffix(worktree, ".git") // TODO: this doesn't support detached git dir`

This pattern appears in:
- [internal/gitinterface/utils.go](https://github.com/gittuf/gittuf/blob/main/internal/gitinterface/utils.go)
- [internal/gitinterface/status.go](https://github.com/gittuf/gittuf/blob/main/internal/gitinterface/status.go)  
- [internal/gitinterface/tree.go](https://github.com/gittuf/gittuf/blob/main/internal/gitinterface/tree.go)

Instead of making assumptions about filesystem structure, could we use Git's internal commands to get the correct working tree path? For example, we could use the repository's [executor function](https://github.com/gittuf/gittuf/blob/91e944a93a50695aed5e4b86b3459a00509acf78/internal/gitinterface/repository.go#L166) to run `git rev-parse --show-toplevel`.

Before implementing this, I wanted to ask:
1. Is this the right approach for solving the detached git directory problem?
2. Are there any security considerations when using the executor function with these git commands? my first thought is that we could get injections from this approach.
3. Should I create a centralized method (like `GetWorkTree()`) that other code can use?

I'd appreciate any guidance before proceeding with a fix!

### Relevant log output if the discussion pertains to existing gittuf functionality

```Shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

fahdfady:
I'm pretty sure that the solutions is not that straightforward, so, i can make use of your help to explain the problem more clearly for me

adityasaky:
Hi @fahdfady!

There's really two things we need to consider here.
A) The .git directory may be detached, so we can't even rely on that
B) A single repositories may have multiple git worktrees (https://git-scm.com/docs/git-worktree)

I _think_ `git rev-parse --show-toplevel` would work for A, I'm unsure about B. Do you think you could test out a couple scenarios? 😄

vee1e:
Hi! Picking this up. This revisits the closed #1106, which changed how commands run but left the path-chopping logic untouched. Following @adityasaky's note there about handling detached directories more generally.

PR #1106, opened 2025-10-02
gitinterface: Add GetBlobID() and make Status() worktree-safe

fr0m-scratch:
- Added Repository.GetBlobID(ref, path) to resolve a file’s blob ID from a ref or the working tree.
  - Makes Repository.Status() more robust by avoiding global cwd changes and invoking Git with -C <worktree>.

### DIFF
--- a/internal/gitinterface/blob.go
+++ b/internal/gitinterface/blob.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"strings"
 )
 
 // ReadBlob returns the contents of the blob referenced by blobID.
@@ -41,4 +42,26 @@ func (r *Repository) WriteBlob(contents []byte) (Hash, error) {
 	}
 
 	return hash, nil
+}
+
+// GetBlobID returns the ID of the blob at the specified path in the given
+// reference. If the reference is ":", it will look for the blob at the path
+// in the current working directory of the repository.
+func (r *Repository) GetBlobID(ref, path string) (Hash, error) {
+	var fullRef string
+	if ref == ":" {
+		fullRef = ":" + path
+	} else {
+		fullRef = ref + ":" + path
+	}
+
+	stdout, err := r.executor("rev-parse", fullRef).executeString()
+	if err != nil {
+		return ZeroHash, fmt.Errorf("unable to resolve blobID for %s in %s: %w", path, ref, err)
+	}
+	blobID, err := NewHash(strings.TrimSpace(stdout))
+	if err != nil {
+		return ZeroHash, fmt.Errorf("invalid blob id: %w", err)
+	}
+	return blobID, nil
 }
--- a/internal/gitinterface/status.go
+++ b/internal/gitinterface/status.go
@@ -6,7 +6,6 @@ package gitinterface
 import (
 	"errors"
 	"fmt"
-	"os"
 	"strings"
 )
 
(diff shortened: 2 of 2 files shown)


internal/gitinterface/blob.go
```diff
@@ -42,3 +43,25 @@ func (r *Repository) WriteBlob(contents []byte) (Hash, error) {
 
 	return hash, nil
 }
+
+// GetBlobID returns the ID of the blob at the specified path in the given
+// reference. If the reference is ":", it will look for the blob at the path
+// in the current working directory of the repository.
+func (r *Repository) GetBlobID(ref, path string) (Hash, error) {
+	var fullRef string
+	if ref == ":" {
```
patzielinski, line 52:
```suggestion

	if ref == ":" {
```

internal/gitinterface/blob.go
```diff
@@ -42,3 +43,25 @@ func (r *Repository) WriteBlob(contents []byte) (Hash, error) {
...
+		fullRef = ":" + path
+	} else {
+		fullRef = ref + ":" + path
+	}
+
+	stdout, err := r.executor("rev-parse", fullRef).executeString()
+	if err != nil {
+		return ZeroHash, fmt.Errorf("unable to resolve blobID for %s in %s: %w", path, ref, err)
+	}
+	blobID, err := NewHash(strings.TrimSpace(stdout))
```
patzielinski, line 62:
```suggestion
	blobID, err := NewHash(strings.TrimSpace(stdout))

```

internal/gitinterface/blob.go
```diff
@@ -42,3 +43,25 @@ func (r *Repository) WriteBlob(contents []byte) (Hash, error) {
...
+	if err != nil {
+		return ZeroHash, fmt.Errorf("invalid blob id: %w", err)
+	}
+	return blobID, nil
```
patzielinski, line 66:
```suggestion

	return blobID, nil
```

internal/gitinterface/status.go
```diff
@@ -98,16 +97,9 @@ func (r *Repository) Status() (map[string]FileStatus, error) {
...
-	if err != nil {
-		return nil, err
-	}
-	if err := os.Chdir(worktree); err != nil {
-		return nil, err
-	}
-	defer os.Chdir(cwd) //nolint:errcheck
 
-	output, err := r.executor("status", "--porcelain=1", "-z", "--untracked-files=all", "--ignored").executeString()
+	output, err := r.executor("-C", worktree, "status", "--porcelain=1", "-z", "--untracked-files=all", "--ignored").executeString()
```
patzielinski, line 101: Following the discussion from #1006, it looks like we may need more robust logic to handle detached worktrees? Paging @adityasaky for thoughts.

adityasaky: I'm inclined to leave this to handling detached .git directories more generally.

adityasaky (commented):
Hi @fr0m-scratch can you add some more context about why we need this change in the PR description?

Should we be using the RSL to establish what revision (and thus tree object) we want to inspect for the blob in question? `git rev-parse <ref>:<filepath>` is handy but we don't seem to be considering what the ref points to?

patzielinski:
This PR is related to the Lua Sandbox part of gittuf. Given the age of this PR, I'm closing it, but please feel free to re-open.

PR #1108, opened 2025-10-03
Update get-started.md

sathas1301:
Updated documentation to reflect new way of using principals instead of keys.


docs/get-started.md
```diff
@@ -145,20 +145,19 @@ from TUF) must be signed by keys specified in the root of trust.
 gittuf trust add-policy-key -k ../keys/root --policy-key ../keys/policy.pub
 gittuf policy init -k ../keys/policy --policy-name targets
 ```
-Then, use the policy key to initialize a policy and add a rule protecting the
-`main` branch.
+Add trusted person to the policy file. Then, use the policy key to initialize a policy and add a rule protecting the
+`main` branch. 
 
 ```bash
+gittuf policy add-person --person-ID 'developer' --public-key ../keys/developer.pub
```
patzielinski: The `-k` argument is required here and the old line of adding a key should be removed as well.

```suggestion
gittuf policy add-person -k ../keys/policy --person-ID 'developer' --public-key ../keys/developer.pub
```

docs/get-started.md
```diff
@@ -145,20 +145,19 @@ from TUF) must be signed by keys specified in the root of trust.
 gittuf trust add-policy-key -k ../keys/root --policy-key ../keys/policy.pub
 gittuf policy init -k ../keys/policy --policy-name targets
 ```
-Then, use the policy key to initialize a policy and add a rule protecting the
-`main` branch.
+Add trusted person to the policy file. Then, use the policy key to initialize a policy and add a rule protecting the
+`main` branch.
```
patzielinski: Please also wrap this to 80 cols.

```suggestion
Add the person to the policy file and then, use the person's key to initialize a policy and add a rule protecting the `main` branch. 
```

docs/get-started.md
```diff
@@ -145,20 +145,19 @@ from TUF) must be signed by keys specified in the root of trust.
...
+gittuf policy add-rule -k ../keys/policy --rule-name protect-main --rule-pattern git:refs/heads/main --authorize developer
 ```
 
 Note that `add-key` can also be used to specify a GPG key or a [Sigstore]
 identity for use with [gitsign]. However, we're using SSH keys throughout in
 this guide, as gittuf policy metadata currently cannot be signed using GPG (see
-[#229]). Also, `--authorize-key` in `gittuf policy add-rule` may return a
-deprecation warning. This guide will be updated with the new `--authorize` flag
-in its place.
+[#229]). If you were using `--authorize-key` and getting a deprecation warning, follow this updated guide to use `--authorize`.
```
patzielinski: I'd just remove the part about the deprecation warning.

sathas1301: just want to mention it, so in case if someone who already uses --authorize-key knows that it changed.

patzielinski: I figure we can just defer to addressing these cases if they arise since it was a rather old flag now.

docs/get-started.md
```diff
@@ -145,20 +145,19 @@ from TUF) must be signed by keys specified in the root of trust.
...
 gittuf policy init -k ../keys/policy --policy-name targets
 ```
-Then, use the policy key to initialize a policy and add a rule protecting the
-`main` branch.
+Add trusted person to the policy file. Then, use the policy key to initialize
+a policy and add a rule protecting the `main` branch.
 
 ```bash
+gittuf policy add-person -k ../keys/policy --person-ID 'developer' --public-key ../keys/developer.pub
 gittuf policy add-key -k ../keys/policy --public-key ../keys/developer.pub
```
patzielinski: Unsure if this will render properly but let's remove the `add-key` command right below the `add-person` command.
```suggestion
```

yongjae354 (approved):

docs/get-started.md
```diff
@@ -145,20 +145,18 @@ from TUF) must be signed by keys specified in the root of trust.
...
 ```
-Then, use the policy key to initialize a policy and add a rule protecting the
-`main` branch.
+Add trusted person to the policy file. Then, use the policy key to initialize
+a policy and add a rule protecting the `main` branch.
 
 ```bash
-gittuf policy add-key -k ../keys/policy --public-key ../keys/developer.pub
-gittuf policy add-rule -k ../keys/policy --rule-name protect-main --rule-pattern git:refs/heads/main --authorize-key ../keys/developer.pub
+gittuf policy add-person -k ../keys/policy --person-ID 'developer' --public-key ../keys/developer.pub
```
adityasaky:
```suggestion
gittuf policy add-person -k ../keys/policy --person-ID developer --public-key ../keys/developer.pub
```

for consistency with the tester-expected-*files.

docs/testing/tester-expected-win.txt
```diff
@@ -50,9 +50,8 @@ DEBUG:   11+  >>>> git config --local user.signingkey ../keys/developer
 DEBUG:   12+  >>>> gittuf trust init -k ../keys/root
 DEBUG:   13+  >>>> gittuf trust add-policy-key -k ../keys/root --policy-key ../keys/policy.pub
 DEBUG:   14+  >>>> gittuf policy init -k ../keys/policy --policy-name targets
-DEBUG:   15+  >>>> gittuf policy add-key -k ../keys/policy --public-key ../keys/developer.pub
-DEBUG:   16+  >>>> gittuf policy add-rule -k ../keys/policy --rule-name protect-main --rule-pattern git:refs/heads/main --authorize-key ../keys/developer.pub
-Flag --authorize-key has been deprecated, use --authorize instead
+DEBUG:   15+  >>>> gittuf policy add-person -k ../keys/policy --person-ID 'developer' --public-key ../keys/developer.pub
```
adityasaky: see other comment about consistency wrt `''` for developer.

docs/get-started.md
```diff
@@ -147,20 +147,18 @@ from TUF) must be signed by keys specified in the root of trust.
 gittuf trust add-policy-key -k ../keys/root --policy-key ../keys/policy.pub
 gittuf policy init -k ../keys/policy --policy-name targets
 ```
-Then, use the policy key to initialize a policy and add a rule protecting the
-`main` branch.
+Add trusted person to the policy file. Then, use the policy key to initialize
```
patzielinski:
```suggestion
Add a trusted person to the policy file. Then, use the policy key to initialize
```

docs/cli/gittuf_trust_update-hook.md
Yasho-Bapat, line 3: let's put GITTUF_DEV in a code cell (`GITTUF_DEV`)?

Yasho-Bapat, line 7: Please stick to 80 characters for line width.
```suggestion
Modify the parameters of an existing gittuf hook. Specify the name of the hook
to update and provide all parameters with their updated values. You can specify
multiple stages where the hook is defined. Note that all parameters required to
add the hook must also be provided. If a hook exists in multiple stages, only
the specified stage(s) will be updated. If a hook does not exist in the
specified stage, it won't be updated; you should add a new hook to that stage
instead. Currently, only the 'lua' environment is supported (developer mode
only, set `GITTUF_DEV=1`)
```

docs/development/gittuf-metadata.md
Yasho-Bapat, line 151: Can we have a URL embedded into the word "Attestations" that points to gittuf's documentation on what our attestations are? I think that sort of cross-referencing could be really helpful!

docs/get-started.md
Yasho-Bapat, line 61: @patzielinski do you think we can have some way of automating this? Doesn't make sense for us to update this manually with every release imo

docs/get-started.md
Yasho-Bapat: How about 
```suggestion
Add a trusted entity to the policy file. Then, use the policy key to initialize
```
to make it more all-encompassing?

experimental/gittuf/hook.go
Yasho-Bapat, line 204: curious to know what drove this rename? i might be missing some context.

experimental/gittuf/root.go
Yasho-Bapat, line 1089: nit: lets move this make call below the error check for `WriteBlob`
```suggestion
	blobID, err := r.r.WriteBlob(hookBytes)
	if err != nil {
		return err
	}
	var hashes = make(map[string]string, 2)
	hashes[gitinterface.GitBlobHashName] = blobID.String()
```

experimental/gittuf/root_test.go
Yasho-Bapat, line 1730: Let's add some tests to check some errors as well? Should be easy, like when no hook name is given or the timeout is incorrectly configured. Just to cover all bases!

patzielinski (approved):

PR #1112, opened 2025-10-07
ci: Add dependency review action

patzielinski:
This PR adds GitHub's dependency review action to scan for known vulnerable dependencies and check for licensing information, if available.

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,16 @@
+name: 'Dependency Review'
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@56339e523c0409420f6c2c9a2f4292bbb3c07dd3
+        with:
+            allow-licenses: 'Apache-2.0, MIT'


.github/workflows/dependency-review.yml
patzielinski, line 1: I know the Kusari bot is also on the repo, but I don't know if that does license checks hence my thought to add this action as well.

.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@56339e523c0409420f6c2c9a2f4292bbb3c07dd3
+        with:
+            allow-licenses: 'Apache-2.0, MIT'
```
patzielinski, line 16: Unsure if we want to specify what licenses are allowed and then expand this as we catch missing ones or handle it a different way.

.github/workflows/dependency-review.yml
adityasaky: cc @mlieberman85

.github/workflows/dependency-review.yml
adityasaky: What's the motivation here? Do we have a policy for licenses to allow? Apache-2.0 and MIT seem straightforward but we should likely support others too? I would also see if the OpenSSF has guidance on this.

.github/workflows/dependency-review.yml
mlieberman85: We do license checks as well.

adityasaky: If so, I suggest we just keep kusaribot. @mlieberman85 is it safe to assume that since the bot hasn't flagged anything it hasn't found any licenses we maybe don't want to pull in? Does the check only happen for _new_ dependencies or similar?

mlieberman85: So it happens for any change. If you add a new dependency we'd tell you if it uses a copyleft license or something. 

In the next few weeks we should have a risk posture check that would run on a regular cadence (e.g. nightly) that would also check over time. So if you have existing libraries that are of concern or if someone rug pulls, we'll highlight it.

PR #1114, opened 2025-10-08
docs: Fix some nits in the getting started doc

patzielinski:
Noticed that the getting started docs still showed that manual synchronization was required for gittuf references. I updated that and threw in a mention of our transport to hopefully get more folks to use it. Also fixed some formatting and version number bumps.

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -25,7 +25,7 @@ ARCH=amd64
 # One of: linux, darwin, freebsd
 OS=linux
 # See https://github.com/gittuf/gittuf/releases for the latest version
-VERSION=0.8.0
+VERSION=0.12.0
 cd $(mktemp -d)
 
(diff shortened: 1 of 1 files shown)


docs/get-started.md
```diff
@@ -198,9 +200,15 @@ gittuf verify-ref --verbose main
...
 
-gittuf includes helpers to push and fetch the policy and RSL references.
-However, there are some known issues (see [#328]) with these commands. In the
-meantime, Git can be used to keep gittuf's references updated.
+gittuf includes two main ways to push and fetch the policy and RSL references.
+You may use the `gittuf sync` command to synchronize changes with the remote
+automatically. You may also instead use the [gittuf
+transport](/internal/git-remote-gittuf), which handles the synchronization of
+gittuf metadata transparently upon standard Git pushes and pulls, without
+needing to explicitly invoke gittuf.
```
patzielinski, line 208: I can rearrange this if we feel that we should promote the transport as the recommended way for users to use gittuf.

cc @adityasaky

PR #1122, opened 2025-10-16
Fix git test when global config has format value

alanssitis:
The git config value cannot be set to an empty string, as it would not
overwrite if a value is already set.

### DIFF
--- a/internal/signerverifier/git/git_test.go
+++ b/internal/signerverifier/git/git_test.go
@@ -19,10 +19,7 @@ func TestLoadSignerFromGitConfig(t *testing.T) {
 		tmpDir := t.TempDir()
 		repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
 
-		// If the local machine has a format set, setting it with an
-		// empty string would not overwrite the existing configuration.
-		// Setting it to "gpg" is equivalent to leaving it unspecified.
-		if err := repo.SetGitConfig("gpg.format", "gpg"); err != nil {
+		if err := repo.SetGitConfig("gpg.format", ""); err != nil {
 			t.Fatal(err)
 		}
 		if err := repo.SetGitConfig("user.signingkey", ""); err != nil {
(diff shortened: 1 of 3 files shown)


patzielinski (commented):
I wonder if we should add an `UnsetGitConfig` function to do this?

cc @adityasaky

internal/signerverifier/git/git_test.go
```diff
@@ -19,7 +19,10 @@ func TestLoadSignerFromGitConfig(t *testing.T) {
...
 		repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
 
-		if err := repo.SetGitConfig("gpg.format", ""); err != nil {
+		// If the local machine has a format set, setting it with an
+		// empty string would not overwrite the existing configuration.
+		// Setting it to "gpg" is equivalent to leaving it unspecified.
+		if err := repo.SetGitConfig("gpg.format", "gpg"); err != nil {
 			t.Fatal(err)
 		}
 		if err := repo.SetGitConfig("user.signingkey", ""); err != nil {
```
adityasaky, line 25: do we have a similar issue with this one?

alanssitis: Interestingly, no, in my config this value is set.

Just tried it with CLI manually. Setting `gpg.format` should work when you pass an empty string, but it's not.

Will add tests that replicate.

alanssitis: Resolved in https://github.com/gittuf/gittuf/pull/1122/commits/0efb184bec42dd4ac96ba43b6cab25b71a55dcf7

I have attached the explanation of the issue in the commit comment.

alanssitis:
Was a config print + parsing issue. Resolved in https://github.com/gittuf/gittuf/pull/1122/commits/0efb184bec42dd4ac96ba43b6cab25b71a55dcf7 with the explanation of the issue in the commit message.

Issue #788, opened 2025-02-05
Add duplicate check for `PropagationDirective`s


patzielinski:
Related to #787. We should also add checks to the flows that add `PropagationDirectives` to ensure duplicates are not added to the metadata.

PR #1127, opened 2025-10-20
Add duplicate check for PropagationDirectives

dennisshey:
This PR implements a duplicate check for PropagationDirective objects added to repository metadata, addressing issue #788.

### DIFF
--- a/internal/tuf/v01/root.go
+++ b/internal/tuf/v01/root.go
@@ -440,7 +440,16 @@ func (r *RootMetadata) UpdateGlobalRule(globalRule tuf.GlobalRule) error {
 
 // AddPropagationDirective adds a propagation directive to the root metadata.
 func (r *RootMetadata) AddPropagationDirective(directive tuf.PropagationDirective) error {
-	// TODO: handle duplicates
+	for _, existing := range r.Propagations {
+		if existing.GetUpstreamRepository() == directive.GetUpstreamRepository() &&
+			existing.GetUpstreamReference() == directive.GetUpstreamReference() &&
+			existing.GetUpstreamPath() == directive.GetUpstreamPath() &&
+			existing.GetDownstreamReference() == directive.GetDownstreamReference() &&
+			existing.GetDownstreamPath() == directive.GetDownstreamPath() {
+			return tuf.ErrPropagationDirectiveAlreadyExists
+		}
+	}
+
 	r.Propagations = append(r.Propagations, directive)
 	return nil
 }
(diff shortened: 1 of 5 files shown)


internal/tuf/v01/root.go
```diff
@@ -440,7 +440,12 @@ func (r *RootMetadata) UpdateGlobalRule(globalRule tuf.GlobalRule) error {
 
 // AddPropagationDirective adds a propagation directive to the root metadata.
 func (r *RootMetadata) AddPropagationDirective(directive tuf.PropagationDirective) error {
-	// TODO: handle duplicates
+	for _, existing := range r.Propagations {
+		if existing.GetName() == directive.GetName() {
```
yongjae354: Maybe we should check `Name` and `UpstreamRepository` In case that 2 different upstream repos can have propagation directives with the same name? @patzielinski @adityasaky

internal/tuf/v01/root_test.go
```diff
@@ -86,6 +86,20 @@ func TestRootMetadata(t *testing.T) {
 		assert.Equal(t, 1, len(directives))
 		assert.Equal(t, directive, directives[0])
 
+		duplicate := &PropagationDirective{
+			Name: "test", 
+			UpstreamRepository: "https://example.com/git/other-repo",
+			UpstreamReference: "refs/heads/feature",
+			DownstreamReference: "refs/heads/main",
+			DownstreamPath: "feature/",
+   		}
```
yongjae354: I think we can probably use `directive` again here without creating a new one.

internal/tuf/v02/root_test.go
```diff
@@ -97,6 +97,20 @@ func TestRootMetadata(t *testing.T) {
...
 
+		duplicate := &PropagationDirective{
+			Name: "test", 
+			UpstreamRepository: "https://example.com/git/other-repo",
+			UpstreamReference: "refs/heads/feature",
+			DownstreamReference: "refs/heads/main",
+			DownstreamPath: "feature/",
+   		}
+
+		err = rootMetadata.AddPropagationDirective(duplicate)
```
yongjae354:
```suggestion
		err = rootMetadata.AddPropagationDirective(directive)
```

same here!

internal/tuf/v01/root.go
patzielinski: Looking at the test failures, I think we need to also check the other parts of a propagation directive. Let's see if every field is the same, and if so, error out in that case.

@dennisshey

PR #1129, opened 2025-10-22
transport: Enable verification of gittuf repo

patzielinski:
This enables gittuf verification on the transport.

WIP, still testing.

### DIFF
--- a/internal/git-remote-gittuf/main.go
+++ b/internal/git-remote-gittuf/main.go
@@ -109,7 +109,7 @@ func run(ctx context.Context) error {
 	remoteName := os.Args[1]
 	url := os.Args[2]
 
-	var handler func(context.Context, *gittuf.Repository, string, string) (map[string]string, bool, error)
+	var handler func(context.Context, *gittuf.Repository, string, string) (map[string]string, bool, []string, error)
 	switch {
 	case strings.HasPrefix(url, "https://"), strings.HasPrefix(url, "http://"), strings.HasPrefix(url, "ftp://"), strings.HasPrefix(url, "ftps://"):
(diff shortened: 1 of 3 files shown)


internal/git-remote-gittuf/main.go
```diff
@@ -175,28 +177,24 @@ func run(ctx context.Context) error {
...
-		// 	log(err.Error())
-		// 	if _, nerr := os.Stderr.Write([]byte("gittuf verification failed\n")); nerr != nil {
-		// 		return errors.Join(err, nerr)
-		// 	}
-		// 	return err
-		// }
+		rslTip := gittufRefsTips[rsl.Ref]
+		if rslTip != "" {
+			log("we have rsl tip")
+			err = repo.VerifyRefFromEntry(ctx, "HEAD", rslTip)
```
adityasaky: I think we can roll with this for a start but ideally we should be verifying every ref fetched in that interaction, not just HEAD. Also consider what happens in a bare repository.

internal/git-remote-gittuf/main.go
```diff
@@ -174,31 +177,18 @@ func run(ctx context.Context) error {
...
-		// 	}
-		// 	return err
-		// }
+		for _, ref := range fetchedRefs {
+			if !strings.HasPrefix(ref, "refs/gittuf/") {
+				log(fmt.Sprintf("Verifying %s...", ref))
+				err = repo.VerifyRef(ctx, ref)
+			}
+			if err != nil {
+				msg := fmt.Sprintf("Verification error for reference '%s': '%s'", ref, err.Error())
```
adityasaky, line 186: Can we simplify?

if has prefix { continue }

PR #1044, opened 2025-06-19
[WIP] *:  Add support for teams and hats

yongjae354:
This PR is a work in progress to address issue #231. We are adding support for teams, a group of multiple persons in gittuf metadata v02. For design documentation, see #1097.

PR #1130, opened 2025-10-23
tuf: Add v03 policy metadata with team principal

yongjae354:
This commit introduces v03 of the gittuf policy metadata schema, with a new `Team` Principal type.

### DIFF
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -21,6 +21,7 @@ import (
 	"github.com/gittuf/gittuf/internal/tuf/migrations"
 	tufv01 "github.com/gittuf/gittuf/internal/tuf/v01"
 	tufv02 "github.com/gittuf/gittuf/internal/tuf/v02"
+	tufv03 "github.com/gittuf/gittuf/internal/tuf/v03"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


internal/tuf/v03/tuf.go
```diff
@@ -0,0 +1,81 @@
...
+	return &k
+}
+
+// Person defines the structure for how a person identity is stored in policy
+// metadata. It implements tuf.Principal.
+type Person = v02.Person
+
+// Team defines the structure for how a team identity is stored in policy
+// metadata. It implements tuf.Principal.
+type Team struct {
```
patzielinski, line 41: Reminder that you need to declare a `Team` in the `tuf.go` file in the `tuf` package itself.

yongjae354: I was planning on defining `Team` in v03 similar to how `Person` is defined in v02, as there isn't an explicit `Person` definition in the `tuf` package?

It's a change I made when transporting from the prototype PR #1044 since the Team interface there seemed to be only useful when `Team` is defined in v02.

internal/policy/policy.go
```diff
@@ -1077,6 +1095,11 @@ func (s *State) GetTargetsMetadata(roleName string, migrate bool) (tuf.TargetsMe
 			return nil, fmt.Errorf("unable to unmarshal rule file metadata: %w", err)
 		}
 
+		if tufv03.AllowV03Metadata() && migrate {
```
patzielinski: I'd just simply roll this into the `if migrate` case below as an if statement, e.g.

```
if migrate {
<migrate to v02>
if allowv03 {
<migrate to v03>
}
}
```

internal/tuf/v03/root.go
```diff
@@ -0,0 +1,1066 @@
...
+
+	newKeyIDs := make([]tuf.Principal, 0, len(initialRootPrincipals))
+	for _, principal := range initialRootPrincipals {
+		switch p := principal.(type) {
+		case *Key:
+			newKeyIDs = append(newKeyIDs, p)
+		case *Person:
+			// don't need to be checked for duplicates skip
+		case *Team:
+			// don't need to be checked for duplicates skip
```
patzielinski, line 663: Flagging this as something that was introduced before; now is a good time to verify that this is intended behavior.

cc @adityasaky

patzielinski (commented):
I took a quick look through this. We're not wiring this up to any frontend code yet, so I don't think there should be much (if any) work left before merging?

cc @adityasaky for a review or thoughts on how to proceed.

patzielinski:
Closing this since there's a new prototype in the works. Please feel free to reopen this if you'd like to reuse this PR, otherwise we can open new PRs later.

PR #1138, opened 2025-11-11
feat(tui): support running in read-only mode when no signing key is provided

rithwik0908:
This PR adds support for running the TUI in read-only mode when no signing key (--signing-key or SIGSTORE_KEY) is provided.
This enhancement allows users to inspect and explore existing policy metadata without requiring a signing key, improving usability and accessibility for auditing and verification workflows.

### DIFF
--- a/internal/cmd/policy/tui/tui.go
+++ b/internal/cmd/policy/tui/tui.go
@@ -16,7 +16,6 @@ import (
 	"github.com/charmbracelet/lipgloss"
 	"github.com/gittuf/gittuf/experimental/gittuf"
 	trustpolicyopts "github.com/gittuf/gittuf/experimental/gittuf/options/trustpolicy"
-	"github.com/gittuf/gittuf/internal/cmd/common"
 	"github.com/gittuf/gittuf/internal/cmd/policy/persistent"
 	"github.com/gittuf/gittuf/internal/policy"
 	"github.com/gittuf/gittuf/internal/tuf"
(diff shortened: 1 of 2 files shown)


internal/cmd/policy/tui/tui.go
```diff
@@ -4,10 +4,10 @@
 package tui
 
 import (
-	"context"
-	"fmt"
-	"strconv"
-	"strings"
+    "context"
```
patzielinski: Can you run `gofmt` on your changes? Looks like some formatting was accidentally changed.

internal/cmd/policy/tui/tui.go
```diff
@@ -520,13 +542,20 @@ func (m *model) updateGlobalRuleList() {
...
-			m.mainList.View() + "\n" +
-				lipgloss.NewStyle().Foreground(lipgloss.Color(colorFooter)).Render(m.footer) +
-				"\nRun `gittuf policy apply` to apply staged changes to the selected policy file",
-		)
+    switch m.screen {
+    case screenMain:
+        // Show apply hint only when not in read-only mode.
+        hint := ""
+        if m.readOnly {
+            hint = "Read-only mode: no signing key provided"
```
patzielinski: One of the footer hints is causing a red line of text to appear. I'm not sure which one it is, but let's remove the one that causes it. We don't want to alarm the user that something is wrong when it's just the TUI running in read-only mode.

patzielinski (commented):
Please also run `make generate` and add the updated markdown files to the PR, thanks!

yongjae354 (commented):
I think internal/policy/1 and internal/policy/2 might have been added by accident?
Otherwise I think this looks good 👍

PR #1157, opened 2025-12-11
ci: Generate SBOMs for releases

patzielinski:
This helps us fulfill the baseline requirement [OSPS-QA-02.02](https://baseline.openssf.org/versions/2025-10-10#osps-qa-0202).


adityasaky:
May I suggest https://github.com/actions/attest-sbom and https://github.com/actions/attest-provenance?

PR #1158, opened 2025-12-11
ci: Add dry-run release run

### LINKED ISSUE #1128: Add dry run release on PRs
We should do a dry-run release on PRs. We would check:
a) things build as they should
b) might even be signing these artifacts and catching issues like this update perhaps?

_Originally posted by @adityasaky in https://github.com/gittuf/gittuf/issues/1126#issuecomment-3427851737_

patzielinski:
Closes #1128.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,9 +3,34 @@ on:
   push:
     tags:
       - 'v*'
+  pull_request:
 permissions: read-all
 jobs:
+  test-release:
+    if: github.event_name == 'pull_request'
+    permissions:
+      id-token: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8
+        with:
+          fetch-depth: 0 # fetch full history for previous tag information
+      - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c
+        with:
+          go-version: '1.25'
+          cache: true
+      - uses: sigstore/cosign-installer@faadad0cce49287aee09b3a48701e75088a2c6ad
+      - uses: goreleaser/goreleaser-action@e435ccd777264be153ace6237001ef4d979d3a7a
+        with:
+          distribution: goreleaser-pro
+          version: latest
+          args: release --snapshot --clean --skip=winget
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
+
   release:
+    if: github.event_name == 'push'
     permissions:
       contents: write
       id-token: write


.github/workflows/release.yml
```diff
@@ -3,9 +3,32 @@ on:
...
+      - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c
+        with:
+          go-version: '1.25'
+          cache: true
+      - uses: sigstore/cosign-installer@faadad0cce49287aee09b3a48701e75088a2c6ad
+      - uses: goreleaser/goreleaser-action@e435ccd777264be153ace6237001ef4d979d3a7a
+        with:
+          distribution: goreleaser-pro
+          version: latest
+          args: release --snapshot --clean
```
adityasaky: do we know how versioning works here? Do we need to do anything in the goreleaser config? https://goreleaser.com/customization/snapshots/

adityasaky (approved):
LGTM but a quick note.

.github/workflows/release.yml
adityasaky: Per logs, the versioning looks fine but we're hitting the LF prompt with signing the snapshot binaries. https://github.com/gittuf/gittuf/actions/runs/20146142984/job/57826963654?pr=1158#step:5:93

patzielinski: I'm not sure why the signing step requires a manual code, will debug but worst case I can configure it to just skip signing for snapshots.

adityasaky: We should be signing; we don't want to discover that signing isn't working when we try an actual release.

adityasaky: I think we filed this issue when there was a breaking change in the signing flow?

patzielinski: Fixed the issue. The `id-token` permission needed to be set to `write`.

.github/workflows/release.yml
```diff
@@ -3,9 +3,34 @@ on:
...
+      - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c
+        with:
+          go-version: '1.25'
+          cache: true
+      - uses: sigstore/cosign-installer@faadad0cce49287aee09b3a48701e75088a2c6ad
+      - uses: goreleaser/goreleaser-action@e435ccd777264be153ace6237001ef4d979d3a7a
+        with:
+          distribution: goreleaser-pro
+          version: latest
+          args: release --snapshot --clean --skip=winget
```
adityasaky, line 27: why is the --skip necessary if --snapshot doesn't actually upload any artifacts?

patzielinski: `--snapshot` implies `--skip=announce,publish,validate`, so it looks like winget is still invoked for some reason. 

https://goreleaser.com/cmd/goreleaser_release/

adityasaky: We should file an issue upstream about --snapshot skips being incomplete. :)

adityasaky (approved):

PR #1181, opened 2026-01-29
trust: add add-root-person command

rithwik0908:
Add a new gittuf trust add-root-person subcommand to register a Person principal in the root metadata.
-The command accepts --person-ID, one or more --public-key values, optional --associated-identity, and optional --custom metadata.
-Internally it constructs a tufv02.Person and calls Repository.AddRootKey so the person is authorized for the root role.
Testing
go test ./...

### DIFF
--- a/experimental/gittuf/root.go
+++ b/experimental/gittuf/root.go
@@ -202,6 +202,47 @@ func (r *Repository) AddRootKey(ctx context.Context, signer sslibdsse.SignerVeri
 	return r.updateRootMetadata(ctx, state, signer, rootMetadata, commitMessage, options.CreateRSLEntry, signCommit)
 }
 
+// AddRootPerson is the interface for the user to add an authorized person
+// for the Root role.
+func (r *Repository) AddRootPerson(ctx context.Context, signer sslibdsse.SignerVerifier, newRootPerson tuf.Principal, signCommit bool, opts ...trustpolicyopts.Option) error {
+	if signCommit {
+		slog.Debug("Checking if Git signing is configured...")
+		err := r.r.CanSign()
+		if err != nil {
+			return err
+		}
+	}
+
+	options := &trustpolicyopts.Options{}
+	for _, fn := range opts {
+		fn(options)
+	}
+
+	rootKeyID, err := signer.KeyID()
+	if err != nil {
+		return err
+	}
+
+	slog.Debug("Loading current policy...")
+	state, err := policy.LoadCurrentState(ctx, r.r, policy.PolicyStagingRef, policyopts.BypassRSL())
+	if err != nil {
+		return err
+	}
+
+	rootMetadata, err := r.loadRootMetadata(state, rootKeyID)
+	if err != nil {
+		return err
+	}
+
+	slog.Debug("Adding root person...")
+	if err := rootMetadata.AddRootPrincipal(newRootPerson); err != nil {
+		return err
+	}
+
+	commitMessage := fmt.Sprintf("Add root person '%s' to root", newRootPerson.ID())
+	return r.updateRootMetadata(ctx, state, signer, rootMetadata, commitMessage, options.CreateRSLEntry, signCommit)
+}
+
 // RemoveRootKey is the interface for the user to de-authorize a key
 // trusted to sign the Root role.
 func (r *Repository) RemoveRootKey(ctx context.Context, signer sslibdsse.SignerVerifier, keyID string, signCommit bool, opts ...trustpolicyopts.Option) error {
--- a/internal/cmd/trust/addrootperson/addrootperson.go
+++ b/internal/cmd/trust/addrootperson/addrootperson.go
@@ -1,7 +1,7 @@
 // Copyright The gittuf Authors
 // SPDX-License-Identifier: Apache-2.0
 
-package trust
+package addrootperson
 
 import (
 	"fmt"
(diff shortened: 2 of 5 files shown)


internal/cmd/trust/addrootperson.go
```diff
@@ -0,0 +1,126 @@
+// Copyright The gittuf Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package trust
```
yongjae354:
```suggestion
package addrootperson
```

internal/cmd/trust/trust.go
```diff
@@ -59,6 +59,7 @@ func New() *cobra.Command {
 	cmd.AddCommand(addpolicykey.New(o))
 	cmd.AddCommand(addpropagationdirective.New(o))
 	cmd.AddCommand(addrootkey.New(o))
+	cmd.AddCommand(newAddRootPersonCommand(o))
```
yongjae354:
```suggestion
	cmd.AddCommand(addrootperson.New(o))
```

internal/cmd/trust/addrootperson/addrootperson.go
yongjae354, line 1: We should put this file in it's own package like the other commands. 
The structure should be `internal/cmd/trust/addrootperson/addrootperson.go`

internal/cmd/trust/addrootperson.go
```diff
@@ -0,0 +1,126 @@
...
+
+	opts := []trustpolicyopts.Option{}
+	if o.p.WithRSLEntry {
+		opts = append(opts, trustpolicyopts.WithRSLEntry())
+	}
+
+	return repo.AddRootKey(cmd.Context(), signer, person, true, opts...)
+}
+
+func newAddRootPersonCommand(p *persistent.Options) *cobra.Command {
```
yongjae354:
```suggestion
func New(p *persistent.Options) *cobra.Command {
```

internal/cmd/trust/addrootperson.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/gittuf/gittuf/experimental/gittuf"
+	trustpolicyopts "github.com/gittuf/gittuf/experimental/gittuf/options/trustpolicy"
+	"github.com/gittuf/gittuf/internal/cmd/trust/persistent"
+	"github.com/gittuf/gittuf/internal/tuf"
+	tufv02 "github.com/gittuf/gittuf/internal/tuf/v02"
+	"github.com/spf13/cobra"
+)
+
+type addRootPersonOptions struct {
```
yongjae354:
```suggestion
type options struct {
```

Let's change this to match the style to other commands.

internal/cmd/trust/addrootperson.go
```diff
@@ -0,0 +1,126 @@
...
+		Custom:               custom,
+	}
+
+	opts := []trustpolicyopts.Option{}
+	if o.p.WithRSLEntry {
+		opts = append(opts, trustpolicyopts.WithRSLEntry())
+	}
+
+	return repo.AddRootKey(cmd.Context(), signer, person, true, opts...)
+}
```
yongjae354: The code here should look similar to the `Run()` function in [`addrootkey.go`](https://github.com/gittuf/gittuf/blob/main/internal/cmd/trust/addrootkey/addrootkey.go). We should instead place the core logic inside `experimental/gittuf` and internal functions (inside the `internal` directory).

For example, in `experimental/gittuf/root.go` You can create a function `AddRootPerson` that is similar to:

https://github.com/gittuf/gittuf/blob/a32a337ebc0a8a355ca85ccc4e91ad7614436f41/experimental/gittuf/root.go#L166-L203

that similarly calls the internal function `AddRootPrincipal`.
I think it would be appropriate to put the `person` initializing logic there as well.

yongjae354 (commented):
Hey @rithwik0908! Thank you for the PR. Looking good; I made a couple suggestions regarding the structure, as we want to have a similar structure across all the commands.

experimental/gittuf/root.go
```diff
@@ -202,6 +202,47 @@ func (r *Repository) AddRootKey(ctx context.Context, signer sslibdsse.SignerVeri
 	return r.updateRootMetadata(ctx, state, signer, rootMetadata, commitMessage, options.CreateRSLEntry, signCommit)
 }
 
+// AddRootPerson is the interface for the user to add an authorized person
+// for the Root role.
+func (r *Repository) AddRootPerson(ctx context.Context, signer sslibdsse.SignerVerifier, newRootPerson tuf.Principal, signCommit bool, opts ...trustpolicyopts.Option) error {
```
patzielinski, line 207: I think we can replace the `AddRootKey` function with this function. I suggest renaming this to `AddRootPrincipal` to be more precise, as a principal can encode a key or a person.

internal/cmd/trust/addrootperson/addrootperson.go
patzielinski, line 1: _Similar_ story here to the API function. We can deprecate the `add-root-key` function and replace it with this, but as with the policy command (https://github.com/gittuf/gittuf/blob/main/internal/cmd/policy/addperson/addperson.go), we should keep this `add-root-person`.

experimental/gittuf/root.go
patzielinski: Please also add tests to ensure that the function behaves properly with principals. See https://github.com/gittuf/gittuf/blob/main/experimental/gittuf/targets_test.go#L270.

Issue #1136, opened 2025-11-05
Discussion: Update gittuf Design Document to explain Principals


yongjae354:
With the changes introduced from [GAP-5](https://github.com/gittuf/gittuf/tree/main/docs/gaps/5), would it be worth modifying the gittuf Design document to explain that actors are defined with principals, not just keys?

Some relevant parts in the document are:

https://github.com/gittuf/gittuf/blob/9b802cc78c0dfc51997448cd65bea39e765be3b0/docs/design-document.md?plain=1#L80-L81

https://github.com/gittuf/gittuf/blob/9b802cc78c0dfc51997448cd65bea39e765be3b0/docs/design-document.md?plain=1#L221-L245

PR #1183, opened 2026-01-29
docs: Add notion of principals in `design-document.md`

yongjae354:
This PR is a small update to the gittuf design document, explaining the notion of "principals" introduced in GAP 5. 

Fixes: #1136

### DIFF
--- a/docs/design-document.md
+++ b/docs/design-document.md
@@ -1,6 +1,6 @@
 # gittuf Design Document
 
-Last Modified: May 29, 2025
+Last Modified: Mar 30, 2026
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


docs/design-document.md
```diff
@@ -213,13 +217,13 @@ as the primary rule file. Signatures are omitted.
 
 ```
 rootOfTrust:
-keys: {R1, R2, R3, P1, P2, P3}
+principals: {R1, R2, R3, P1, P2, P3}
```
patzielinski, line 220: Make sure to include the policy file's mapping of principals to their keys.

docs/design-document.md
```diff
@@ -598,8 +602,9 @@ entry is `P` and the namespace being checked is `N`. Then:
    checked.
 1. Begin traversing the delegations graph rooted at the primary rule file
    metadata.
-1. Verify the signatures of the primary rule file using the trusted keys in the
-   root of trust. If a threshold of signatures cannot be verified, abort.
+1. Verify the signatures of the primary rule file using the keys of the trusted
```
patzielinski, line 605: We should include something that specifies that a principal is only counted once even if they sign the same thing with more than one of their keys.

docs/design-document.md
```diff
@@ -78,12 +78,16 @@ Repository
 In a Git repository, an "actor" is any party, human or bot, who makes changes to
 the repository. These changes can involve any part of the repository, such as
 modifying files, branches or tags. In gittuf, each actor is identified by a
-unique signing key that they use to cryptographically sign their contributions.
-gittuf uses cryptographic signatures to authenticate actors as these signatures
-can be verified by anyone who has the corresponding public key, fundamental to
-gittuf's mission to enable independent verification of repository actions. Note
-that gittuf does not rely on Git commit metadata (e.g., author email, committer
-email) to identify the actor who created it, as that may be trivially spoofed.
+"principal". In the simplest form, a principal is just a single signing key.
```
patzielinski, line 81: I'd say that an actor is bound to or represented by a principal.

docs/design-document.md
yongjae354: Would this be similar to the schema in GAP5?
https://github.com/gittuf/gittuf/blob/a32a337ebc0a8a355ca85ccc4e91ad7614436f41/docs/gaps/5/README.md?plain=1#L47-L72

I'm not too sure how I would show this because we have 2 different implementations (key or person) of principals depending on our metadata version.

docs/design-document.md
```diff
@@ -598,8 +602,10 @@ entry is `P` and the namespace being checked is `N`. Then:
    checked.
 1. Begin traversing the delegations graph rooted at the primary rule file
    metadata.
-1. Verify the signatures of the primary rule file using the trusted keys in the
-   root of trust. If a threshold of signatures cannot be verified, abort.
+1. Verify the signatures of the primary rule file using the keys of the trusted
+   principals in the root of trust. During this process, a principal can be only
```
adityasaky:
```suggestion
   principals in the root of trust. During this process, a principal must only be
```

docs/design-document.md
yongjae354: Taking another look at this, I think GAP5 addresses this mapping and it might be redundant to include it here again. I think the explanation at Line 80-84 suffices; let me know if you think otherwise @patzielinski

patzielinski: Sure, we can defer to GAP 5.

PR #1188, opened 2026-02-03
tui: Fix logic for determining read-only mode

yongjae354:
This PR fixes the logic for entering read-only mode in the TUI.

Previously, the TUI would enter read-only mode if the `--signing-key` flag was not set by the user. Now it attempts to load the signing key from git config, before it falls back to read-only mode.

### DIFF
--- a/internal/cmd/tui/model.go
+++ b/internal/cmd/tui/model.go
@@ -4,6 +4,7 @@
 package tui
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


internal/cmd/tui/model.go
```diff
@@ -72,21 +73,21 @@ func initialModel(o *options) (model, error) {
...
 	var footer string
-	if !o.readOnly && o.p.SigningKey == "" {
-		footer = "No signing key provided, running in read-only mode."
-	}
-	readOnly := o.p.SigningKey == "" || o.readOnly
 
-	var signer dsse.SignerVerifier
 	if !readOnly {
-		// Load signer only if a signing key was provided.
+		// Try to load signer. Uses git config if signing key not explicitly provided
```
patzielinski:
```suggestion
		// Try to load signer. Uses Git config if signing key not explicitly provided
```

internal/cmd/tui/model.go
```diff
@@ -72,21 +73,21 @@ func initialModel(o *options) (model, error) {
...
 		signer, err = gittuf.LoadSigner(repo, o.p.SigningKey)
 		if err != nil {
-			// If a signing key was specified but cannot be loaded, return an empty model
-			// to preserve existing error behavior.
-			return model{}, err
+			if !errors.Is(err, gittuf.ErrSigningKeyNotSpecified) {
+				// If a signing key was found but cannot be loaded, return error
+				return model{}, fmt.Errorf("failed to load signing key from git config: %w", err)
```
patzielinski:
```suggestion
				return model{}, fmt.Errorf("failed to load signing key from Git config: %w", err)
```

internal/cmd/tui/model.go
```diff
@@ -72,21 +73,21 @@ func initialModel(o *options) (model, error) {
...
+			}
+			readOnly = true
+			footer = "No signing key found in git config, running in read-only mode."
```
patzielinski:
```suggestion
			footer = "No signing key found in Git config, running in read-only mode."
```

PR #1227, opened 2026-03-24
feat(tui): scaffold root principals support

PrinceAsiedu:
## Summary
- scaffold Trust menu support for root principals in TUI
- add root-principal listing read path wiring
- isolate root-principal actions as no-ops for now (step 4 follow-up)

## Notes
- root-principal mutations are intentionally deferred to a follow-up step
- this PR is opened as draft while implementation continues


docs/cli/gittuf_policy_tui.md
patzielinski: The Markdown files for commands can't be manually edited, they are generated from the data in the command definition (e.g. for the `tui` it's https://github.com/gittuf/gittuf/blob/main/internal/cmd/tui/tui.go#L49). Please make sure to remove the manual markdown edits, and, if you want to update the descriptions, edit the `.go` files.

docs/cli/gittuf_tui.md
patzielinski, line 1: Note that the Markdown files in the `docs/cli` folder are automatically generated, please do not manually edit them. To make changes to the markdown docs, instead edit the `.go` files where the commands are defined, e.g. https://github.com/gittuf/gittuf/blob/main/internal/cmd/tui/tui.go#L49

docs/cli/gittuf_policy_tui.md
patzielinski, line 1: This file is left over from a previous refactor, and will be removed in #1228 as a heads up.

experimental/gittuf/policy.go
```diff
@@ -106,6 +106,12 @@ type DelegationWithDepth struct {
 	Depth      int
 }
 
+type RootPrincipalWithRoles struct {
+	Principal  tuf.Principal
+	IsRoot	 bool
+	IsPrimaryRuleFileRole bool
+}
+
```
patzielinski: I suggest removing this from here to keep in line with how principals are processed in gittuf. You can instead perform two API calls to retrieve a set of principals trusted for root operations, and those trusted for toplevel/primary policy operations.

I think you can use such a struct internally in the TUI, but we want to keep the API uniform.

```suggestion
```

experimental/gittuf/policy.go
```diff
@@ -197,6 +203,59 @@ func (r *Repository) ListPrincipals(ctx context.Context, targetRef, policyName s
 	return metadata.GetPrincipals(), nil
 }
 
+// ListRootPrincipals returns all principals present in root metadata along with
+// role membership information relevant to root policy management.
+func (r *Repository) ListRootPrincipals(ctx context.Context, targetRef string) (map[string]RootPrincipalWithRoles, error) {
```
patzielinski: This function should instead be in `experimental/root.go`, as it deals with root of trust metadata. I also suggest splitting it into one for principals trusted for root operations and one for principals trusted for toplevel/primary policy operations.

experimental/gittuf/policy_test.go
```diff
@@ -326,3 +326,55 @@ func TestListPrincipals(t *testing.T) {
 		assert.Nil(t, principals)
 	})
 }
+
+func TestListRootPrincipals(t *testing.T) {
```
patzielinski: As with `ListRootPrincipals`, the tests added for it should be in `experimental/root_test.go`.

internal/cmd/tui/model.go
```diff
@@ -27,9 +27,12 @@ const (
 	screenPolicyAddRule                     // Form: add a new policy rule
 	screenPolicyEditRule                    // Form: edit selected rule (prefilled)
 	screenTrust                             // Menu for Trust operations
-	screenTrustGlobalRules                  // Global rule management screen
+	screenTrustGlobalRules
```
patzielinski: A comment was removed here?

internal/cmd/tui/trusthelpers.go
```diff
@@ -142,3 +150,97 @@ func repoUpdateGlobalRule(ctx context.Context, o *options, gr globalRule) error
...
+
+		roles := []string{}
+		if principalWithRoles.IsRoot {
+			roles = append(roles, "root")
+		}
+		if principalWithRoles.IsPrimaryRuleFileRole {
+			roles = append(roles, "policy")
+		}
+		if len(roles) == 0 {
+			roles = append(roles, "unscoped")
```
patzielinski: Principals that are not trusted for: a) root operations only, b) toplevel/primary policy file operations only, or c) both a & b must not be added to the root metadata. Make sure to have input validation or a handler that rejects non-conformant principals.

internal/cmd/tui/trusthelpers.go
```diff
@@ -142,3 +150,97 @@ func repoUpdateGlobalRule(ctx context.Context, o *options, gr globalRule) error
...
+		currPrincipals = append(currPrincipals, rootPrincipal{
+			principalID: id,
+			roles:       strings.Join(roles, ", "),
+			keyCount:    len(principalWithRoles.Principal.Keys()),
+		})
+	}
+	return currPrincipals
+}
+
+func repoAddRootPrincipal(ctx context.Context, o *options, role, principalSource string) error {
```
patzielinski, line 207: I wonder which would be better UX here for editing root of trust metadata:

Option A) Two screens in the TUI, one for editing principals trusted for root metadata operations and the other for editing principals trusted for toplevel/primary policy file operations, or,

Option B) One screen that combines the two.

gittuf's APIs currently correspond to Option A, while this PR appears to select Option B.

I personally lean towards Option A out of preference and how the APIs are currently written, but ccing @adityasaky for input.

internal/cmd/tui/model.go
```diff
@@ -260,3 +282,28 @@ func (m *model) updateGlobalRuleList() {
 	}
 	m.globalRuleList.SetItems(items)
 }
+
+func (m *model) initRootPrincipalInputs() {
+	m.inputs = initInputs([]inputField{
+		{"root or policy", "Role:"},
+		{"Path/gpg/fulcio principal (e.g. key.pub, gpg:<fingerprint>)", "Principal Source:"},
```
patzielinski, line 324: The description strings in the PR need to be clearer. For this one, I suggest something like the following:
```suggestion
		{"SSH key path / GPG fingerprint / Fulcio ID (e.g. key.pub, gpg:<fingerprint>)", "Principal Key:"},
```

patzielinski (changes requested):
Thanks for the PR, @PrinceAsiedu!

I'm reviewing this PR even though it's in draft stage to give some early feedback. I'll need to take a look at this at least one more time when the changes here can be compiled.

A few items in addition to my specific comments:

1. There are linting issues, make sure to run `gofmt` so that the linter and CI can run properly.
2. There's some important design decisions we need to think through with how to represent root/targets principals in the TUI so that it's clear to end-users how this works. (cc @adityasaky)
3. Please also note that in order for us to merge this PR, you will need to add a [DCO Signoff](https://wiki.linuxfoundation.org/dco) to *all commits* on the PR. You will therefore need to squash your existing commits to satisfy this requirement.

Overall, this is certainly a valuable addition to gittuf. Please feel free to reach out for clarification/if you have questions.

patzielinski:
Superseded by #1422 and the mentorship work in progress.

PR #1228, opened 2026-03-24
verify: Add support for specifying initial root keys for verification

patzielinski:
This adds support for specifying the expected root keys to `VerifyRef` and `VerifyRefFromEntry`, as is supported with `gittuf clone`.


docs/cli/gittuf_policy_tui.md
patzielinski, line 1: Noticed this file wasn't cleaned up when refactoring the TUI (`gittuf policy tui` -> `gittuf tui`), so bundling it here to remove it.

internal/cmd/verifyref/verifyref.go
```diff
@@ -41,6 +44,12 @@ func (o *options) AddFlags(cmd *cobra.Command) {
 		"",
 		"name of remote reference, if it differs from the local name",
 	)
+
+	cmd.Flags().Var(
+		&o.expectedRootKeys,
+		"root-key",
+		"set of initial root of trust keys for the repository (supported values: paths to SSH keys, GPG key fingerprints, Sigstore/Fulcio identities)",
+	)
```
adityasaky, line 52: do we expect a threshold of root keys to be provided?

adityasaky: if yes, we should clarify.

Also, we need to think about the UX a bit more, both here and in clone. If a user provides no root keys, verification may pass because initial keys are not verified at all. If the user provides a threshold of root keys, verification may pass. If they provide a subset of correct keys, verification may fail? This feels wonky and we may accidentally drive people towards dropping this in favor of TOFU. cc @JustinCappos for some thoughts on this.

patzielinski: > if yes, we should clarify.

Yes, I'll update the dosctring for this.

>  If a user provides no root keys, verification may pass because initial keys are not verified at all.

I figure this is the default/current behavior, no? We can add a flag to explicitly opt to use TOFU for the root keys.

I agree on the UX thoughts, we should discuss this.

PR #1232, opened 2026-03-27
test: add basic TUI tests using teatest

### LINKED ISSUE #1217: Add Testing to the TUI
The TUI (https://github.com/gittuf/gittuf/tree/main/internal/cmd/tui) does not currently have tests written for it. There is a testing framework for the TUI (I haven't looked at how usable/good it is, though): https://github.com/charmbracelet/x/tree/main/exp/teatest.

Let's add some tests to both the TUI (if possible) and the helper methods to ensure it all works as intended.

Aaravanand00:
Description

Adds basic tests for the TUI using the teatest framework.

Includes checks for initialization and quitting behavior, along with a unit test for helper logic. The scope is intentionally minimal to avoid fragile UI assertions and keep the tests stable.

Tests pass for the TUI package locally.

Fixes #1217

 AI Usage

Used AI assistance for guidance on structuring tests and refining the approach. All code and changes were manually reviewed and adjusted.

Contributor Checklist

  request.
  applicable.

### DIFF
--- a/internal/cmd/tui/tui_test.go
+++ b/internal/cmd/tui/tui_test.go
@@ -1,0 +1,117 @@
+// Copyright The gittuf Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package tui
+
+import (
+	"context"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/x/exp/teatest"
+	"github.com/gittuf/gittuf/pkg/gitinterface"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTUI(t *testing.T) {
+	tmpDir := t.TempDir()
+	currentDir, err := os.Getwd()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := os.Chdir(tmpDir); err != nil {
+		t.Fatal(err)
+	}
+	defer os.Chdir(currentDir) //nolint:errcheck
+
+	gitinterface.CreateTestGitRepository(t, tmpDir, false)
+
+	t.Run("Start and Quit", func(t *testing.T) {
+		o := &options{
+			readOnly: true,
+		}
+
+		m, err := initialModel(context.Background(), o)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		tm := teatest.NewTestModel(t, m, teatest.WithInitialTermSize(80, 24))
+		tm.Send(tea.KeyMsg{Type: tea.KeyRunes, Runes: []rune("q")})
+		tm.WaitFinished(t, teatest.WithFinalTimeout(time.Second*15))
+	})
+
+	t.Run("Policy UI Navigation", func(t *testing.T) {
+		o := &options{
+			readOnly: true,
+		}
+
+		m, err := initialModel(context.Background(), o)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		tm := teatest.NewTestModel(t, m, teatest.WithInitialTermSize(80, 24))
+
+		// Wait for main menu to render
+		teatest.WaitFor(t, tm.Output(), func(out []byte) bool {
+			return strings.Contains(string(out), "Policy")
+		}, teatest.WithCheckInterval(time.Millisecond*100), teatest.WithDuration(time.Second*15))
+
+		// Select "Policy" (already selected by default, so just press enter)
+		tm.Send(tea.KeyMsg{Type: tea.KeyEnter})
+
+		// Now we should be on the Policy Operations screen.
+		teatest.WaitFor(t, tm.Output(), func(out []byte) bool {
+			return strings.Contains(string(out), "gittuf Policy Operations")
+		}, teatest.WithCheckInterval(time.Millisecond*100), teatest.WithDuration(time.Second*15))
+
+		// Select "View Rules" (already selected by default)
+		tm.Send(tea.KeyMsg{Type: tea.KeyEnter})
+
+		// Now we should be on the Policy Rules screen.
+		// We check for "Policy Rules" title OR the "No items." placeholder for an empty list
+		teatest.WaitFor(t, tm.Output(), func(out []byte) bool {
+			content := string(out)
+			return strings.Contains(content, "Policy Rules") || strings.Contains(content, "No items.")
+		}, teatest.WithCheckInterval(time.Millisecond*100), teatest.WithDuration(time.Second*15))
+
+		// Send "q" to quit
+		tm.Send(tea.KeyMsg{Type: tea.KeyRunes, Runes: []rune("q")})
+		tm.WaitFinished(t, teatest.WithFinalTimeout(time.Second*15))
+	})
+}
+
+func TestSplitAndTrim(t *testing.T) {
+	tests := map[string]struct {
+		input    string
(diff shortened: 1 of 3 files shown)


patzielinski (commented):
Thanks for the PR, @Aaravanand00.

Can you elaborate on what happened with the tests you added in the first commit, but removed in the second commit? We want to have tests that walk through the TUI and actually test its functionality if possible.

Also, please reformat your PR description to match the [PR template](https://raw.githubusercontent.com/gittuf/gittuf/refs/heads/main/.github/pull_request_template.md).

Thanks!

Aaravanand00:
Hi @patzielinski 

I spent some time improving the TUI tests and making them more stable.

1.Earlier issue
The tests were sending input before the TUI was fully rendered. Since the updates are async, this was causing timing issues and making the tests flaky sometimes.

2.What I changed
I removed the timing based approach and switched to a deterministic pattern. Now the tests follow a simple flow:
wait for expected UI state, then send input, then wait again for the next state.

3.New test added
I added a new test TestTUIInteractionFlow which covers a basic user flow :-
start TUI, press Enter to navigate, press ESC to go back, and then quit using q.

4.Fix in existing test
I also updated TestTUINavigation to make it more robust. It now waits for the UI to be ready before sending input, and handles cases where the list might be empty.

5.Why this helps
This removes flakiness completely and makes the tests reliable. It also validates real user navigation instead of just checking isolated behavior. It also gives us a base to add more interaction-based tests going forward.

6.Local verification
I ran all TUI tests locally and everything is passing including the new test.

Screenshot 2026-04-07 162619

Aaravanand00:
@patzielinski Updated the PR to fix TUI test failures. I increased the teatest.WaitFor timeout from 5s to 15s to handle slower CI environments and fixed a compilation mismatch in tui_test.go related to the new initialModel signature. The tests should be stable now.

internal/cmd/tui/model.go
patzielinski, line 1: Why was the `Init` function moved?

internal/cmd/tui/tui_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/charmbracelet/lipgloss"
+	"github.com/charmbracelet/x/exp/teatest"
+	"github.com/gittuf/gittuf/pkg/gitinterface"
+	"github.com/muesli/termenv"
+	"github.com/stretchr/testify/assert"
+)
+
+func init() {
+	lipgloss.SetColorProfile(termenv.Ascii)
+}
```
patzielinski: This doesn't look to be used?

internal/cmd/tui/tui_test.go
```diff
@@ -0,0 +1,162 @@
...
+		{"a", []string{"a"}},
+		{" a ,b, c ", []string{"a", "b", "c"}},
+		{"", []string{""}},
+	}
+
+	for _, test := range tests {
+		result := splitAndTrim(test.input)
+		assert.Equal(t, test.expected, result)
+	}
+}
```
patzielinski: Can you move this to the bottom of the file?

internal/cmd/tui/tui_test.go
```diff
@@ -0,0 +1,162 @@
...
+	tm.Send(tea.KeyMsg{Type: tea.KeyRunes, Runes: []rune("q")})
+	tm.WaitFinished(t, teatest.WithFinalTimeout(time.Second*5))
+}
+
+func TestSplitAndTrim(t *testing.T) {
+	tests := []struct {
+		input    string
+		expected []string
+	}{
+		{"a, b, c", []string{"a", "b", "c"}},
```
patzielinski: Can you update this to follow the format we use for these? e.g. https://github.com/gittuf/gittuf/blob/e66eb9d0e593f7d5e1ed7c9bfab2324bd6843012/internal/display/display_test.go#L35

internal/cmd/tui/tui_test.go
```diff
@@ -0,0 +1,162 @@
...
+		{"", []string{""}},
+	}
+
+	for _, test := range tests {
+		result := splitAndTrim(test.input)
+		assert.Equal(t, test.expected, result)
+	}
+}
+
+func TestTUINavigation(t *testing.T) {
```
patzielinski:
```suggestion
func TestTUIPolicyUI(t *testing.T) {
```

internal/cmd/tui/tui_test.go
```diff
@@ -0,0 +1,162 @@
...
+		content := string(out)
+		return strings.Contains(content, "Policy Rules") || strings.Contains(content, "No items.")
+	}, teatest.WithCheckInterval(time.Millisecond*100), teatest.WithDuration(time.Second*5))
+
+	// Send "q" to quit
+	tm.Send(tea.KeyMsg{Type: tea.KeyRunes, Runes: []rune("q")})
+	tm.WaitFinished(t, teatest.WithFinalTimeout(time.Second*5))
+}
+
+func TestTUIInteractionFlow(t *testing.T) {
```
patzielinski: Let's omit this test for now.

internal/cmd/tui/model.go
Aaravanand00: Thanks for pointing this out  I moved the Init function while experimenting with making the test setup align better with the async initialization flow used in the TUI. 

The idea was to ensure the model was initialized in a way that reflects how the program actually starts so that the tests behave closer to real execution. While doing that I moved Init but looking at it again this change isn’t really required for the scope of this PR and also introduces unnecessary noise. I’ll revert this and keep the change minimal and focused purely on adding tests.

internal/cmd/tui/tui_test.go
Aaravanand00: yeah you're right this isn't strictly required for the tests. I initially added it to ensure consistent rendering across environments but it's not being used directly. I'll remove it to keep things clean.

internal/cmd/tui/tui_test.go
Aaravanand00: Sure I'll move this to the bottom of the file to match the existing structure.

internal/cmd/tui/tui_test.go
Aaravanand00: Makes sense this better reflects what the test is covering. I'll update the name.

internal/cmd/tui/tui_test.go
Aaravanand00: Understood. I’ll remove this test for now and we can revisit interaction-level testing once we have a more stable approach.

Aaravanand00:
Hi @patzielinski  thanks again for the detailed feedback. I’ve gone through all the comments and made the suggested changes :-

1. Reverted the unnecessary Init function change
2. Removed the unused init() from tests
3. Updated the test structure to use the map-based format
4. Renamed the test to TestTUIPolicyUI
5. Removed the interaction flow test for now
6. Moved TestSplitAndTrim to the bottom of the file

I first ran the tests locally on my Windows setup. The TUI-related tests were passing, but some other tests in the repository were failing due to environment related issues like ssh keygen, missing system commands like cat, and symlink permissions.

To verify this properly, I then ran the full test suite in a Linux environment (Ubuntu). I used:

go test ./...

In Ubuntu, all tests passed successfully, including the TUI tests and the previously failing ones. This confirmed that the failures I saw earlier were environment specific and not related to the changes in this PR.

patzielinski:
Superseded by #1312.

PR #1265, opened 2026-04-07
fix:honor gpg.program in LoadPublicKeyFromGitConfig

### LINKED ISSUE #1261: LoadPublicKeyFromGitConfig ignores gpg.program and always executes gpg from PATH
## Summary
`LoadPublicKeyFromGitConfig` does not honor `gpg.program` from Git config. It always shells out to `gpg` from `PATH`, while `LoadSignerFromGitConfig` correctly uses `gpg.program`.

## Why this is a real bug
- `keys.go` (line 53) hardcodes `exec.Command("gpg", "--export", "--armor", ...)`
- `keys.go` (line 123) routes GPG public-key loading through that path
- `keys.go` (line 209) and (line 227) show signer loading already respects `gpg.program`

This creates inconsistent behavior within the same configuration.

## Affected code
- `experimental/gittuf/keys.go`
  - `LoadPublicKey`: hardcoded `exec.Command("gpg", "--export", "--armor", ...)`
  - `LoadPublicKeyFromGitConfig`: GPG branch calls `LoadPublicKey("gpg:<fingerprint>")`
  - `LoadSignerFromGitConfig`: uses `getGPGOptions(config)` and honors `gpg.program`

## Reproduction
1. Configure repo:
   - `gpg.format = gpg`
   - `user.signingkey = <valid fingerprint>`
   - `gpg.program = /usr/bin/gpg` (or any valid absolute path)

2. Use an environment where:
   - `git` and `go` are available
   - `gpg` is **not present in PATH**

3. Call both:
   - `LoadSignerFromGitConfig` → succeeds
   - `LoadPublicKeyFromGitConfig` → fails with:
     ```
     exec: "gpg": executable file not found in $PATH
     ```

## Expected
Both signer loading and public-key loading should use the same GPG program resolution (i.e., respect `gpg.program` when configured).

## Actual
Public-key loading ignores `gpg.program` and depends on `gpg` being available in `PATH`.

## Impact
Breaks setups that intentionally use a minimal `PATH` and rely on explicit `gpg.program` configuration.

## Suggestion
Pass GPG program resolution through the public-key loading path as well, or introduce a dedicated `LoadPublicKeyFromGitConfig` branch that reuses the same GPG options logic as signer loading.

### Code of Conduct

ayuxsh009:
## Description
Fixes #1261.

LoadPublicKeyFromGitConfig previously ignored gpg.program and always executed gpg from PATH.  
LoadSignerFromGitConfig already honored gpg.program, which caused inconsistent behavior in minimal PATH environments.

This PR:
- Updates the GPG public-key loading path used by LoadPublicKeyFromGitConfig to respect gpg.program when configured.
- Keeps default behavior when gpg.program is not set.
- Adds a regression test that simulates restricted PATH and verifies both signer loading and public-key loading work when gpg.program is configured.

Testing performed:
- go test -count=1 -run TestLoadPublicKeyFromGitConfig ./experimental/gittuf -v
- go test -count=1 -race -run TestLoadPublicKeyFromGitConfig ./experimental/gittuf -v
- go test -count=1 ./experimental/gittuf -v
- go test -count=1 ./...

## AI Usage

- [x] I did not use generative AI at all in making the content of this pull request.
- [ ] I did use generative AI in some form in making the content of this pull request. I have described my use of AI below.

## Contributor Checklist

- [x] I have manually reviewed all content submitted to gittuf in this pull request.
- [x] All commits in this pull request include a DCO Signoff.
- [x] By submitting this pull request, I agree to follow the gittuf Code of Conduct.

### DIFF
--- a/experimental/gittuf/keys.go
+++ b/experimental/gittuf/keys.go
@@ -38,6 +38,8 @@ var (
 	ErrUnsupportedX509Method    = errors.New("unsupported X509 certificate specified in Git configuration")
 )
 
+const defaultGPGProgram = "gpg"
+
 // LoadPublicKey returns a signerverifier.SSLibKey object for a PGP / Sigstore
 // Fulcio / SSH (on-disk) key for use in gittuf metadata.
 func LoadPublicKey(keyRef string) (tuf.Principal, error) {
@@ -50,7 +52,15 @@ func LoadPublicKey(keyRef string) (tuf.Principal, error) {
 	case strings.HasPrefix(keyRef, GPGKeyPrefix):
 		fingerprint := strings.ToLower(strings.TrimSpace(strings.TrimPrefix(keyRef, GPGKeyPrefix)))
 
-		command := exec.Command("gpg", "--export", "--armor", fingerprint)
+		program := defaultGPGProgram
+		configCommand := exec.Command("git", "config", "--get", "gpg.program")
+		if stdOut, err := configCommand.Output(); err == nil {
+			if configuredProgram := strings.TrimSpace(string(stdOut)); configuredProgram != "" {
+				program = configuredProgram
+			}
+		}
+
+		command := exec.Command(program, "--export", "--armor", fingerprint) //nolint:gosec
 		stdOut, err := command.Output()
 		if err != nil {
 			return nil, err
--- a/experimental/gittuf/keys.go
+++ b/experimental/gittuf/keys.go
@@ -16,6 +16,7 @@ import (
 	sslibdsse "github.com/gittuf/gittuf/internal/third_party/go-securesystemslib/dsse"
 	"github.com/gittuf/gittuf/internal/tuf"
 	tufv01 "github.com/gittuf/gittuf/internal/tuf/v01"
+	"github.com/gittuf/gittuf/pkg/gitinterface"
 	"github.com/secure-systems-lab/go-securesystemslib/signerverifier"
 )
 
@@ -40,7 +41,7 @@ var (
 
 // LoadPublicKey returns a signerverifier.SSLibKey object for a PGP / Sigstore
 // Fulcio / SSH (on-disk) key for use in gittuf metadata.
-func LoadPublicKey(keyRef string) (tuf.Principal, error) {
+func LoadPublicKey(repo *Repository, keyRef string) (tuf.Principal, error) {
 	var (
 		keyObj *signerverifier.SSLibKey
 		err    error
@@ -50,7 +51,24 @@ func LoadPublicKey(keyRef string) (tuf.Principal, error) {
 	case strings.HasPrefix(keyRef, GPGKeyPrefix):
 		fingerprint := strings.ToLower(strings.TrimSpace(strings.TrimPrefix(keyRef, GPGKeyPrefix)))
 
-		command := exec.Command("gpg", "--export", "--armor", fingerprint)
+		program := gpg.DefaultGPGProgram
+		var config map[string]string
+		if repo != nil {
+			config, err = gitinterface.GetGitConfig(repo.r.GetGitDir())
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			config, err = gitinterface.GetGitConfig("")
+			if err != nil {
+				return nil, err
+			}
+		}
+		if value, has := config["gpg.program"]; has {
+			program = value
+		}
+
+		command := exec.Command(program, "--export", "--armor", fingerprint) //nolint:gosec
 		stdOut, err := command.Output()
 		if err != nil {
 			return nil, err
@@ -89,7 +107,7 @@ func LoadPublicKey(keyRef string) (tuf.Principal, error) {
 // LoadPublicKeyFromGitConfig loads a public key as with LoadPublicKey above,
 // but from the key specified in the Git configuration of the target repository.
 func LoadPublicKeyFromGitConfig(repo *Repository) (tuf.Principal, error) {
-	config, err := repo.r.GetGitConfig()
+	config, err := gitinterface.GetGitConfig(repo.r.GetGitDir())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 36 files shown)


experimental/gittuf/keys.go
```diff
@@ -119,8 +125,19 @@ func LoadPublicKeyFromGitConfig(repo *Repository) (tuf.Principal, error) {
...
+		if value, has := config["gpg.program"]; has {
+			program = value
+		}
+
+		keyObj, err := loadGPGPublicKey(fingerprint, program)
+		if err != nil {
+			return nil, err
+		}
+
+		return tufv01.NewKeyFromSSLibKey(keyObj), nil
```
patzielinski: This change is not necessary as this calls through to `LoadPublicKey`, which was fixed.

experimental/gittuf/keys.go
```diff
@@ -38,6 +38,18 @@ var (
...
+
+func loadGPGPublicKey(fingerprint, program string) (*signerverifier.SSLibKey, error) {
+	command := exec.Command(program, "--export", "--armor", fingerprint) //nolint:gosec
+	stdOut, err := command.Output()
+	if err != nil {
+		return nil, err
+	}
+
+	return gpg.LoadGPGKeyFromBytes(stdOut)
+}
```
patzielinski:
```suggestion
```

We can just make the fix in the `LoadPublicKey` function.

experimental/gittuf/keys.go
```diff
@@ -50,13 +62,7 @@ func LoadPublicKey(keyRef string) (tuf.Principal, error) {
 	case strings.HasPrefix(keyRef, GPGKeyPrefix):
 		fingerprint := strings.ToLower(strings.TrimSpace(strings.TrimPrefix(keyRef, GPGKeyPrefix)))
 
-		command := exec.Command("gpg", "--export", "--armor", fingerprint)
-		stdOut, err := command.Output()
-		if err != nil {
-			return nil, err
-		}
-
-		keyObj, err = gpg.LoadGPGKeyFromBytes(stdOut)
```
patzielinski, line 59: Please re-add the removed code and update it with the fix to query the gpg program from the Git config.

experimental/gittuf/keys_test.go
patzielinski, line 1: We need to add other tests for `LoadPublicKeyFromGitConfig` here anways, so I suggest just removing this file. We also need to write similar tests for the `gpg.program` bits in `signerverifier` so we can defer this until later.

patzielinski (changes requested):
Thanks for reporting the issue and sending this PR in, @ayuxsh009.

The solution to this is simpler than the code in the PR, we just need to have the GPG case in `LoadPublicKey` inspect the Git config. We also want to mirror the behavior of `signerverifier`, where the Git config is always inspected for the preferred GPG program.

Please make the changes I've suggested, squash commits when done, and we should get this merged soon. Thanks!

ayuxsh009:
> Thanks for reporting the issue and sending this PR in, @ayuxsh009.
> 
> The solution to this is simpler than the code in the PR, we just need to have the GPG case in `LoadPublicKey` inspect the Git config. We also want to mirror the behavior of `signerverifier`, where the Git config is always inspected for the preferred GPG program.
> 
> Please make the changes I've suggested, squash commits when done, and we should get this merged soon. Thanks!

Implemented. I moved the fix into LoadPublicKey so the GPG path now checks Git config for gpg.program (with fallback to gpg), matching signer behavior. I also squashed this PR to one commit.

experimental/gittuf/keys.go
ayuxsh009: Updated. I removed the extra GPG-specific logic from LoadPublicKeyFromGitConfig and routed it back through LoadPublicKey.

experimental/gittuf/keys.go
ayuxsh009: Done. The fix is now only in LoadPublicKey.

experimental/gittuf/keys.go
ayuxsh009: Done. The GPG export flow in LoadPublicKey is restored and now queries gpg.program from Git config before running export.

experimental/gittuf/keys_test.go
ayuxsh009: Done. I removed the added test changes from this PR so test expansion can be handled separately later.

patzielinski:
@adityasaky I've reworked this to now support using the user's chosen `gpg.program` even if a repository hasn't been loaded. It's not a super-elegant fix, so let me know if you have thoughts on this.

pkg/gitinterface/config.go
```diff
@@ -5,15 +5,30 @@ package gitinterface
...
+	var stdOut string
+	var err error
+
+	if repo != nil {
+		stdOut, err = repo.executor("config", "--get-regexp", `.*`).executeString()
+		if err != nil {
+			return nil, fmt.Errorf("unable to read Git config: %w", err)
+		}
+	} else {
+		output, err := exec.Command("config", "--global", "--get-regexp", `.*`).Output()
```
adityasaky: Missing `git`? Guessing this is not tested then? 😄

adityasaky: Also do we need `--global` here?

pkg/gitinterface/config.go
```diff
@@ -5,15 +5,30 @@ package gitinterface
...
+// If the supplied repository is nil, the system or global Git config will be
+// used as per
+// https://github.com/ayuxsh009/gittuf/tree/fix/gpg-program-publickey-path. The
+// "keys" for each config are normalized to lowercase.
+func GetGitConfig(repo *Repository) (map[string]string, error) {
+	var stdOut string
+	var err error
+
+	if repo != nil {
+		stdOut, err = repo.executor("config", "--get-regexp", `.*`).executeString()
```
adityasaky: given this is part of the now public gitinterface package, I wonder if this should be part of (r *Repository) still. If we're _anyway_ going to use exec.Command, that can happen in exp/gittuf which wraps this. We're not benefiting from the repo.executor pattern anyway.

PR #1266, opened 2026-04-07
fix: use --topo-order in GetCommitsBetweenRange instead of hash sort

### LINKED ISSUE #1254: `GetCommitsBetweenRange` sorts commits by hash string instead of topological order
### What happened?

I noticed that `GetCommitsBetweenRange` in `pkg/gitinterface/log.go` sorts commits by their hash string (lines 48-50) instead of preserving topological order. There's already a FIXME comment in the code that acknowledges this.

This function is called during policy verification (`internal/policy/verify.go`, lines 283, 982, 985), so commits could end up being checked out-of-order relative to when policy changes actually took effect. In most linear histories this probably goes unnoticed, but with branching and merging it could lead to incorrect verification results.

I'd expect commits to come back in topological order, parents before children which is how Git naturally understands history. `git rev-list` already supports `--topo-order`, so using that and dropping the hash-based sort should fix it.

I'd be happy to work on a fix for this, could you assign this to me?

Version: main branch (0ed0d9a)

### Relevant log output

```Shell

```

### Code of Conduct

1PoPTRoN:
## Description

`GetCommitsBetweenRange` was sorting commits by their hash string instead of preserving topological order. There was a FIXME comment in the code acknowledging this.

This PR:
- Adds `--topo-order` to the `git rev-list` call and removes the `sort.Slice` workaround
- Updates the doc comment to document the topological ordering contract
- Fixes tests that were sorting expected results by hash (which hid the issue), linear cases now assert exact topo-order with `assert.Equal`, branch/merge cases use `assert.ElementsMatch`

Fixes #1254

## AI Usage

- [ ] I **did not** use generative AI at all in making the content of this pull request.
- [x] I **did** use generative AI in some form in making the content of this pull request. I have described my use of AI below.

Used AI to help explore the codebase and suggest an initial approach. I wrote and reviewed the final code myself.

## Contributor Checklist

### DIFF
--- a/pkg/gitinterface/log.go
+++ b/pkg/gitinterface/log.go
@@ -5,18 +5,18 @@ package gitinterface
 
 import (
 	"fmt"
-	"sort"
 	"strings"
 )
 
 // GetCommitsBetweenRange returns the IDs of the commits that exist between the
-// specified new and old commit identifiers.
+// specified new and old commit identifiers. Commits are returned in
+// topological order, where each commit appears before its parents.
 func (r *Repository) GetCommitsBetweenRange(commitNewID, commitOldID Hash) ([]Hash, error) {
 	var args []string
 	if commitOldID.IsZero() {
-		args = []string{"rev-list", commitNewID.String()}
+		args = []string{"rev-list", "--topo-order", commitNewID.String()}
 	} else {
-		args = []string{"rev-list", fmt.Sprintf("%s..%s", commitOldID.String(), commitNewID.String())}
+		args = []string{"rev-list", "--topo-order", fmt.Sprintf("%s..%s", commitOldID.String(), commitNewID.String())}
 	}
 
 	commitRangeString, err := r.executor(args...).executeString()
@@ -37,17 +37,6 @@ func (r *Repository) GetCommitsBetweenRange(commitNewID, commitOldID Hash) ([]Ha
 		}
 		commitRange = append(commitRange, hash)
 	}
-
-	// FIXME: we should ideally be sorting this in the order of occurrence
-	// rather than by commit ID. The only reason this is happening is because
-	// the ordering of commitRange by default is not deterministic. Rather than
-	// walking through them and identifying the right order, we're sorting by
-	// commit ID. The intended use case of this function is to get a list of
-	// commits that are then checked for the changes they introduce. At that
-	// point, they must be diffed with their parent directly.
-	sort.Slice(commitRange, func(i, j int) bool {
-		return commitRange[i].String() < commitRange[j].String()
-	})
 
 	return commitRange, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/gitinterface/log_test.go
patzielinski, line 1: The changed tests in this file no longer check for ordering; isn't this something we should maintain when we have an order we can expect out of `GetCommitsBetweenRange`?

pkg/gitinterface/log.go
```diff
@@ -38,16 +38,5 @@ func (r *Repository) GetCommitsBetweenRange(commitNewID, commitOldID Hash) ([]Ha
 		commitRange = append(commitRange, hash)
 	}
 
-	// FIXME: we should ideally be sorting this in the order of occurrence
-	// rather than by commit ID. The only reason this is happening is because
-	// the ordering of commitRange by default is not deterministic. Rather than
-	// walking through them and identifying the right order, we're sorting by
-	// commit ID. The intended use case of this function is to get a list of
-	// commits that are then checked for the changes they introduce. At that
-	// point, they must be diffed with their parent directly.
```
patzielinski, line 47: @adityasaky is a walk needed here, or does `--topo-order` suffice?

patzielinski (changes requested):
Thanks for sending this in, @1PoPTRoN. Please see my comment on the tests.

pkg/gitinterface/log_test.go
1PoPTRoN: You are right that linear history tests should still verify the ordering. I was second guessing myself on whether maintaining those assertions was worth it but yes, `--topo-order` is what guarantees the order and for linear history its deterministic, so we should keep checking it.

I'll go ahead and restore `assert.Equal` for the linear cases. Should I also add a focused assertion for topological order in the merge cases or is `assert.ElementsMatch` sufficient since the order isnt deterministic there?

pkg/gitinterface/log_test.go
```diff
@@ -109,32 +105,23 @@ func TestGetCommitsBetweenRangeRepository(t *testing.T) {
 		// Check merge to first commit in main branch (not initial common commit)
 		expectedCommits := append([]Hash{mergeCommit}, mainBranchCommits[1:]...)
 		expectedCommits = append(expectedCommits, featureBranchCommits...)
-		sort.Slice(expectedCommits, func(i, j int) bool {
-			return expectedCommits[i].String() < expectedCommits[j].String()
-		})
 		commits, err := repo.GetCommitsBetweenRange(mergeCommit, mainBranchCommits[0])
 		assert.Nil(t, err)
-		assert.Equal(t, expectedCommits, commits)
+		assert.ElementsMatch(t, expectedCommits, commits)
```
adityasaky, line 110: Instead of ElementsMatch, can we prepare `expectedCommits` so it's already topologically sorted?

adityasaky: Once we do start sorting them topologically, I suspect someone somewhere will start relying on it even if we prefer they don't (this is now a public package). `ElementsMatch` checks all commits are included but it doesn't verify this implicit contract on ordering that someone may come to rely on.

PR #1273, opened 2026-04-09
fix(gitinterface): parse renamed entries in status output

### LINKED ISSUE #1272: pkg/gitinterface: Status() fails for renamed files (missing 'R' status parsing)
### Description
`Repository.Status()` fails when a file is renamed (`git mv`), because the status parser does not recognize the `'R'` porcelain status code.

### Steps to Reproduce
1. Create a repo and commit `old.txt`.
2. Rename it to `new.txt` using `git mv old.txt new.txt`.
3. Call `repo.Status()` from `pkg/gitinterface`.

### Actual Result
`repo.Status()` returns:
`unable to parse status code 'R' for path 'new.txt': status code string is unrecognized`

### Expected Result
`repo.Status()` should parse renamed-file status correctly and return a valid `FileStatus` map without error.

### Why this is a bug
- The status enum includes `StatusCodeRenamed`, but the parser (`NewStatusCodeFromByte`) does not handle `'R'`.
- This causes valid Git porcelain output to fail parsing.

### Impact
Any feature that depends on `Repository.Status()` can break on staged renames.

### Suggested Fix
- Add `'R'` handling in `NewStatusCodeFromByte`.
- Add a regression test in `pkg/gitinterface/status_test.go` for `git mv` rename cases.

### Code of Conduct

ayuxsh009:
## Description

Fixes #1272 a bug in `pkg/gitinterface.Repository.Status()` where renamed files (`git mv`) caused parsing failures.

### What changed
- Added support for `'R'` in `NewStatusCodeFromByte`.
- Updated porcelain `--porcelain=1 -z` parsing to correctly handle rename/copy records that include an extra NUL-separated source path token.
- Added regression test `TestStatusWithRenamedFile` in `pkg/gitinterface/status_test.go`.

### Verification
I manually tested and verified:
- `go test ./pkg/gitinterface -run TestStatusWithRenamedFile -v`
- `go test ./pkg/gitinterface`
- `go test ./internal/luasandbox`
- Manual rename reproduction (`git mv old.txt new.txt`) now returns parsed status without error.

## AI Usage

  request.

Used AI minimally for wording/help drafting.  
Code changes, review, and testing were done manually and verified by me.

## Contributor Checklist

  request.
  applicable.

### DIFF
--- a/pkg/gitinterface/status.go
+++ b/pkg/gitinterface/status.go
@@ -71,6 +71,8 @@ func NewStatusCodeFromByte(s byte) (StatusCode, error) {
 		return StatusCodeAdded, nil
 	case 'D':
 		return StatusCodeDeleted, nil
+	case 'R':
+		return StatusCodeRenamed, nil
 	case 'C':
 		return StatusCodeCopied, nil
 	case 'U':
@@ -114,22 +116,26 @@ func (r *Repository) Status() (map[string]FileStatus, error) {
 
 	statuses := map[string]FileStatus{}
 
-	lines := strings.Split(output, string('\000'))
-	for _, line := range lines {
-		if len(line) == 0 {
+	// `git status --porcelain=1 -z` emits NUL-separated tokens.
+	// For rename/copy records, the source path is emitted as an additional
+	// token after the main status token.
+	tokens := strings.Split(output, string('\000'))
+	for i := 0; i < len(tokens); i++ {
+		token := tokens[i]
+		if len(token) == 0 {
 			continue
 		}
 
 		// first two characters are status codes, find the corresponding
 		// statuses
-		xb := line[0]
-		yb := line[1]
+		xb := token[0]
+		yb := token[1]
 		// Note: we identify the status after inspecting the path so we can
 		// provide better error messages
 
 		// then, we have a single space followed by the path, ignore space and
 		// read in the rest as the filepath
-		filePath := strings.TrimSpace(line[2:])
+		filePath := strings.TrimSpace(token[2:])
 
 		xStatus, err := NewStatusCodeFromByte(xb)
 		if err != nil {
@@ -144,6 +150,16 @@ func (r *Repository) Status() (map[string]FileStatus, error) {
 		status := FileStatus{X: xStatus, Y: yStatus}
 
 		statuses[filePath] = status
+
+		// After splitting on NUL, rename/copy records have an additional token
+		// for the source path immediately after the main token.
+		if xStatus == StatusCodeRenamed || xStatus == StatusCodeCopied ||
+			yStatus == StatusCodeRenamed || yStatus == StatusCodeCopied {
+			if i+1 >= len(tokens) || len(tokens[i+1]) == 0 {
+				return nil, fmt.Errorf("unable to parse rename/copy status for path '%s': missing source path", filePath)
+			}
+			i++
+		}
 	}
 
 	return statuses, nil
--- a/pkg/gitinterface/status_test.go
+++ b/pkg/gitinterface/status_test.go
@@ -138,4 +138,48 @@ func TestStatus(t *testing.T) {
 	statuses, err = repo.Status()
 	assert.Nil(t, err)
 	assert.Empty(t, statuses)
+
+	// Rename one of the existing files
+	renamedPath := "baz"
+	if _, err := repo.executor("mv", filename, renamedPath).executeString(); err != nil {
+		t.Fatal(err)
+	}
+
+	statuses, err = repo.Status()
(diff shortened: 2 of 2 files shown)


pkg/gitinterface/status_test.go
```diff
@@ -139,3 +139,38 @@ func TestStatus(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Empty(t, statuses)
 }
+
+func TestStatusWithRenamedFile(t *testing.T) {
```
patzielinski: Please integrate this test into the `TestStatus` function above. You can likely just add it to the end and move one of the existing files.

patzielinski (changes requested):
cc @adityasaky for further review.

pkg/gitinterface/status_test.go
ayuxsh009: Made the changes as suggested - moved the rename case into TestStatus and removed the separate test. I reused an existing tracked file for the rename to keep it consistent with the rest of the test.

Re-ran `go test ./pkg/gitinterface -run TestStatus -count=1 -v` and `go test ./pkg/gitinterface -count=1`; both are passing.

pkg/gitinterface/status.go
```diff
@@ -115,10 +117,14 @@ func (r *Repository) Status() (map[string]FileStatus, error) {
...
 	lines := strings.Split(output, string('\000'))
-	for _, line := range lines {
+	for i := 0; i < len(lines); i++ {
+		line := lines[i]
 		if len(line) == 0 {
 			continue
 		}
+		if len(line) < 3 {
+			return nil, fmt.Errorf("unable to parse status entry '%s': malformed status output", line)
+		}
```
adityasaky, line 127: When can this happen?

pkg/gitinterface/status.go
```diff
@@ -144,6 +150,16 @@ func (r *Repository) Status() (map[string]FileStatus, error) {
...
+
+		// In porcelain v1 -z mode, rename/copy records include a second NUL
+		// terminated path token that points to the source path.
+		if xStatus == StatusCodeRenamed || xStatus == StatusCodeCopied ||
+			yStatus == StatusCodeRenamed || yStatus == StatusCodeCopied {
+			if i+1 >= len(lines) || len(lines[i+1]) == 0 {
+				return nil, fmt.Errorf("unable to parse rename/copy status for path '%s': missing source path", filePath)
+			}
+			i++
+		}
```
adityasaky, line 162: I'm a little confused by this, and it's possible I need to refresh my memory with this bit of gitinterface. Why are we looking at subsequent lines?

adityasaky (commented):
Thanks for filing this and apologies for the delay with my review.

pkg/gitinterface/status.go
ayuxsh009: This check was intended as a defensive safeguard against malformed or unexpected output. However, since git status --porcelain=1 -z guarantees the structure of each entry, I have removed the length check to keep the parsing logic simpler and aligned with the expected format.

pkg/gitinterface/status.go
ayuxsh009: I have clarified this in the code comments. In porcelain v1 -z mode, rename and copy entries are represented using two NUL-separated tokens - the first for the destination path and the second for the source path. Because of this, we need to look ahead and consume the next token as part of the same logical record.

ayuxsh009:
Hi @adityasaky i have made requested changes can you please review it ?
Thanks!

pkg/gitinterface/status_test.go
```diff
@@ -138,4 +138,16 @@ func TestStatus(t *testing.T) {
...
+	renamedPath := "baz"
+	if _, err := repo.executor("mv", filename, renamedPath).executeString(); err != nil {
+		t.Fatal(err)
+	}
+
+	statuses, err = repo.Status()
+	assert.Nil(t, err)
+	assert.Equal(t, map[string]FileStatus{
+		renamedPath: {X: StatusCodeRenamed, Y: StatusCodeUnmodified},
+	}, statuses)
```
adityasaky, line 152: Can you add a test case where the response has a renamed file and some other changes (e.g., added, modified, etc.)?

ayuxsh009: Added a test case that stages a rename (baz → quux) and an add (qux) simultaneously, then asserts both entries appear correctly in the Status() output. This verifies that the NUL-separated source-path token for the rename is properly consumed and doesn't break parsing of the following entry.

adityasaky (approved):

patzielinski (approved):

Issue #1289, opened 2026-04-21
No Error message while passing an invalid threshold to policy rules and global rules


IITI-tushar:
### What happened?

The Threshold should be a value, if the user types abc or leaves threshold field blank in TUI, then `thr, _ = strconv.Atoi(m.inputs[3].Value())` , here `strconv.Atoi` and `_` swallows the error, as a result `thr` sets to 0, and no error message is shown.
**Since the threshold is '0', the repo is unprotected.**

<img width="1311" height="256" alt="Image" src="https://github.com/user-attachments/assets/4e9b3b3e-8d16-4d26-86fd-9b9c8b0f44b8" />

<img width="1311" height="256" alt="Image" src="https://github.com/user-attachments/assets/ccdcb415-e849-4250-9213-18a81eeda57b" />

<img width="1604" height="776" alt="Image" src="https://github.com/user-attachments/assets/a282155b-95ce-4f3c-9693-32afc0eccc41" />

### Relevant log output

```Shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

patzielinski:
Hi @IITI-tushar, thanks for filing this.

I looked through the code and you are correct that the TUI will let the user set the threshold to 0 for a rule or global rule. However, the repo will not be unprotected as the verifier created to verify the rule will fail a sanity check here: https://github.com/gittuf/gittuf/blob/6dfe99adb488f9b3a9467dbe77c3590c7cd01dcc/internal/policy/signature.go#L57.

Regardless, I think this is a good time to add checking to the `tuf` package to prevent such nonsensical thresholds from getting written to the metadata, e.g. here: https://github.com/gittuf/gittuf/blob/6dfe99adb488f9b3a9467dbe77c3590c7cd01dcc/internal/tuf/v02/targets.go#L72

Do you want to take a pass at this? If not, I'll open a fix for this.

IITI-tushar:
@patzielinski Sure Sir, I would like to work on this!!!

patzielinski:
Closed by #1290.

PR #1290, opened 2026-04-22
reject invalid (zero or negative) thresholds at write time, and added error message(for invalid threshold) in TUI 

IITI-tushar:
Refers #1289 
## Description

This pull request introduces stricter validation for threshold values in both the TUI forms and the TUF metadata logic, ensuring that thresholds must be positive integers. It improves user feedback in the TUI and adds error handling to prevent invalid thresholds from being set at both the UI and data model levels.

**Threshold validation improvements:**

* Added checks in the TUI form submission handlers (`handlePolicyFormSubmit` and `handleGlobalFormSubmit` in `internal/cmd/tui/update.go`) to ensure that the threshold input is a valid number, displaying a user-friendly error message if not.
* Introduced a new error constant `ErrInvalidThreshold` in `internal/tuf/tuf.go` to represent invalid threshold values.

**TUF metadata enforcement:**

* Updated the `AddRule` and `UpdateRule` methods in both `internal/tuf/v01/targets.go` and `internal/tuf/v02/targets.go` to return an error if the threshold is not a positive integer. 
* Enhanced the `AddGlobalRule` and `UpdateGlobalRule` methods in both `internal/tuf/v01/root.go` and `internal/tuf/v02/root.go` to enforce that global rule thresholds must be positive integers, returning `ErrInvalidThreshold` otherwise.

## AI Usage

  request.
  
I used Copilot to **write description** for this Pull request :)

## Contributor Checklist

  request.
  applicable.

### DIFF
--- a/internal/cmd/tui/update.go
+++ b/internal/cmd/tui/update.go
@@ -267,7 +267,14 @@ func (m model) handlePolicyFormSubmit() (tea.Model, tea.Cmd) {
 		return m, nil
 	}
 
-	thr, _ := strconv.Atoi(m.inputs[3].Value())
+	var err error
+
+	parsedThr, err := strconv.ParseInt(m.inputs[3].Value(), 10, 0)
+	thr := int(parsedThr)
+	if err != nil {
+		m.errorMsg = "Invalid value specified; threshold must be a positive integer."
+		return m, nil
+	}
 	r := rule{
 		name:      m.inputs[0].Value(),
 		pattern:   m.inputs[1].Value(),
@@ -276,7 +283,6 @@ func (m model) handlePolicyFormSubmit() (tea.Model, tea.Cmd) {
 	}
 	authorizedKeys := splitAndTrim(m.inputs[2].Value())
 
-	var err error
 	switch m.screen {
 	case screenPolicyAddRule:
 		err = repoAddRule(m.ctx, m.options, r, authorizedKeys)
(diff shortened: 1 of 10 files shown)


internal/cmd/tui/update.go
```diff
@@ -267,7 +267,13 @@ func (m model) handlePolicyFormSubmit() (tea.Model, tea.Cmd) {
 		return m, nil
 	}
 
-	thr, _ := strconv.Atoi(m.inputs[3].Value())
+	var err error
+
+	thr, err := strconv.Atoi(m.inputs[3].Value())
```
patzielinski: Can we update the call to just use `ParseInt` instead of `Atoi`?

internal/cmd/tui/update.go
```diff
@@ -267,7 +267,13 @@ func (m model) handlePolicyFormSubmit() (tea.Model, tea.Cmd) {
...
+	if err != nil {
+		m.errorMsg = "Threshold must be a number or it will set to 0. Please correct it."
```
patzielinski:
```suggestion
		m.errorMsg = "Invalid value specified; threshold must be a positive integer."
```

internal/cmd/tui/update.go
```diff
@@ -308,8 +313,13 @@ func (m model) handleGlobalFormSubmit() (tea.Model, tea.Cmd) {
...
 	parts := splitAndTrim(m.inputs[2].Value())
 	thr := 0
+	var err error
 	if m.inputs[1].Value() == tuf.GlobalRuleThresholdType {
-		thr, _ = strconv.Atoi(m.inputs[3].Value())
+		thr, err = strconv.Atoi(m.inputs[3].Value())
+		if err != nil {
+			m.errorMsg = "Threshold must be a number or it will set to 0. Please correct it."
+			return m, nil
+		}
```
patzielinski: Same as above with `Atoi` and the error message.

patzielinski (changes requested):
Please also add tests to the respective `targets` and `root` files in the `tuf` package that check whether the threshold is sane. You can just add a line/two to the `TestGlobalRules` test that tries to add a threshold global rule with an invalid threshold, and similar for `TestAddRuleAndGetRules` (and their updater functions).

Let me know if you have questions, thanks!

internal/cmd/tui/update.go
IITI-tushar: I searched about using ParseInt instead of Atoi, go to know it is more formal approach, and is better if there is a massive value

internal/cmd/tui/update.go
IITI-tushar: done

patzielinski (commented):
Looks good, please squash commits down to one. Thanks!

IITI-tushar:
>Looks good, please squash commits down to one. Thanks!

@patzielinski Done

patzielinski (approved):

Issue #1259, opened 2026-04-06
Add Tests for Sigstore Functionality


patzielinski:
gittuf supports signing and validating data with [Sigstore](sigstore.dev), but given Sigstore's requirement to either sign using the public good instance of Rekor or use a private instance, we do not have unit tests for this functionality at the moment. We **do not** want to spam the public-good instance of Rekor in testing this functionality. https://github.com/sigstore/sigstore-go may have useful functionality for this.

We should add testing for this functionality. Examples of files where we need tests for this are https://coveralls.io/builds/78664614/source?filename=experimental%2Fgittuf%2Fkeys.go and https://coveralls.io/builds/78664614/source?filename=internal%2Fsignerverifier%2Fsigstore%2Fsigstore.go.

This is an offshoot of #463 and #73.

pushkarscripts:
Hi, I’d like to work on this issue.

patzielinski:
Sure, please let us know if you have questions.

pushkarscripts:
I’ve added tests around key loading (Fulcio parsing, GPG/SSH error cases) and basic Sigstore config handling in my recent PR.

I’d like to continue working on this and expand coverage for the remaining Sigstore functionality. Let me know if there are specific areas you’d prefer me to focus on.

patzielinski:
A decent amount of the code in question for Sigstore coverage is https://coveralls.io/builds/79013234/source?filename=internal%2Fsignerverifier%2Fsigstore%2Fsigstore.go. You might want to look into if it's possible to mock the Sigstore API (ideally) or if we can spin up a private instance so we don't pollute the public-good log with test calls.

pushkarscripts:
> A decent amount of the code in question for Sigstore coverage is https://coveralls.io/builds/79013234/source?filename=internal%2Fsignerverifier%2Fsigstore%2Fsigstore.go. You might want to look into if it's possible to mock the Sigstore API (ideally) or if we can spin up a private instance so we don't pollute the public-good log with test calls.

Got it, thanks!

I’ll take a look at this and explore mocking the Sigstore interactions.

PR #1299, opened 2026-04-24
tests: add unit tests for sigstore verifier and signer logic

pushkarscripts:
## Description

Related to #1259

Adds unit tests to improve coverage for Sigstore verifier and signer logic in `internal/signerverifier/sigstore`.

The tests cover:
- Verifier behavior (`KeyID`, `ExpectedExtensionKind`, `SetExtension`)
- Verifier construction via `NewVerifierFromIdentityAndIssuer` (default, with options, and edge cases)
- Signer behavior, including option propagation and `KeyID`
- Metadata key generation (`MetadataKey`)
- Basic workflow-level coverage for `Sign` and `Verify`

The tests now attempt to move closer to the actual signing and verification workflow. Since Sigstore depends on external services (e.g., Rekor/Fulcio), these tests currently validate expected failure behavior in this setup rather than full end-to-end verification.

## AI Usage

  request.

I used generative AI for general guidance while structuring tests and thinking about coverage. All code and tests were written, reviewed, and understood by me before submission.

## Contributor Checklist

  request.
  applicable.


patzielinski (commented):
I'm not sure if this meaningfully tests our Sigstore workflow. The main concern I have is that we don't test any kind of signing or verification, which is done for other methods, e.g. GPG: https://github.com/gittuf/gittuf/blob/main/internal/signerverifier/gpg/gpg_test.go.

This is what I was referring to in the issue about using Sigstore's package to possibly mock the API or just spin up a Rekor server here.

pushkarscripts:
I’ve updated the patch to move a bit closer to the signing and verification workflow by exercising `Sign` and `Verify`. Since Sigstore depends on external services, the current tests mostly cover the expected failure paths in this setup.

I’ll spend some time next looking into mocking Sigstore interactions or using a local Rekor instance so the full signing and verification flow can be tested more realistically.

pushkarscripts:
@patzielinski @adityasaky just checking in on this PR. Would appreciate any feedback. Thanks!

internal/signerverifier/sigstore/options/signer/signer.go
```diff
@@ -11,11 +17,15 @@ const (
...
-	ClientID    string
-	RedirectURL string
-	FulcioURL   string
-	RekorURL    string
+	IssuerURL     string
+	ClientID      string
+	RedirectURL   string
+	FulcioURL     string
+	RekorURL      string
+	IDTokenGetter oauthflow.TokenGetter
```
wlynch: I don't think we're setting this anywhere today - remove?

internal/signerverifier/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,220 @@
...
+		signer.rekor = &mockTransparency{}
+
+		_, err = signer.Sign(context.Background(), data)
+		// It will fail because our mocks return errors, but it confirms we hit the sign path
+		// and that identity/issuer were correctly parsed from the token.
+		assert.Error(t, err)
+
+		assert.Equal(t, identity, signer.identity)
+		assert.Equal(t, issuer, signer.issuer)
+	})
```
wlynch: This is mostly equivalent to 1 above (1 uses an in-memory implementation while this uses a mock). I don't think this gives use that much more, so we can omit this.

internal/signerverifier/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,220 @@
...
+	virtualSigstore, err := ca.NewVirtualSigstore()
+	require.NoError(t, err)
+
+	identity := "user@example.com"
+	issuer := "https://issuer.example.com"
+	data := []byte("test data")
+
+	// 1. Test Verification with externally signed artifact
+	t.Run("Verify externally signed artifact", func(t *testing.T) {
+		entity, err := virtualSigstore.Sign(identity, issuer, data)
```
wlynch, line 135: Instead of calling virtualSigstore.Sign, call gittuf's Signer.Sign. This should also fix the extension issue and let you actually be able to verify.

internal/signerverifier/sigstore/options/signer/signer.go
```diff
@@ -11,11 +17,15 @@ const (
...
+	Fulcio        sign.CertificateProvider
```
wlynch: Add documentation that this takes priority over FulcioURL (same with rekor)

pushkarscripts:
Thanks @wlynch for the review!

I've addressed the comments about the unused ID token injection support and added documentation clarifying that the injected Fulcio/Rekor implementations take precedence over the configured URLs. I also removed the redundant signer mock test.

I'm still working through the remaining feedback about using gittuf's `Signer.Sign()` instead of `virtualSigstore.Sign()` for the workflow test and will follow up with an update there.

internal/signerverifier/sigstore/sigstore_test.go
pushkarscripts: > Instead of calling virtualSigstore.Sign, call gittuf's Signer.Sign. This should also fix the extension issue and let you actually be able to verify.

I've been looking into this but got a bit stuck.

`VirtualSigstore` implements `root.TrustedMaterial`, but I don't see a way to use it directly with `Signer.Sign()` since it doesn't implement `sign.CertificateProvider` or `sign.Transparency`.

Am I missing something, or should I add test-only wrappers for those interfaces?

PR #1361, opened 2026-06-11
Attestations tests

R4R3D1FF:
## Description

Adds test to ensure that the 'if has' condition of getGitHubPullRequestReviewDetails is covered. Added more tests to ensure coverage of more lines.

```
func (r *Repository) getGitHubPullRequestReviewDetails(ctx context.Context, currentAttestations *attestations.Attestations, client *gogithub.Client, githubBaseURL, owner, repository string, pullRequestNumber int, reviewID int64, useGitHubAPI bool) (string, string, string, error) {
        indexPath, has, err := currentAttestations.GetGitHubPullRequestApprovalIndexPathForReviewID(githubBaseURL, reviewID)
        if err != nil {
                return "", "", "", err
        }
        if has {
                base, from, to := indexPathToComponents(indexPath)
                return base, from, to, nil
        }
```

## AI Usage

- [X] I **did not** use generative AI at all in making the content of this pull
  request.
- [ ] I **did** use generative AI in some form in making the content of this

## Contributor Checklist

- [X] I **have manually reviewed all content** submitted to gittuf in this pull
  request.
- [X] I fully understand the content I am submitting.
- [X] The changes introduced are documented and have tests included if
  applicable.
- [X] My changes do not infringe on copyright/trademarks/etc.
- [X] All commits in this pull request include a [DCO
- [X] By submitting this pull request, I agree to follow the gittuf [Code of

### DIFF
--- a/experimental/gittuf/attestations_test.go
+++ b/experimental/gittuf/attestations_test.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"net/url"
 	"os"
-	"os/exec"
 	"strings"
 	"testing"
 	"time"
@@ -1007,21 +1006,7 @@ func TestGetGitHubPullRequestReviewDetails(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		cmd := exec.Command(
-			"git",
-			"-C", testDir,
-			"commit-tree",
-			"-m", "Merge feature into main\n",
-			"-p", targetHeadCommitID.String(),
-			"-p", featureHeadCommitID.String(),
-			mergeTreeID.String(),
-		)
-
-		out, err := cmd.Output()
-		assert.Nil(t, err)
-
-		mergeCommitID, err := gitinterface.NewHash(strings.TrimSpace(string(out)))
-		assert.Nil(t, err)
+		mergeCommitID := r.CommitWithParents(t, mergeTreeID, []gitinterface.Hash{targetHeadCommitID, featureHeadCommitID}, "Merge feature into main", false)
 
 		err = r.SetReference(absTargetRef, mergeCommitID)
 		assert.Nil(t, err)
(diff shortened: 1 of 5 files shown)


yongjae354:
Is the PR description not updated? Seems like there are a lot of changes in this PR.

R4R3D1FF:
> Is the PR description not updated? Seems like there are a lot of changes in this PR.

Oh yes, I added coverage for other statements in the function as well, so its coverage bumped from 50% to 75%. I'll add it shortly.

R4R3D1FF:
> Is the PR description not updated? Seems like there are a lot of changes in this PR.

Wait, someghing's wrong. It should only show 1 file as changed but for dome

> Is the PR description not updated? Seems like there are a lot of changes in this PR.

Oh wait, something is wrong. I had only changed one file, and the commits from the previous PR show as new commits, perhaps because they were squashed in the upstream. I'll create a new branch from the upstream then.

I don't think it showed 36 files changed earlier.

R4R3D1FF:
> Is the PR description not updated? Seems like there are a lot of changes in this PR.

@yongjae354 I have pushed on top of the latest upstream again. Every test case I have added is for covering a specific block of code within the getGitHubPullRequestReviewDetails function, I can add these to the description and/or the comments above the testcases if you want.

experimental/gittuf/attestations_test.go
```diff
@@ -958,3 +959,409 @@ func TestGetGitHubClient(t *testing.T) {
...
+		if err := repo.RecordRSLEntryForReference(testCtx, featureRef, false, rslopts.WithRecordLocalOnly()); err != nil {
+			t.Fatal(err)
+		}
+
+		mergeTreeID, err := r.GetMergeTree(targetHeadCommitID, featureHeadCommitID)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		cmd := exec.Command(
```
patzielinski: Can we just export https://github.com/gittuf/gittuf/blob/main/pkg/gitinterface/commit.go#L132 and use that here instead of calling the git binary directly?

experimental/gittuf/attestations_test.go
```diff
@@ -958,3 +959,409 @@ func TestGetGitHubClient(t *testing.T) {
...
```
patzielinski: Same thing here with the git command.

experimental/gittuf/attestations_test.go
R4R3D1FF: That should work, I've pushed changes to accomplish that.

patzielinski (approved):
Looks good, I'll iterate on this in the coverage branch.

PR #1453, opened 2026-06-23
Fix/policy delete delegated rule files

leelearner:
## Description

Fixes a rule deletion bug caused by gittuf not providing a way to delete delegated policy files.

Previously, when a user removed a delegation rule with `gittuf policy remove-rule`, the corresponding delegated policy file could remain in policy metadata. During `gittuf policy apply`, this stale, unreachable delegated metadata caused `ErrDanglingDelegationMetadata`.

This PR:

1. Adds a new `gittuf policy delete-file` CLI command.
2. Deletes the corresponding delegated policy file by removing its entry from the current state's `DelegationEnvelopes` and then committing a newly written policy tree.
3. Updates rollback verification so that a missing delegated rule file is treated as valid only when the new policy graph no longer references that delegated role. If the new `targets` metadata can still reach a rule with that delegated role name, the missing file is still treated as rollback/corruption.

## AI Usage

  request.

I use AI to help analyze possible causes of the bug, think through potential solutions, and check whether my code covers all the cases it should consider.

## Contributor Checklist

  request.
- [ ] The changes introduced are documented and have tests included if
  applicable.

### DIFF
--- a/docs/cli/gittuf_policy.md
+++ b/docs/cli/gittuf_policy.md
@@ -31,6 +31,7 @@ The 'policy' command provides a suite of tools for managing gittuf policy config
 * [gittuf policy add-person](gittuf_policy_add-person.md)	 - Add a trusted person to a policy file
 * [gittuf policy add-rule](gittuf_policy_add-rule.md)	 - Add a new rule to a policy file
 * [gittuf policy apply](gittuf_policy_apply.md)	 - Validate and apply changes from policy-staging to policy
+* [gittuf policy delete-file](gittuf_policy_delete-file.md)	 - Delete a delegated policy file
 * [gittuf policy discard](gittuf_policy_discard.md)	 - Discard the currently staged changes to policy
 * [gittuf policy increment-version](gittuf_policy_increment-version.md)	 - Increment the integer version of the specified rule file metadata
 * [gittuf policy init](gittuf_policy_init.md)	 - Initialize policy file
@@ -45,4 +46,3 @@ The 'policy' command provides a suite of tools for managing gittuf policy config
 * [gittuf policy stage](gittuf_policy_stage.md)	 - Stage and push local policy-staging changes to remote repository
 * [gittuf policy update-person](gittuf_policy_update-person.md)	 - Update a person in a policy file
 * [gittuf policy update-rule](gittuf_policy_update-rule.md)	 - Update an existing rule in a policy file
-
--- a/experimental/gittuf/targets.go
+++ b/experimental/gittuf/targets.go
@@ -19,6 +19,8 @@ import (
 )
 
 var ErrInvalidPolicyName = errors.New("invalid rule or policy file name, cannot be 'root'")
+var ErrCannotDeletePrimaryPolicyFile = errors.New("cannot delete primary policy file")
+var ErrCannotDeleteReachablePolicyFile = errors.New("cannot delete reachable delegated policy file")
 
 // InitializeTargets is the interface for the user to create the specified
 // policy file.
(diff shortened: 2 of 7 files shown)


experimental/gittuf/targets.go
```diff
@@ -275,6 +277,46 @@ func (r *Repository) RemoveDelegation(ctx context.Context, signer sslibdsse.Sign
...
+	if err != nil {
+		return err
+	}
+	if referenced {
+		return ErrCannotDeleteReachablePolicyFile
+	}
+
+	delete(state.Metadata.DelegationEnvelopes, targetsRoleName)
+
+	commitMessage := fmt.Sprintf("Delete policy '%s'", targetsRoleName)
```
patzielinski:
```suggestion
	commitMessage := fmt.Sprintf("Delete policy file '%s'", targetsRoleName)
```

internal/policy/verify.go
```diff
@@ -796,6 +796,14 @@ func (s *StateMetadata) VerifyNewStateMetadata(_ context.Context, newStateMetada
 			// file, so if the new state doesn't have it, it's a rollback.
 			// This will change once we support deleting rule files.
```
patzielinski: This comment needs to be updated.

internal/policy/verify.go
patzielinski: Comment needs to be updated.

patzielinski (changes requested):
cc @adityasaky for review as well.

experimental/gittuf/targets.go
leelearner: Updated

internal/policy/verify.go
leelearner: Updated

internal/policy/verify.go
leelearner: Updated

docs/cli/gittuf_policy.md
```diff
@@ -45,4 +46,3 @@ The 'policy' command provides a suite of tools for managing gittuf policy config
 * [gittuf policy stage](gittuf_policy_stage.md)	 - Stage and push local policy-staging changes to remote repository
 * [gittuf policy update-person](gittuf_policy_update-person.md)	 - Update a person in a policy file
 * [gittuf policy update-rule](gittuf_policy_update-rule.md)	 - Update an existing rule in a policy file
-
```
adityasaky, line 48: this is strange -- is this auto generated?

internal/policy/policy.go
```diff
@@ -186,6 +186,45 @@ func (s *StateMetadata) GetTargetsMetadata(roleName string, migrate bool) (tuf.T
...
+		if err != nil {
+			return false, err
+		}
+		seen[rule.ID()] = true
+		queue = append(delegatedMetadata.GetRules(), queue...)
+	}
+
+	return false, nil
+}
+
```
adityasaky, line 227: The State already tracks `ruleNames`. Can we avoid walking the graph again? If we confirm `allRules` is populated based on rule entries and not existence of a delegated file, of course, but that's how I remember that working.

experimental/gittuf/targets.go
```diff
@@ -275,6 +277,46 @@ func (r *Repository) RemoveDelegation(ctx context.Context, signer sslibdsse.Sign
 	return r.updateTargetsMetadata(ctx, state, signer, targetsRoleName, targetsMetadata, commitMessage, options.CreateRSLEntry, signCommit)
 }
 
+func (r *Repository) DeletePolicyFile(ctx context.Context, targetsRoleName string, signCommit bool, opts ...trustpolicyopts.Option) error {
+	if targetsRoleName == policy.RootRoleName || targetsRoleName == policy.TargetsRoleName {
+		return ErrCannotDeletePrimaryPolicyFile
+	}
+
+	if signCommit {
```
adityasaky, line 285: The reason we didn't add this was it was unclear who should be able to delete a rule file. I notice we don't really do authentication here. cc @JustinCappos for thoughts on the mechanics of this.

Suppose targets had a rule for Alice, Bob, Carol (threshold 1) and they added a delegated rule file. Now if targets removes the rule, can any party remove the delegated rule file? Should it be one of Alice, Bob, or Carol? Should it be the targets signer? This can go on transitively because Alice/Bob/Carol may have delegated further that are now unreachable.

Something else to consider: what if the delegating role is compromised and deletes a rule? Should the attacker also be allowed to delete the rule file? I think this makes recovery harder, possibly, because you now need all the parties to come back and sign the metadata again.

adityasaky (changes requested):
I'm not sure this is the way to go with this. I've stated my specific concerns below but want to expand on the ultimate aim up here.

If the only problem of having dangling rule files is their mere existence, then I think the solution presented here isn't the right one. I would opt for having a threshold of authenticated root of trust holders be able to clean up dangling rule files as needed. From what I can tell, dangling rule files don't have any impact on verification though they contribute to sync overhead. And I can see rule name reuse to a different delegated set being an issue too, but that also strikes me as a case where an ultimate authority like the root of trust holders can intervene if needed. If this is a bigger problem in practice, we can revisit and identify more solutions, but we shouldn't start with as permissive a capability.

experimental/gittuf/targets.go
leelearner: Thanks for the thoughtful review! I focused too narrowly on implementing delegated policy file deletion and didn’t fully consider the authorization and recovery implications. I’ll step back and revisit the design, including whether deletion is the right way to address dangling metadata, and propose a safer solution.

patzielinski:
Closing since we'll need to think about the solution here a bit.

PR #1455, opened 2026-06-24
fix: record RSL entry in Discard to keep staging ref in sync`

### LINKED ISSUE #1342: RSL Entry is out of sync with PolicyStagingRef after doing a discard operation
### Steps to reproduce:
1. Execute any set of commands to make some changes to the PolicyStagingRef. Here's what I've executed.
`gittuf trust init -k C:\Users\R4R3D1FF\Documents\sources\gittuf\gittuf-get-started\keys\root`
`gittuf trust add-policy-key -k C:\Users\R4R3D1FF\Documents\sources\gittuf\gittuf-get-started\keys\root --policy-key C:\Users\R4R3D1FF\Documents\sources\gittuf\gittuf-get-started\keys\policy.pub
`
2. Stage all changes.
`gittuf policy stage --local-only`
3. Discard the staged changes.
`gittuf policy discard`
4. Attempt to apply changes.
`gittuf policy apply --local-only`
Expected outcome: There should be a message informing the user that PolicyStagingRef and PolicyRef are the same.
Actual outcome: There is a message indicating the ref is out of sync with the RSL.
Error: invalid policy state (is policy reference out of sync with corresponding RSL entry?)
Image

### Probable Cause:
The discard command doesn't make a new entry in the RSL to indicate the updated value of PolicyStagingRef.

### Version Info:
gittuf version 0.13.2-0.20260413125142-e66eb9d0e593+dirty

### Relevant log output

```Shell
Error: invalid policy state (is policy reference out of sync with corresponding RSL entry?)
```

### Code of Conduct

PratGit1606:
Fixes #1342

## Description

`policy.Discard` resets `PolicyStagingRef` to match `PolicyRef` but did not record a corresponding RSL entry after doing so. This broke the invariant that a ref's tip must equal its latest RSL entry target. When `policy apply` was subsequently called, `ReconcileStaging` compared the staging ref tip (now pointing at the policy commit) against the latest staging RSL entry target (still pointing at the discarded staged commit) and returned `ErrInvalidPolicy`.

This PR fixes the issue by appending an RSL reference entry for `PolicyStagingRef` after `SetReference` in `policy.Discard`. An early-return guard is also added to skip both the ref update and RSL write when staging already equals policy tip, avoiding redundant entries in the no-op case.

To thread signing behavior consistently with the rest of the policy workflow, `signCommit bool` is added to `Discard` and `DiscardPolicy`, with the CLI passing `true` to match `apply`.

A regression test is added to `experimental/gittuf/policy_test.go` that drives the real `StagePolicy → DiscardPolicy → Apply` flow. The existing `TestDiscard` did not cover this path because it creates the staging ref via a raw `repo.Commit` call, bypassing the RSL entirely, so no staging RSL entry existed to conflict with.

## AI Usage

- [x] I **did not** use generative AI at all in making the content of this pull
  request.

## Contributor Checklist

  request.
  applicable.

### DIFF
--- a/experimental/gittuf/policy_test.go
+++ b/experimental/gittuf/policy_test.go
@@ -213,7 +213,7 @@ func TestDiscardPolicy(t *testing.T) {
 		initialPolicyRef, err := repo.r.GetReference(policy.PolicyRef)
 		assert.Nil(t, err)
 
-		err = repo.DiscardPolicy()
+		err = repo.DiscardPolicy(false)
 		assert.Nil(t, err)
 
 		stagingRef, err := repo.r.GetReference(policy.PolicyStagingRef)
@@ -226,7 +226,7 @@ func TestDiscardPolicy(t *testing.T) {
 		r := gitinterface.CreateTestGitRepository(t, tmpDir, false)
 		repo := &Repository{r: r}
 
-		err := repo.DiscardPolicy()
+		err := repo.DiscardPolicy(false)
 		assert.Nil(t, err)
 
 		_, err = repo.r.GetReference(policy.PolicyStagingRef)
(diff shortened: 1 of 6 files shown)


experimental/gittuf/policy_test.go
```diff
@@ -243,13 +243,55 @@ func TestDiscardPolicy(t *testing.T) {
...
 
-		err = repo.DiscardPolicy()
+		err = repo.DiscardPolicy(false)
 		assert.Nil(t, err)
 
 		stagingRef, err := repo.r.GetReference(policy.PolicyStagingRef)
 		assert.Nil(t, err)
 		assert.Equal(t, initialRef, stagingRef)
 	})
+	// Regression test for issue #1342: Discard moved PolicyStagingRef without
```
patzielinski, line 254:
```suggestion

	// Regression test for issue #1342: Discard moved PolicyStagingRef without
```

experimental/gittuf/policy_test.go
```diff
@@ -243,13 +243,55 @@ func TestDiscardPolicy(t *testing.T) {
...
+	// recording an RSL entry, leaving the staging ref out of sync with its
+	// latest RSL entry and causing a subsequent Apply to fail with
+	// ErrInvalidPolicy. createTestRepositoryWithPolicy applies the policy, so we
+	// must advance the staging ref past the policy ref before discarding to
+	// exercise the bug path (otherwise Discard hits its no-op guard).
+	t.Run("discard after real stage flow then apply succeeds", func(t *testing.T) {
```
patzielinski, line 260:
```suggestion
	t.Run("discard after stage then apply succeeds", func(t *testing.T) {
```

patzielinski (commented):
Two nits on a quick inspection, need to look at this in more detail.

experimental/gittuf/policy_test.go
PratGit1606: Made the change

experimental/gittuf/policy_test.go
patzielinski: Bump on this (add a line above the comment).

patzielinski (commented):
Can you rebase this? I'm going to take a look at this again.

PR #1472, opened 2026-07-02
feat(sha256): support SHA-256 object format

pjbgf:
## Description

Make gittuf hash-algorithm aware so it operates on `SHA-256` repositories:

- gitinterface.Repository detects its object format (via `git rev-parse --show-object-format`) and exposes GetObjectFormat and a per-repo ZeroHash. GetReference returns the format-appropriate zero hash so ref updates to new refs succeed in SHA-256 repositories.
- Commit and tag signing/verification use the `gpgsig-sha256` header for SHA-256 objects (go-git's SignatureSHA256), matching Git's behavior. The verification field is chosen by the object's OID length. Fixes verify-ref failing on Git-signed SHA-256 RSL entries.
- git-remote-gittuf negotiates object-format from the repository instead of hardcoding sha1, and uses the repo-appropriate zero hash for ref tips.
- Test helpers can create SHA-256 repositories; gitinterface, rsl, policy, and experimental/gittuf suites now run under both formats, plus an end-to-end Git-signed record+verify test.

## AI Usage

  pull request. I have described my use of AI below:

AI was used to expand test coverage, to review/map gaps of SHA-256 support and to improve code documentation. All AI-assisted output was iteratively reviewed, refined, and validated through continuous developer feedback before being included in the PR.

## Contributor Checklist

  request.
  applicable.

### DIFF
--- a/experimental/gittuf/helpers_test.go
+++ b/experimental/gittuf/helpers_test.go
@@ -32,9 +32,13 @@ var (
 	ecdsaKeyBytes           = artifacts.SSHECDSAPrivate
 
 	testCtx = context.Background()
+
+	// testObjectFormats enumerates the Git object formats that behavioral test
+	// suites are run against.
+	testObjectFormats = []gitinterface.ObjectFormat{gitinterface.ObjectFormatSHA1, gitinterface.ObjectFormatSHA256}
 )
 
-func createTestRepositoryWithRoot(t *testing.T, location string) *Repository {
+func createTestRepositoryWithRoot(t *testing.T, location string, opts ...gitinterface.TestRepositoryOption) *Repository {
 	t.Helper()
 
 	signer := setupSSHKeysForSigning(t, rootKeyBytes, rootPubKeyBytes)
@@ -42,9 +46,9 @@ func createTestRepositoryWithRoot(t *testing.T, location string) *Repository {
 	var repo *gitinterface.Repository
 	if location == "" {
 		tempDir := t.TempDir()
-		repo = gitinterface.CreateTestGitRepository(t, tempDir, false)
+		repo = gitinterface.CreateTestGitRepository(t, tempDir, false, opts...)
 	} else {
-		repo = gitinterface.CreateTestGitRepository(t, location, false)
+		repo = gitinterface.CreateTestGitRepository(t, location, false, opts...)
 	}
 
 	r := &Repository{r: repo}
@@ -63,10 +67,10 @@ func createTestRepositoryWithRoot(t *testing.T, location string) *Repository {
 	return r
 }
 
-func createTestRepositoryWithPolicy(t *testing.T, location string) *Repository {
+func createTestRepositoryWithPolicy(t *testing.T, location string, opts ...gitinterface.TestRepositoryOption) *Repository {
 	t.Helper()
 
-	r := createTestRepositoryWithRoot(t, location)
+	r := createTestRepositoryWithRoot(t, location, opts...)
 
 	rootSigner := setupSSHKeysForSigning(t, rootKeyBytes, rootPubKeyBytes)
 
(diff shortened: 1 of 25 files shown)


pkg/gitinterface/tag_test.go
```diff
@@ -107,6 +107,99 @@ func TestRepositoryVerifyTag(t *testing.T) {
...
+			require.Nil(t, err)
+
+			tagID, err := repo.TagUsingSpecificKey(commitID, "v1", "v1\n", artifacts.SSHED25519Private)
+			require.Nil(t, err)
+
+			raw, err := repo.executor("cat-file", "tag", tagID.String()).executeString()
+			require.Nil(t, err)
+
+			// In SHA-256 repositories the signature must be stored under the
+			// `gpgsig-sha256` header, matching Git's own behavior.
```
pjbgf: Upstream x-ref:
- https://github.com/git/git/commit/42d4e1d1128fa1cb56032ac58f65ea3dd1296a9a
- https://github.com/git/git/commit/51acda73d3ca96b763f0fca3d7b33b4beaef786d

pkg/gitinterface/hash.go
```diff
@@ -47,10 +47,29 @@ func (h Hash) Equal(other Hash) bool {
...
+}
+
+// ZeroHash represents an empty SHA-1 Hash. It is safe to use as an
+// error-return sentinel and in comparisons via Hash.IsZero (which matches both
+// SHA-1 and SHA-256 zero hashes). When the value is passed to Git or compared
+// against a Git-produced ID, prefer Repository.ZeroHash, which returns the zero
+// hash matching the repository's object format.
 var ZeroHash = Hash(zeroSHA1HashBytes[:])
 
+// ZeroHash returns the all-zeroes Hash for the repository's object format. Git
```
patzielinski, line 63: I wonder if we could just simplify this to just have the func return the appropriate zero hash for the specified format of the repo?

pkg/gitinterface/signature.go
```diff
@@ -260,3 +261,41 @@ func getSigningKeyInfo(gitConfig map[string]string) string {
 	}
 	return keyInfo
 }
+
+// signatureFields returns the (Signature, SignatureSHA256) field values to set
+// on a go-git commit or tag signed for this repository. SHA-256 repositories
+// store the signature under the `gpgsig-sha256` header (SignatureSHA256),
+// matching Git's own behavior, so it can be read back during verification.
+func (r *Repository) signatureFields(signature string) (string, string) {
```
patzielinski: Could this be simplified to just one return string?

experimental/gittuf/helpers_test.go
```diff
@@ -32,19 +32,29 @@ var (
...
+
+	// testObjectFormats enumerates the Git object formats that behavioral test
+	// suites are run against.
+	testObjectFormats = []gitinterface.ObjectFormat{gitinterface.ObjectFormatSHA1, gitinterface.ObjectFormatSHA256}
 )
 
 func createTestRepositoryWithRoot(t *testing.T, location string) *Repository {
 	t.Helper()
 
+	return createTestRepositoryWithRootUsingObjectFormat(t, location, gitinterface.ObjectFormatSHA1)
```
patzielinski: We could update the helpers throughout gittuf to use our option format to create the repo in SHA-1 by default, but in SHA-256 if the caller requests (e.g. WithSHA256Format() as an option). Don't think we should block on that here, rather we can ticketize this.

internal/git-remote-gittuf/ssh.go
```diff
@@ -585,18 +586,18 @@ func handleSSH(ctx context.Context, repo *gittuf.Repository, remoteName, url str
 					// because of inconsistencies between receive-pack
 					// implementations in sending status messages.
 					// TODO: check that server advertises all of these
-					pushCmd = fmt.Sprintf("%s%s report-status-v2 atomic object-format=sha1 agent=git/%s", pushCmd, string('\x00'), gitVersion)
+					pushCmd = fmt.Sprintf("%s%s report-status-v2 atomic object-format=%s agent=git/%s", pushCmd, string('\x00'), objectFormat, gitVersion)
```
patzielinski, line 589: I assume that if there is a mismatch of SHA formats (e.g. a user pushes to the wrong repo) that will be caught here?

pkg/gitinterface/tag.go
```diff
@@ -111,36 +111,38 @@ func (r *Repository) verifyTagSignature(ctx context.Context, tagID Hash, key *si
...
+		return fmt.Errorf("unable to encode tag contents for verification: %w", err)
+	}
+
+	tagSignature := signatureForObjectID(tagID, tag.Signature, tag.SignatureSHA256)
+
 	switch key.KeyType {
 	case gpg.KeyType:
-		if _, err := tag.Verify(key.KeyVal.Public); err != nil {
+		if signatureBlockCount(tagSignature) > 1 {
+			return errors.Join(ErrIncorrectVerificationKey, ErrMultipleSignatures)
```
patzielinski: Is this check exclusive to GPG?

patzielinski (commented):
Looks good overall, I haven't had the time to sit down and read the appropriate docs on SHA-256 support in Git. cc @adityasaky in case you'd like to review as well.

internal/git-remote-gittuf/ssh.go
pjbgf: That's correct. This tells the server the object format of the push, if there is a mismatch with the server's view of the repository, the push will be rejected.

pkg/gitinterface/signature.go
pjbgf: The idea here was to simplify callers to avoid code duplication when assigning either `Signature` or `SignatureSHA256`. Given that there is just two occurrences (tag and commit), I'd drop this helper instead.

pkg/gitinterface/signature.go
```diff
@@ -260,3 +261,41 @@ func getSigningKeyInfo(gitConfig map[string]string) string {
...
+func signatureForObjectID(objectID Hash, signature, signatureSHA256 string) string {
+	if objectID.IsSHA256() {
+		return signatureSHA256
+	}
+	return signature
+}
+
+// signatureBlockCount reports how many armored signature blocks appear in a
+// signature. Verification rejects values carrying more than one block, which
+// are ambiguous.
```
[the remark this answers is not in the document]
pjbgf: Incoming fix on next commit.

pkg/gitinterface/tag.go
pjbgf: Good catch, this should be for all formats. It should be fixed now.

experimental/gittuf/helpers_test.go
pjbgf: +1. I added the variadic option for this now.

pkg/gitinterface/hash.go
pjbgf: If I understood it correctly, that would mean removing `gitinterface.ZeroHash` and just keep `Repository.ZeroHash()`. I think that's a good call, as most of places we are using that just as a placeholder value when returning errors.

I'll open this as a separate stacked PR, as that change alone touches over 40 files.

pkg/gitinterface/common.go
```diff
@@ -23,14 +32,41 @@ var (
...
+// format.
+func WithSHA256Format() TestRepositoryOption {
+	return WithObjectFormat(ObjectFormatSHA256)
+}
+
+// CreateTestGitRepository creates a Git repository in the specified directory,
+// using the SHA-1 object format unless overridden via options. This is meant
+// to be used by tests across gittuf packages. This helper also sets up an
+// ED25519 signing key that can be used to create reproducible commits.
+func CreateTestGitRepository(t *testing.T, dir string, bare bool, opts ...TestRepositoryOption) *Repository {
```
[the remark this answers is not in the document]
pjbgf: This will be fixed shortly.

patzielinski (approved):

PR #1480, opened 2026-07-07
Delegation metadata

moogchi:
## Description

During delegation removal. Removing the protect-main rule from targets updated the parent metadata, but it didn't remove the delegated rule file metadat (protect-main.json).
So during policy verification went through the reachable delegation graph, saw that protect-main.json was no longer referenced. So it  failed with unreachable targets metadata found. 

## Fix
gittuf/experimental/gittuf/targets.go
Prune unreachable delegated metadata when a delegation is removed

experimental/gittuf/targets_test.go
Regression coverage added.

## AI Usage
- [X] I **did** use generative AI in some form in making the content of this

Had the use ai to check if my pr doesn't have grammar or wording mistakes. I wrote the code with the help of ai. Then I verified the final code.


patzielinski:
Going to close this as this effort is already in #1453, and edits to that will likely take some design work. We can definitely reopen this depending on how the changes in that PR go.

PR #1490, opened 2026-07-11
test: add RunScript coverage

Sahil-13082003:
## Description

While looking through the coverage gaps for #463, I noticed that `RunScript` in `internal/luasandbox` was not directly covered by tests.

`RunScript` has three logical branches:
- the script runs and returns a number
- the script runs but returns a non-numeric value
- the script fails with a syntax error

This PR adds a test for each of those branches. Each subtest creates its own fresh `LuaEnvironment` to avoid any shared Lua state between test cases.

## AI Usage

I'm new to gittuf and Go, so I used generative AI to help me understand the codebase and learn Go testing patterns. I read the implementation of `RunScript` in `luasandbox.go` myself to identify the three uncovered branches, then manually reviewed, refined, and validated the tests before submitting this PR.

## Contributor Checklist

  request.
  applicable.

### DIFF
--- a/internal/luasandbox/luasandbox_test.go
+++ b/internal/luasandbox/luasandbox_test.go
@@ -12,6 +12,7 @@ import (
 	artifacts "github.com/gittuf/gittuf/internal/testartifacts"
 	"github.com/gittuf/gittuf/pkg/gitinterface"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	lua "github.com/yuin/gopher-lua"
 )
 
@@ -24,8 +25,9 @@ func TestNewLuaEnvironment(t *testing.T) {
 	repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
 
 	environment, err := NewLuaEnvironment(testCtx, repo)
+	require.NoError(t, err)
 	defer environment.Cleanup()
-	assert.Nil(t, err)
+
 	assert.NotNil(t, environment)
 }
 
@@ -34,9 +36,7 @@ func TestAPIMatchRegex(t *testing.T) {
 	repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
 
 	environment, err := NewLuaEnvironment(testCtx, repo)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 	defer environment.Cleanup()
 
 	t.Run("exact match", func(t *testing.T) {
@@ -46,7 +46,7 @@ func TestAPIMatchRegex(t *testing.T) {
 		`
 
 		err = environment.lState.DoString(testScript)
-		assert.Nil(t, err)
+		assert.NoError(t, err)
 
 		result := environment.lState.Get(-1)
 		environment.lState.Pop(1)
@@ -61,7 +61,7 @@ func TestAPIMatchRegex(t *testing.T) {
 		`
 
 		err = environment.lState.DoString(testScript)
-		assert.Nil(t, err)
+		assert.NoError(t, err)
 
 		result := environment.lState.Get(-1)
 		environment.lState.Pop(1)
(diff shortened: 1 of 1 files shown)


PratGit1606:
Hi Sahil, thanks for adding this!

I had a couple thoughts: The exit code assertions look right and the branches are all covered but one thing I wanted to flag though was that the parameters table gets pushed onto the stack before the script runs, but none of the test scripts actually touch it. So if something was broken in how parameters get passed to the script, these tests would still pass. Feels like it's testing the exit code interpretation more than RunScript itself.

Also each subtest creates its own fresh environment to avoid "shared Lua state" but RunScript balances its own stack so I don't think there's actually any bleed between calls. The rest of the file shares one environment across subtests, I recommend staying consistent with that.

Nits: t.TempDir() inlined instead of assigned to tmpDir first, and assert.NoError where the rest of the file uses assert.Nil (Not really a code breaking nit but more just so that we have everything consistent).
Thanks!

Sahil-13082003:
> Hi Sahil, thanks for adding this!
> 
> I had a couple thoughts: The exit code assertions look right and the branches are all covered but one thing I wanted to flag though was that the parameters table gets pushed onto the stack before the script runs, but none of the test scripts actually touch it. So if something was broken in how parameters get passed to the script, these tests would still pass. Feels like it's testing the exit code interpretation more than RunScript itself.
> 
> Also each subtest creates its own fresh environment to avoid "shared Lua state" but RunScript balances its own stack so I don't think there's actually any bleed between calls. The rest of the file shares one environment across subtests, I recommend staying consistent with that.
> 
> Nits: t.TempDir() inlined instead of assigned to tmpDir first, and assert.NoError where the rest of the file uses assert.Nil (Not really a code breaking nit but more just so that we have everything consistent). Thanks!

Thanks for the review! I made the style changes by switching to assert.Nil and using tmpDir before creating the test repository to match the rest of the file.

I also looked into the other points. I kept a separate LuaEnvironment for each subtest because RunScript returns early on the error path before removing the pushed parameters table, so I thought it was safer to keep each test separate.

For the parameters, I checked how DoString works in GopherLua. From what I understood, the pushed table doesn't seem to be available to the Lua script, so I kept these tests focused on the return value and error cases.

If I've misunderstood how the parameters are meant to work, or if you'd prefer a different approach, I'm happy to update the PR.

internal/luasandbox/luasandbox_test.go
```diff
@@ -656,3 +656,53 @@ func TestAPIGitGetRemoteURL(t *testing.T) {
...
+
+func TestRunScript(t *testing.T) {
+	t.Run("numeric return value", func(t *testing.T) {
+		tmpDir := t.TempDir()
+		repo := gitinterface.CreateTestGitRepository(t, tmpDir, false)
+
+		environment, err := NewLuaEnvironment(testCtx, repo)
+		if err != nil {
+			t.Fatal(err)
+		}
```
patzielinski: Let's have this (and other `t.Fatal` cases) as a one-liner that uses `require.NoError`. Please update the rest of the file.

internal/luasandbox/luasandbox_test.go
```diff
@@ -656,3 +656,53 @@ func TestAPIGitGetRemoteURL(t *testing.T) {
...
+		defer environment.Cleanup()
+
+		exitCode, err := environment.RunScript("return 0", lua.LTable{})
+
+		assert.Nil(t, err)
```
patzielinski:
```suggestion
		assert.NoError(t, err)
```

For the rest of the file as well please.

patzielinski (changes requested):
Two style nits as we're updating the style of our tests; this should be good to go after that. We can address any bits related to parameters in another PR.

Sahil-13082003:
> Two style nits as we're updating the style of our tests; this should be good to go after that. We can address any bits related to parameters in another PR.

Thanks Patrick! I updated the file to use require.NoError for the setup/error checks and assert.NoError for the assertions throughout the test file.

patzielinski (approved):
Thanks!

PR #1494, opened 2026-07-16
refactor: remove gitinterface.ZeroHash

pjbgf:
## Description

The package-level SHA-1 zero hash made it easy to leak a 40-zero ID
into SHA-256 repositories. This removes it to ensure that can no longer
happen.

Error paths now return nil, 'no object' markers are nil (Hash.IsZero
now treats nil/empty as zero), and anything that stringifies a zero
hash or hands it to Git must derive it from the repository via
`Repository.ZeroHash`.

## AI Usage

AI was used to expand test coverage, and doing the leg work across the touch points of the refactoring. All AI-assisted output was iteratively reviewed, refined, and validated through continuous developer feedback before being included in the PR.

## Contributor Checklist

  request.
  applicable.

### DIFF
--- a/experimental/gittuf/attestations_test.go
+++ b/experimental/gittuf/attestations_test.go
@@ -73,7 +73,7 @@ func TestApplyAttestations(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	env, err := allAttestations.GetReferenceAuthorizationFor(repo.r, targetTagRef, gitinterface.ZeroHash.String(), initialCommitID.String())
+	env, err := allAttestations.GetReferenceAuthorizationFor(repo.r, targetTagRef, repo.r.ZeroHash().String(), initialCommitID.String())
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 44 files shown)


PratGit1606:
Hi Paulo,
Thanks for putting this together! Personally, the core refactor looks solid. The ZeroHash replacement does seem to work. Since the updated files hardcoded with 40-zero strings are in SHA-1 only contexts, all the ID entry reads go through the right checks so I didn't see any panic risks there either.

I did have one finding though: `internal/rsl/cache_test.go` has `var testZeroHash = gitinterface.Hash(make([]byte, 20))` which is a hardcoded 20-byte SHA-1 zero. It's only used in in-memory tests right now so nothing is broken today, but `internal/rsl` has SHA-256 parametrized tests elsewhere in the package. I tried out a test locally by using `testZeroHash` in a SHA-256 repo context and `Commit` returned nil with no error but the RSL entry came back with a 40-char zero instead of 64, showing signs of silent data corruption, but it got masked by `IsZero()` so none of the existing guards would catch it.

I didn't find any nits personally, I'll try to dig into the macOS and Windows CI failures and see what went wrong but this is my understanding so far. Thanks! Lmk what you think

pjbgf:
@PratGit1606 thank you for reviewing this, I agree with your point and actually found a specific gap where that could be problematic. I have fixed that on #1498 which now supersedes this PR.

PR #1498, opened 2026-07-22
gitstore: introduce a shared storage interface, make rsl and verification dependency-light

pjbgf:
## Description

This restructures gittuf's storage access so verification can run over backends other than the git binary (e.g. `go-git`).

- `pkg/gitstore`: a new `Storer` interface which aligns with existing `*gitinterface.Repository`.
- `pkg/rsl` (from `internal/rsl`): dependency-light, requiring no `gitinterface/sigstore` imports, removing the need for the storer adapter.
- Signature verification moves to `internal/signerverifier/gitobject`, operating on bytes. `gitinterface` drops `sigstore/cosign/gitsign` from its dependency tree.
- attestations, cache, and policy consume the interface.

Breaking (`pkg/gitinterface`): `Repository.VerifySignature` and its
sentinels are removed. `rsl.Hash` is now a type alias of `githash.Hash`.

Points of note for reviewers:
- Policy reads git config once per verification, but now errors fail closed.
- Most cases lean on object-aware representations of "Zero Hash". In other cases, `nil` is being used instead.
- Fixes approval-attestation lookups on SHA-256 repositories, which used a SHA-1-width zero `fromID`. Which was one of the concerns [raised](https://github.com/gittuf/gittuf/pull/1494#issuecomment-5030248799) by @PratGit1606.

## AI Usage

AI was used to expand test coverage, and doing the leg work across the touch points of the refactoring. All AI-assisted output was iteratively reviewed, refined, and validated through continuous developer feedback before being included in the PR.

## Contributor Checklist

  request.
  applicable.


pkg/gitstore/gitstore.go
```diff
@@ -0,0 +1,112 @@
...
+
+	// GetCommitMessage returns the specified commit's message.
+	GetCommitMessage(commitID githash.Hash) (string, error)
+
+	// GetCommitParentIDs returns the IDs of the specified commit's parents.
+	GetCommitParentIDs(commitID githash.Hash) ([]githash.Hash, error)
+
+	// GetCommitsBetweenRange returns the commits reachable from commitNewID
+	// but not from commitOldID.
+	GetCommitsBetweenRange(commitNewID, commitOldID githash.Hash) ([]githash.Hash, error)
```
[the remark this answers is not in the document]
pjbgf: Resolved as per last push.

pkg/gitinterface/tree.go
```diff
@@ -324,6 +324,23 @@ func (r *Repository) CreateSubtreeFromUpstreamRepository(upstream *Repository, u
...
+	entries := make([]TreeEntry, 0, len(blobs)+len(subtrees))
+	for path, blobID := range blobs {
+		entries = append(entries, NewEntryBlob(path, blobID))
+	}
+	for path, treeID := range subtrees {
+		entries = append(entries, NewEntryTree(path, treeID))
+	}
+
+	return NewTreeBuilder(r).WriteTreeFromEntries(entries)
+}
```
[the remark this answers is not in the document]
pjbgf: As per [official docs](https://git-scm.com/docs/git-mktree#_description), this is a red herring when `git mktree` is being used:
> The order of the tree entries is normalized by mktree so pre-sorting the input is not required

However, I added some tests to pin the behaviour in case we are to ever change the Tree-creation back-end.

PratGit1606:
Hi Paulo,
Thanks a lot for putting this together. I like the idea behind the core architectural change! Firstly, the [fix ](https://github.com/gittuf/gittuf/pull/1498/changes#diff-db7db5dff0f25e4ab170f0233765cea49eda7172353c07f8bb8995349b4609a4R287)for the SHA-1 zero hash width issue from #1494 looks good here and no need to reopen that PR, having it bundled in this PR is fine.

I found some tiny things to flag here:

1. The list for `signatureBlockCount` in `gitobject.go` counts the PGP Signature, Message and SSH Signature to catch doubled signatures but is missing gitsign's header which is is `-----BEGIN SIGNED MESSAGE-----`. Because of this, a doubled Sigstore signature returns a count of 0 so the `> 1` guard silently doesn't fire and `ErrMultipleSignatures` never gets returned. I think you would need to add that to the list as well. 

2. On test coverage, the Sigstore/gitsign verify path in `gitobject.go` currently doesnt have test coverage. The `overrideStorer` in `attestations_test.go` is also missing blob error injection for the `codeReviewApprovalIndex` branches, and `pkg/rsl/storer_error_test.go` doesn't cover `CommitWithoutNumber`, `PropagationEntry.Commit`, or a terminal `storer.Commit` failure.

Other than that, everything looks good to me so far!

pkg/rsl/hash.go
```diff
@@ -0,0 +1,27 @@
...
+package rsl
+
+import (
+	"github.com/gittuf/gittuf/pkg/githash"
+	"github.com/gittuf/gittuf/pkg/gitstore"
+)
+
+// Hash is the Git object hash the RSL operates over, aliased from
+// githash.Hash (and therefore identical to gitinterface.Hash).
+type Hash = githash.Hash
```
adityasaky: curious, do we need this? since githash is also a public package, trying to understand what this provides. Single import for a consumer of `pkg/rsl` instead of two?

pkg/gitinterface/tree.go
```diff
@@ -324,6 +324,24 @@ func (r *Repository) CreateSubtreeFromUpstreamRepository(upstream *Repository, u
 	return commitID, nil
 }
 
+// WriteTree writes a Git tree containing the specified blobs and subtrees,
+// keyed by their path in the tree, and returns its ID. Blob entries use
+// default permissions. Subtree values are the IDs of existing trees grafted
+// at their path. It covers gittuf's tree-writing needs without exposing the
+// full TreeBuilder. The returned ID is independent of the maps' iteration
+// order because git mktree normalizes entry order.
+func (r *Repository) WriteTree(blobs, subtrees map[string]Hash) (Hash, error) {
```
adityasaky: Trying to make sure I understand this right because the interface is a bit confusing. If we want to write:

```
foo/bar.txt
baz/foobar.txt
```

So two subtrees, two blobs, what would we pass in? And similarly, what would we pass for nested subtrees (`foo/bar/baz.txt`)?

The tree builder infers the subtrees from a map of the blob's relative path and its ID, iirc, though I need to verify this claim of mine.

adityasaky: Per the implementation, blobs will be with their relative paths to the root of the repository and subtrees are those that already exist? Can this be problematic if the blob's relative path and a subtree's relative path to the repo root overlap?

internal/cache/cache.go
```diff
@@ -64,8 +65,7 @@ func (p *Persistent) Commit(repo *gitinterface.Repository) error {
 		return err
 	}
 
-	treeBuilder := gitinterface.NewTreeBuilder(repo)
-	treeID, err := treeBuilder.WriteTreeFromEntries([]gitinterface.TreeEntry{gitinterface.NewEntryBlob(persistentTreeEntryName, blobID)})
+	treeID, err := repo.WriteTree(map[string]gitinterface.Hash{persistentTreeEntryName: blobID}, nil)
 	if err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
pjbgf: False positive, the code compile as-is.

internal/policy/verify.go
```diff
@@ -908,7 +909,7 @@ func verifyEntry(ctx context.Context, repo *gitinterface.Repository, policy *Sta
 	return nil
 }
 
-func verifyTagEntry(ctx context.Context, repo *gitinterface.Repository, policy *State, attestationsState *attestations.Attestations, entry *rsl.ReferenceEntry) error {
+func verifyTagEntry(ctx context.Context, repo gitstore.Storer, policy *State, attestationsState *attestations.Attestations, entry *rsl.ReferenceEntry) error {
 	entryTagRef, err := repo.GetReference(entry.RefName)
 	if err != nil {
 		return err
```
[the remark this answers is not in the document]
pjbgf: Same here, however this pointed out that in some cases we were missing the use of the new interface. That change is incoming.

pjbgf:
Fixed the issues, rebased and collapsed the commits into a single commit. PTAL

pkg/rsl/hash.go
patzielinski, line 1: With the change to pull the `Hash` type out of `gitinterface` and place it instead in `githash`, I'm wondering if we really need this additional (aliased) type here?

pkg/rsl/storer_error_test.go
patzielinski, line 1: This file naming departs from our convention, may I suggest moving this to `repository_test` or renaming as appropriate?

pkg/rsl/storer_error_test.go
```diff
@@ -0,0 +1,232 @@
...
+	"testing"
+
+	artifacts "github.com/gittuf/gittuf/internal/testartifacts"
+	"github.com/gittuf/gittuf/pkg/gitinterface"
+	"github.com/gittuf/gittuf/pkg/gitstore"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// overrideStorer wraps a real Storer, injecting failures for the methods the
```
patzielinski: Nit: `testStorer` or similar instead?

patzielinski: On second thought, could we move this under `gitstore` somehow and export it? IIRC we did something similar with `internal/common` to cut down on the repetition?

internal/policy/signature.go
```diff
@@ -71,14 +73,30 @@ func (v *SignatureVerifier) Verify(ctx context.Context, gitObjectID gitinterface
...
+
+		verifyOpts := []gitobject.Option{}
+		config, err := v.repository.GetGitConfig()
+		if err != nil {
+			return nil, err
+		}
+		if rekorURL, has := config[sigstore.GitConfigRekor]; has {
+			slog.Debug(fmt.Sprintf("Using '%s' as Rekor server...", rekorURL))
+			verifyOpts = append(verifyOpts, gitobject.WithRekorURL(rekorURL))
+		}
```
adityasaky, line 92: I worry a bit about consistently injecting this config for signature verification, anywhere where the sig verification is multiple steps that aren't wrapped with single point of entry. Not sure it blocks this but thoughts on improving this?

adityasaky (commented):
I think this largely looks fine! I have some questions about the new package boundaries and a small concern about how GetObjectSignature changes things by setting new expectations for consumers of the method. cc @patzielinski for eyes too

internal/policy/signature.go
```diff
@@ -71,14 +73,30 @@ func (v *SignatureVerifier) Verify(ctx context.Context, gitObjectID gitinterface
 	gitObjectVerified := false
 
 	// First, verify the gitObject's signature if one is presented
-	if gitObjectID != nil && !gitObjectID.IsZero() {
+	if !gitObjectID.IsZero() {
```
patzielinski, line 76: Can't this still be `nil`?

internal/signerverifier/gitobject/gitobject.go
```diff
@@ -0,0 +1,222 @@
...
+	ErrVerifyingSigstoreSignature = errors.New("unable to verify Sigstore signature")
+	ErrVerifyingSSHSignature      = errors.New("unable to verify SSH signature")
+	ErrMultipleSignatures         = errors.New("object has multiple signatures")
+)
+
+type options struct {
+	rekorURL string
+}
+
+// Option configures Verify.
```
patzielinski:
```suggestion
```

pkg/gitinterface/references.go
```diff
@@ -19,7 +21,9 @@ const (
 )
 
 var (
-	ErrReferenceNotFound = errors.New("requested Git reference not found")
+	// ErrReferenceNotFound aliases gitstore.ErrReferenceNotFound so a
+	// Repository satisfies the gitstore.Storer error contract directly.
```
patzielinski:
```suggestion
```

pkg/gitstore/gitstore.go
```diff
@@ -0,0 +1,117 @@
...
+	// CommitUsingSpecificKey is Commit signing with the provided PEM encoded
+	// key. Intended for gittuf's developer mode and tests.
+	CommitUsingSpecificKey(treeID githash.Hash, targetRef, message string, signingKeyPEMBytes []byte) (githash.Hash, error)
+
+	// ZeroHash returns the all-zeroes hash matching the store's object
+	// format.
+	ZeroHash() githash.Hash
+
+	// GetGitConfig returns the store's Git configuration.
+	GetGitConfig() (map[string]string, error)
```
patzielinski: How would retrieving the config work from backends that don't directly use the Git binary? Would they just pass in values that gittuf requires? If so, I'm concerned about gittuf expecting some value in the config for an important codepath. (I just skimmed a basic search of queries for the Git config being used).

pkg/gitinterface/gitstore_test.go
patzielinski, line 1: Is this needed?

patzielinski (commented):
Thanks for the PR, @pjbgf. I've noted some questions and nits here and there.

I noticed that there's a fair amount of deviation from either the conventions present elsewhere in gittuf (mostly test structure), as well as a fair amount of AI-like text that I personally find hard to mentally parse at times. We can defer a bit of a cleanup to a future PR/issue, as these are just nits.

pkg/gitinterface/gitstore_test.go
pjbgf: I think both assertions are important to avoid long-term issues:
1. `Repository` matches/implements `gitstore.Storer`.
2. The `ErrReferenceNotFound` matches `gitstore`'s error with the same name.

Based on your feedback I'm assuming the issue is mostly the addition of the new file. So I'll move these two things into `repository_test.go` instead. Let me know if you prefer them dropped instead.

pkg/rsl/storer_error_test.go
pjbgf: Good shout, will be moving it to `repository_test`.

pkg/gitstore/gitstore.go
pjbgf: My initial thoughts were to simply return the consolidated (system, global, local and worktree) list. Thinking deeper on this, we seem to only need:

- `user.signingkey`
- `gpg.format`
- `gpg.x509.program`
- `user.name`
- `user.email`
- `gitsign.issuer`
- `gitsign.clientid`
- `gitsign.fulcio`
- `gitsign.rekor`
- `gitsign.redirecturl`
- `gittuf.cache.automatic`

Based on that, I'm now leaning on a simpler approach that returns a single key (still consolidated): `LookupConfig(key ConfigKey) (value string, ok bool, err error)`. I'll add that as a separate commit.

pjbgf: The initial approach was cheaper from the back-end perspective (process all the Configs once, and return the list). The new feels better as an API, but for performance sensitive workloads (e.g. gittuf server-side) the back-end will need to implement some level of caching.

internal/policy/signature.go
pjbgf: Yep. But that is handled via `IsZero()`:

```golang
	if len(h) == 0 {
		return true
	}
```

`len` returns `0` for `nil` slices.

pkg/rsl/storer_error_test.go
pjbgf: Good shout, that was indeed redundant. I'm adding that as `FakeStorer` in a different `internal` package.

pkg/rsl/hash.go
pjbgf: TL;DR: yes, we can drop it, it costs an additional `import` for consumers.

I did several prototypes whilst refactoring. The idea was to actually move `githash.Hash` under `internal`, keeping both aliases so that from an API UX perspective, both `pkg/rsl` and `pkg/gitinterface` had their `Hash`. But it is cleaner to completely remove the aliases and retain `githash.Hash` instead.

I'd drop this alias now and mark the one in `gitinterface` as deprecated, we can drop it on the next major.

pkg/gitinterface/tree.go
pjbgf: I've made some changes to this for a better API UX. The initial thought process was to reduce the need of creating new types, but that didn't work quite well for `Tree` representation. So ended up with `WriteTree(entries []TreeEntry) (githash.Hash, error)` instead.

That way each entry is self-contained. You don't need to create subtrees, unless you are trying to refer to an existing one.

pkg/gitinterface/gitstore_test.go
patzielinski: Sounds good on moving them.

internal/propagation/propagation_storer_test.go
patzielinski, line 1: I think these can be moved to `propagation_test.go`?

patzielinski (commented):
I think this is good enough that after the commits have been cleaned up we can get this in to keep the momentum going. The linter's flagged some test helpers - if it's not a huge delta we should update the consumers otherwise suppress the lint issue and ticketize.

I don't have any further comments to add on this PR.

internal/propagation/propagation_storer_test.go
pjbgf: Good shout, this is now done.

patzielinski (approved):
Approving this; if there are any issues let's address them in a quick followup PR or two.

PR #1512, opened 2026-08-04
rsl: Add support for application-defined custom fields

pjbgf:
## Description

Add support for application-defined custom.* fields in RSL entries. This enables applications to enrich gittuf objects with signed, tamper-evident data, backed by the cryptographic assurances inherent in gittuf.

Keys and values are restricted to a supported character set, enforced on write. For readers, a field whose key or value falls outside the grammar is dropped rather than failing the entry, at most `MaxCount` fields are kept, and the first value for a repeated key wins, so a crafted entry cannot surface unbounded or non-canonical data while traversal still succeeds.

## AI Usage

Opus 4.8 was used mostly around prototyping and testing. All changes were then reviewed, manually tested and verified.

## Contributor Checklist

  request.
  applicable.

### DIFF
--- a/docs/design-document.md
+++ b/docs/design-document.md
@@ -342,20 +342,27 @@ custom.<namespace>/<name>: <value>
 
 RSL entries may include application-defined custom fields after their standard
 fields and before an annotation's message block. Custom field names begin with
-`custom.` followed by one or more of `[A-Za-z0-9._/-]`, and are unique in the
-canonical format. Values may contain `[A-Za-z0-9-+./,()]`. These character sets
-are enforced on write. Readers accept whatever is stored. Writers sort fields by
-name so their encoding is deterministic. Readers accept fields in any order and,
-when a non-canonical entry repeats a name, use its first value. Custom fields
-are part of the RSL commit message and are therefore covered by the entry's Git
-signature.
+`custom.` and follow the Kubernetes annotation convention: a lowercase DNS
+subdomain the application controls (for example `gitforge.com`), a `/`, and a
+name of at most 63 lowercase characters that starts and ends alphanumeric with
+`.`, `_`, or `-` allowed in between. Namespacing keys to a domain you control
+avoids collisions between applications. The whole key is fewer than 250
+characters and unique in the canonical format. Values may contain
+`[A-Za-z0-9-+./,()_ ]`, have no leading or trailing spaces, and are fewer than
+500 characters. An entry carries at most 20 custom fields, and empty values are
+dropped. These constraints are enforced on write. Writers sort fields by name so
+their encoding is deterministic. Readers accept fields in any order and, when a
+non-canonical entry repeats a name, use its first value; fields exceeding the
+length limits are ignored and the count is capped, so a crafted entry cannot
+force gittuf to surface unbounded data. Custom fields are part of the RSL commit
+message and are therefore covered by the entry's Git signature.
 
 For example, a forge might record who pushed a change and the server version
 that processed it:
 
 ```
 custom.gitforge.com/pusher: <handle>(<id>)
-custom.gitforge.com/server-version: v4.2.0-coffee
+custom.gitforge.com/server-version: v4.2.0-c0ffee
 ```
 
 Custom fields do not impact gittuf verification. gittuf treats them as opaque
--- a/docs/design-document.md
+++ b/docs/design-document.md
@@ -309,6 +309,7 @@ RSL Reference Entry
 ref: <ref name>
 targetID: <target ID>
 number: <number>
+custom.<namespace>/<name>: <value>
 ```
 
 The `targetID` is typically the ID of a commit for references that are branches.
@@ -333,10 +334,34 @@ entryID: <RSL entry ID 2>
 ...
 skip: <true/false>
 number: <number>
+custom.<namespace>/<name>: <value>
 -----BEGIN MESSAGE-----
 <message>
 ------END MESSAGE------
 ```
+
+RSL entries may include application-defined custom fields after their standard
+fields and before an annotation's message block. Custom field names begin with
+`custom.` followed by one or more of `[A-Za-z0-9._/-]`, and are unique in the
+canonical format. Values may contain `[A-Za-z0-9-+./,()]`. These character sets
+are enforced on write. Readers accept whatever is stored. Writers sort fields by
+name so their encoding is deterministic. Readers accept fields in any order and,
+when a non-canonical entry repeats a name, use its first value. Custom fields
+are part of the RSL commit message and are therefore covered by the entry's Git
+signature.
+
+For example, a forge might record who pushed a change and the server version
+that processed it:
+
+```
+custom.gitforge.com/pusher: <handle>(<id>)
+custom.gitforge.com/server-version: v4.2.0-coffee
+```
+
+Custom fields do not impact gittuf verification. gittuf treats them as opaque
+metadata: their presence, absence, or contents never change a verification
+outcome. The entry's signature still covers them, so they cannot be tampered
+with, but interpreting them is left entirely to the application.
 
 ##### Example Entries
 
--- a/docs/design-document.md
+++ b/docs/design-document.md
@@ -309,6 +309,7 @@ RSL Reference Entry
 ref: <ref name>
 targetID: <target ID>
 number: <number>
+custom.<namespace>/<name>: <value>
 ```
 
 The `targetID` is typically the ID of a commit for references that are branches.
@@ -333,9 +334,30 @@ entryID: <RSL entry ID 2>
 ...
 skip: <true/false>
 number: <number>
+custom.<namespace>/<name>: <value>
 -----BEGIN MESSAGE-----
 <message>
 ------END MESSAGE------
+```
+
+RSL entries may carry application-defined custom fields after their standard
+fields (for annotation entries, before the message block), each a
+`custom.<namespace>/<name>: <value>` line. Custom fields are advisory only:
+gittuf treats them as opaque metadata whose presence, absence, or contents
+never change a verification outcome. Applications must not make security
+decisions based on them and should treat values as untrusted input unless
+they generated the data themselves. Because the fields are part of the
+entry's commit message, they are covered by the entry's signature and are as
+tamper-evident as the entry itself. The key grammar, value alphabet, length
+and count limits, along with how writers and readers handle malformed fields,
+are specified in [GAP-7](/docs/gaps/7/README.md).
+
+For example, a forge might record who pushed a change and the server version
+that processed it:
+
+```
+custom.gitforge.com/pusher: <id> (<handle>)
+custom.gitforge.com/server-version: v4.2.0-c0ffee
 ```
 
 ##### Example Entries
@@ -348,6 +370,9 @@ entry a5ea2c6ee7e8b577f6be6fcee5b06e6cac7166fa (skipped)
   Ref:    refs/heads/main
   Target: 6cb8e5c546eab3d0e1d245014de8003febb8e9b3
   Number: 5
+  Custom Fields:
+    custom.gitforge.com/pusher: 01ARZ3NDEKTSV4RRFFQ69G5FAV (jane)
+    custom.gitforge.com/server-version: v4.2.0-c0ffee
 
     Annotation ID: cccfb6f27b2a71c33e9a55bc82f084e2445aa398
     Skip:          yes
@@ -404,6 +429,8 @@ RSL Reference Entry
 ref: refs/heads/main
 targetID: 6cb8e5c546eab3d0e1d245014de8003febb8e9b3
 number: 5
+custom.gitforge.com/pusher: 01ARZ3NDEKTSV4RRFFQ69G5FAV (jane)
+custom.gitforge.com/server-version: v4.2.0-c0ffee
 ```
 
 Similarly, the commit object for the annotation entry is as follows:
--- a/docs/gaps/7/README.md
+++ b/docs/gaps/7/README.md
@@ -1,0 +1,327 @@
+# Application-Defined Custom Fields
+
+## Metadata
+
+* **Number:** 7
+* **Title:** Application-Defined Custom Fields
+* **Implemented:** Yes
+* **Withdrawn/Rejected:** No
+* **Sponsors:** Paulo Gomes (pjbgf)
+* **Related GAPs:** [GAP-2](/docs/gaps/2/README.md), [GAP-3](/docs/gaps/3/README.md)
+* **Last Modified:** August 27, 2026
+
+## Abstract
+
+gittuf records every reference update as a signed entry in the Reference State
+Log (RSL). Applications that build on gittuf often have context about a change
+that the standard entry does not capture, such as the actor who pushed it or
+the server that processed it. This GAP introduces custom fields,
+application-defined key-value pairs carried inside an RSL entry's commit
+message. Custom fields are opaque to gittuf and are never consulted during
+verification.
+
+## Motivation
+
+gittuf's RSL is increasingly used as a building block by other software rather
+than only through the gittuf CLI. A forge, a mirroring service, or an internal
+platform tool may want to attach its own context to the record of a reference
+update. Until now the entry schema was closed: the only way to associate extra
+data with an entry was to model it as a separate gittuf construct (e.g., a
+separate annotation entry or an attestation).
+
+The concrete driver is applications that want the RSL to double as an audit
+ledger for a repository. Each push already produces a signed, ordered, and
+tamper-evident record. If that record can also carry a small amount of
+application context, the application gains a durable audit trail that is
+co-located with the change it describes and covered by the same signature.
+
+## Specification
+
+### Field Structure
+
+An RSL entry may carry zero or more custom fields after its standard fields.
+For an annotation entry, they appear before the message block. A field is a
+single line of the form `custom.<namespace>/<name>: <value>`.
+
+```
+RSL Reference Entry
+
+ref: <ref name>
+targetID: <target ID>
+number: <number>
+custom.<namespace>/<name>: <value>
+```
+
+Field names follow the Kubernetes annotation convention: a lowercase DNS
+subdomain the application controls (for example `gitforge.com`), a `/`, and a
+name of at most 63 lowercase characters that starts and ends alphanumeric,
+with `.`, `_`, or `-` permitted in between. Namespacing a key to a domain the
+application controls avoids collisions between applications writing to the
+same RSL.
+
+The whole key, including the `custom.` prefix, MUST be fewer than 250
+characters and MUST be unique within a canonical entry. Values MAY contain
+ASCII letters, digits, and the characters `-+./,()_@:=%#?&~` as well as
+internal spaces. This covers common value shapes such as emails, URLs,
+RFC 3339 timestamps, and base64. Values MUST NOT have leading or trailing
+spaces and MUST be fewer than 500 characters. An entry MUST NOT carry more
+than 20 custom fields.
+
+A recorded reference entry carrying one field looks like:
+
+```
+RSL Reference Entry
+
+ref: refs/heads/main
+targetID: a1b2c3d4e5f60718293a4b5c6d7e8f9012345678
+number: 42
+custom.gitforge.com/pusher: jane (01ARZ3NDEKTSV4RRFFQ69G5FAV)
+```
+
+Parentheses and internal spaces are in the value alphabet, so pairing a handle
+with the forge's account identifier needs no encoding.
+
+Values SHOULD be human-readable, so that an operator inspecting an entry with
+standard Git tooling can read them directly without decoding. Where content
+needs characters the value alphabet does not permit, such as line breaks, the
+application SHOULD base64 the value rather than expect the grammar to widen.
+
+Custom fields are not intended to carry large amounts of data. An application
+with such a payload is better off storing it elsewhere and recording only a
+pointer to it here, for example a URL to an external service or an internal
+Git ref that points to a blob.
+
+### Supported Types
+
+All three RSL entry types carry custom fields: reference entries, annotation
+entries, and propagation entries. The key grammar, value alphabet, and length
+and count limits are identical across them and are enforced by a single shared
+implementation.
+
+Nothing outside the RSL carries custom fields, including policy metadata and
+the commits that record it. The RSL is where an application's context about a
+change belongs, because the record of the change and the context describing it
+then live in one signed object. Policy states describe who is trusted to do
+what, and a policy change is already recorded by an RSL entry, so an
+application that wants to mark one has somewhere to put the mark without
+widening the documents verification reads.
+
+### Reading and Writing
+
+These constraints are enforced on write. Writers drop fields with empty values
+and sort the rest by name so the encoding is deterministic and the resulting
+commit is stable across implementations. An empty value therefore means the
+field is absent rather than present and blank.
+
+Readers require custom fields to appear after the standard fields and reject
+entries that interleave them, but accept the custom fields themselves in any
+order. When a non-canonical entry repeats a name, the reader uses the first
+value it encounters. A field whose key or value falls outside the grammar is
+ignored, as is a field with an empty value, and the total count is capped. The
+rule is the same on both sides, so a field a canonical writer could not have
+produced is never surfaced by a read. This matters because consumers, and
+`gittuf rsl log`, render values as recorded, and an entry is a commit message
+into which anything can be written directly with Git.
+
+Ignoring such a field rather than rejecting the entry keeps a single bad line
+from making an entry unreadable, which would turn a cosmetic defect into a
+traversal failure for a log gittuf still has to walk.
+
+### Relationship to Verification
+
+Custom fields MUST NOT influence verification. gittuf treats them as opaque
+metadata whose presence, absence, or contents never change a verification
+outcome. The entry's signature covers the whole commit message, including the
+custom fields, so they cannot be altered without invalidating the entry. This
+is the only property gittuf provides for them.
+
+### Tooling
+
+The gittuf CLI does not create custom fields. They are aimed at applications
+that embed gittuf as a library and populate the fields programmatically.
+`gittuf rsl log` does render the fields an entry carries, so an operator can
+read them without knowing which names to ask for:
+
+```
+entry a5ea2c6ee7e8b577f6be6fcee5b06e6cac7166fa
+
+  Ref:    refs/heads/main
+  Target: 6cb8e5c546eab3d0e1d245014de8003febb8e9b3
+  Number: 42
+  Custom Fields:
+    custom.gitforge.com/pusher: jane (01ARZ3NDEKTSV4RRFFQ69G5FAV)
+```
+
+The fields remain readable with standard Git tooling as well, since they are
+plain lines in a commit message:
+
+```
+git cat-file -p <rsl-entry-oid>
+```
+
+## Reasoning
+
+### Why not attestations
+
+Attestations (see [GAP-3](/docs/gaps/3/README.md)) are the right tool when the
+extra data is itself security-relevant and needs an independent, verifiable
+structure: a DSSE envelope with a predicate type, stored under a path in
+`refs/gittuf/attestations`. Each attestation is a blob under a path, so
+recording one writes the blob, the trees along its path, and a commit on the
+attestations ref. Because the attestations ref itself moves, that movement in
+turn requires its own RSL entry. A single enriched change can therefore add on
+the order of five objects to the repository. Custom fields are text inside a
+commit message that already exists, so recording a reference update with them
+writes exactly the same single entry commit that gittuf would write anyway.
+For advisory context that does not need to be independently verifiable, the
+attestation cost per change is disproportionate.
+
+### Why not annotations
+
+Annotations are entries in their own right. Using one to attach context to a
+change means writing a second commit on the RSL ref that points back at the
+original entry. That roughly doubles the object cost of recording a change and
+splits the context away from the entry it describes, so a reader has to
+correlate two objects to reconstruct one event.
+
+### Cost of storing data outside the RSL Entry
+
+An RSL reference entry is a single commit whose tree is the empty tree. The
+empty tree object is written once and shared by every entry thereafter, so
+each recorded change contributes one new object to the repository. Custom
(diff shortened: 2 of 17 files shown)


docs/design-document.md
```diff
@@ -333,11 +334,22 @@ entryID: <RSL entry ID 2>
...
+RSL entries may include application-defined custom fields after their standard
+fields and before an annotation's message block. Custom field names begin with
+`custom.` followed by one or more of `[A-Za-z0-9._/-]`, and are unique in the
+canonical format. Values may contain `[A-Za-z0-9-+./,()]`. These character sets
+are enforced on write; readers accept whatever is stored. Writers sort fields by
+name so their encoding is deterministic. Readers accept fields in any order and,
+when a non-canonical entry repeats a name, use its first value. Custom fields
+are part of the RSL commit message and are therefore covered by the entry's Git
+signature.
+
```
adityasaky: We should discuss expectations for the client wrt how to treat custom fields. Can something be added in a custom field such that policy verification fails for a client that understands the custom field but policy verification passes for a "standard" client that doesn't understand the field?

adityasaky: Related: some examples might be worth rolling in ~here~ below.

pjbgf: Good shout. I expanded the doc changes to cover those points.

TL;DR:
- Metadata is opaque for the purposes of gittuf.
- Examples on it being used: a git forge may add the user that pushed the changes and also what version of its software (potentially even gittuf's) that recorded that entry.

pjbgf: This certainly could be followed-up with support for policy verification for such metadata. The prefix-based convention for fields could support that (e.g. map fields with validators).

But that feels like a proposal/GAP of its own.

pkg/rsl/rsl.go
```diff
@@ -92,6 +101,7 @@ type Entry interface {
 	GetID() githash.Hash
 	Commit(gitstore.Storer, bool) error
 	GetNumber() uint64
+	GetCustomFields() CustomFields
 	createCommitMessage(bool) (string, error)
 }
```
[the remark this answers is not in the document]
pjbgf: I'm assuming this would be OK to leave as-is given that this interface is unreleased.

But happy to split it if that feels like the right thing to do.

pkg/rsl/rsl.go
```diff
@@ -511,9 +577,139 @@ func (e *PropagationEntry) createCommitMessage(includeNumber bool) (string, erro
...
+	return cloned
+}
+
+// customFieldDomainLabel matches a single lowercase DNS label. customFieldName
+// matches a Kubernetes-style name segment: lowercase, starting and ending
+// alphanumeric, with '.', '_', and '-' allowed in between.
+var (
+	customFieldDomainLabel = regexp.MustCompile(`^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`)
+	customFieldName        = regexp.MustCompile(`^[a-z0-9]([a-z0-9._-]*[a-z0-9])?$`)
+)
```
pjbgf: I added some strict validation in the proviso that it is always possible to loosen it in the future without breaking backwards compatibility. 

Currently this is up to 20 custom fields per entry, the key-name format is inspired on the kubernetes' annotations. Values are limited to 500 chars.

docs/design-document.md
patzielinski, line 1: @adityasaky would you say a GAP is too much for this or is it something we should have?

I lean towards not needing one but I'd like to get input from others so we make a conscious decision here.

docs/design-document.md
```diff
@@ -333,11 +334,42 @@ entryID: <RSL entry ID 2>
...
 number: <number>
+custom.<namespace>/<name>: <value>
 -----BEGIN MESSAGE-----
 <message>
 ------END MESSAGE------
 ```
 
+RSL entries may include application-defined custom fields after their standard
+fields and before an annotation's message block. Custom field names begin with
+`custom.` and follow the Kubernetes annotation convention: a lowercase DNS
```
patzielinski: This is a stricter naming convention than we have for other custom data (e.g. in principals: https://github.com/gittuf/gittuf/blob/main/internal/tuf/v02/tuf.go#L37). The custom fields serve different purposes but having one adopt a standard and the other be a free-for-all is a bit odd for me.

Is there a specific reason that the k8s standard was chosen?

pkg/rsl/rsl.go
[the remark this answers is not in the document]
patzielinski: It is under `pkg` and we have not yet released `v1.0.0`, so I think this sort of change is OK.

docs/design-document.md
patzielinski: > But that feels like a proposal/GAP of its own.

I agree with this. I am averse to people starting to stuff information in these fields that impacts how a client should verify things without a unified way to verify this information.

We should document these changes on the website, but I can handle that / open an issue.

docs/design-document.md
```diff
@@ -333,11 +334,42 @@ entryID: <RSL entry ID 2>
...
+force gittuf to surface unbounded data. Custom fields are part of the RSL commit
+message and are therefore covered by the entry's Git signature.
+
+For example, a forge might record who pushed a change and the server version
+that processed it:
+
+```
+custom.gitforge.com/pusher: <handle>(<id>)
+custom.gitforge.com/server-version: v4.2.0-c0ffee
+```
```
patzielinski, line 361: This reminds me that we should address our support for signed pushes in gittuf. I am hesitant with this example since if forges start encoding that information as a custom field then we may end up with fragmentation on signed pushes before we even deploy native support.

Take this with a grain of salt though, I'm not sure how much this will be an issue.

docs/design-document.md
```diff
@@ -333,11 +334,42 @@ entryID: <RSL entry ID 2>
...
+
+Custom fields do not impact gittuf verification. gittuf treats them as opaque
+metadata: their presence, absence, or contents never change a verification
+outcome. The entry's signature still covers them, so they cannot be tampered
+with, but interpreting them is left entirely to the application.
```
JustinCappos: What do you mean by 'interpreting them is left entirely to the application'?  

I agree that we absolutely cannot use this info for verification.

So, my thinking is that if we wanted to use this information for verification, some application specific logic would need to read the relevant info and then create another RSL entry which would be used for verification according to the normal gittuf rules.  

Is this correct?

docs/design-document.md
pjbgf: I'll formalise the advisory nature of these fields with code comments to that effect. If/when we change that, we can revisit the comments.

docs/design-document.md
pjbgf: Fair point. I generally prefer a more strict validation as it tends to be easier to contain abuse. No specific reason on using k8s' approach though apart from my experience with it. The naming ensures no collision with built-in fields and can easily then support external plugins that work on a given context. 

Do we have good examples on how the existing `Person.Custom` is being used in the wild?

docs/design-document.md
pjbgf: For this example specifically, I think there are two parts to this: a) metadata and b) security assurances. 
The metadata here can provide some level of **information** about a push/pusher. A forge may support mirroring of repositories, whereby the source of truth lives elsewhere. In such cases, this could indicate whether this was a direct push or a "back-fill" push to catch-up with the source of truth - the latter being the pusher was an internal service.

This does not provide cryptographic assurances about the user. It only describes who the forge _thinks_ was the last actor before the `git-receive-pack` was processed. This enables the [Configuration F in GAP 2](https://github.com/gittuf/gittuf/pull/1523), whereby a RSL can be established without any `gittuf` tooling or user awareness. For user-identity assurance, we would then need to lean on existing primitives - e.g. the user creating the RSL entry/attestation themselves.

A forge could support/enforce the extremes (user-supplied RSL entries) as well as a mixed approach. But the idea is that, in all cases, `gittuf` would provide that consistent mechanism that ensures the RSL consistency. In repositories whereby users lean on the `gittuf` CLI, they would get the additional identity assurance, otherwise not.

docs/design-document.md
pjbgf: > What do you mean by 'interpreting them is left entirely to the application'?

Largely that the application may have added this data for a specific reason which may later be needed either for tracing/debugging (version of gittuf/forge) or to show to an user. An example for the latter being distinguishing an internal push vs external as per [pusher example above](https://github.com/gittuf/gittuf/pull/1512#discussion_r3758956760).

Using this on an audit log viewer built on the repository's RSL data, the application could for example convey the different levels of assurance:

- Entries provided by users themselves (or otherwise backed by gittuf's security primitives) are highlighted as higher assurance.
- Everything else falls back to the informational, unverified tier.

This aligns with a forge in [Configuration F](https://github.com/gittuf/gittuf/pull/1523), so that every single user push generate a RSL entry, regardless of them being aware of it. The RSL consistency is always enforced. But then if the forge supports user-provided RSL entries, you get the additional assurance of user identity.

> my thinking is that if we wanted to use this information for verification, some application specific logic would need to read the relevant info and then create another RSL entry which would be used for verification according to the normal gittuf rules

Absolutely. IMO, that would then either mean that first the application would run gittuf's verification logic and on success carry on its own heuristics. Or, we would come up with some sort of plugin system to enforce specific entries are verified also by the external plugin. But the latter could bring a lot into gittuf's attack surface which may not be ideal.

docs/design-document.md
adityasaky: GAP is now part of the PR :D

docs/gaps/7/README.md
```diff
@@ -0,0 +1,242 @@
...
+Custom fields are opaque to gittuf and are never consulted during
+verification.
+
+## Motivation
+
+gittuf's RSL is increasingly used as a building block by other software rather
+than only through the gittuf CLI. A forge, a mirroring service, or an internal
+platform tool may want to attach its own context to the record of a reference
+update. Until now the entry schema was closed: the only way to associate extra
+data with an entry was to model it as a separate gittuf construct.
```
adityasaky:
```suggestion
platform tool may want to attach its own context to the record of a reference
update. Until now the entry schema was closed: the only way to associate extra
data with an entry was to model it as a separate gittuf construct (e.g., a separate annotation entry or an attestation).
```

docs/gaps/7/README.md
```diff
@@ -0,0 +1,242 @@
...
+application controls avoids collisions between applications writing to the
+same RSL.
+
+The whole key, including the `custom.` prefix, MUST be fewer than 250
+characters and MUST be unique within a canonical entry. Values MAY contain
+ASCII letters, digits, and the characters `-+./,()_@:=%#?&~` as well as
+internal spaces. This covers common value shapes such as emails, URLs,
+RFC 3339 timestamps, and base64. Values MUST NOT have leading or trailing
+spaces and MUST be fewer than 500 characters. An entry MUST NOT carry more
+than 20 custom fields.
```
adityasaky, line 68: To cover our bases with validity of values, should we b64 encode the value? We do this for annotation messages to not have to make our lives easier with parsing and validation.

adityasaky: If yes, different question then about whether the 500 character limit applies to the raw value or the encoded version (probably the raw value makes more sense).

docs/gaps/7/README.md
```diff
@@ -0,0 +1,242 @@
...
+### Why not annotations
+
+Annotations are entries in their own right. Using one to attach context to a
+change means writing a second commit on the RSL ref that points back at the
+original entry. That roughly doubles the object cost of recording a change and
+splits the context away from the entry it describes, so a reader has to
+correlate two objects to reconstruct one event. Annotations also carry a
+specific semantic in gittuf, namely adding notes to or skipping prior entries,
+and overloading them to carry arbitrary application metadata muddies that
+meaning.
```
adityasaky:
```suggestion
correlate two objects to reconstruct one event.
```

I think the "adding arbitrary additional data" bit is one of the primary motivations for annotations, actually. I think the overhead argument is the key one here.

docs/gaps/7/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+Measured against a year of real reference activity, that is about 900 objects
+for go-git and 3,500 for kubernetes. Higher-traffic repositories see larger
+numbers, but the cost stays at one object per change, so an entry-only RSL
+remains cheap to store, walk, and pack. The same year recorded as
+attestations would cost several times as many objects, and those objects are
+heavier: distinct blobs and trees delta far less cleanly than near-identical
+RSL commits do. Object count is what connectivity checks, reachability walks,
+and bitmap builds pay for, so keeping enrichment inside the entry keeps those
+operations cheap.
```
adityasaky: Could you add a table with the concrete numbers and the graph we discussed the other day?

docs/gaps/7/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+* **Number:** 7
+* **Title:** Application-Defined Custom Fields
+* **Implemented:** Yes
+* **Withdrawn/Rejected:** No
+* **Sponsors:** Paulo Gomes (pjbgf)
+* **Related GAPs:** [GAP-2](/docs/gaps/2/README.md), [GAP-3](/docs/gaps/3/README.md)
+* **Last Modified:** August 19, 2026
+
+## Abstract
```
adityasaky, line 13: Could you add concrete examples for the RSL entries? I'd also like to see examples for these fields / usage in the root metadata. We might have more to consider there.

adityasaky: I also wonder if we should just go to v0.3 if we do this.

adityasaky: And cc @JustinCappos -- how does TUF handle custom fields today?

docs/gaps/7/README.md
patzielinski: I imagine part of the idea here was to make this user-readable given we have `gittuf rsl log`?

docs/gaps/7/README.md
patzielinski: +1 on the examples; with the root metadata, it looks to be a strict increase of the data that's stored?

docs/gaps/7/README.md
```diff
@@ -0,0 +1,242 @@
...
+Annotations are entries in their own right. Using one to attach context to a
+change means writing a second commit on the RSL ref that points back at the
+original entry. That roughly doubles the object cost of recording a change and
+splits the context away from the entry it describes, so a reader has to
+correlate two objects to reconstruct one event. Annotations also carry a
+specific semantic in gittuf, namely adding notes to or skipping prior entries,
+and overloading them to carry arbitrary application metadata muddies that
+meaning.
+
+### Cost of an audit ledger
```
patzielinski: This seems a bit confusing given that's the point of the RSL? Something to the effect of `Cost of storing data outside the RSL` or similar seems clearer to me.

docs/gaps/7/README.md
```diff
@@ -0,0 +1,242 @@
...
+metadata whose presence, absence, or contents never change a verification
+outcome. The entry's signature covers the whole commit message, including the
+custom fields, so they cannot be altered without invalidating the entry.
+Custom fields in root metadata are likewise covered by the DSSE signature
+over the metadata payload. This is the only property gittuf provides for
+them.
+
+### Tooling
+
+The gittuf CLI does not create or render custom fields. They are aimed at
```
patzielinski: Should we add the requisite functionality to `gittuf rsl log` to read and display this data at least?

docs/design-document.md
patzielinski: Good points. Given the details in the GAP I think this is sufficient.

internal/tuf/v02/root.go
```diff
@@ -1067,4 +1073,35 @@ func (r *RootMetadata) GetHooks(stage tuf.HookStage) ([]tuf.Hook, error) {
 	return hooks, nil
 }
 
+// The custom-field methods make RootMetadata satisfy customfields.Store.
+// Fields share the format and validation rules of RSL entry custom fields.
+var _ customfields.Store = (*RootMetadata)(nil)
```
patzielinski: Is there a reason this is being assigned to the blank identifier?

pkg/customfields/customfields.go
patzielinski, line 1: +1 on centralizing this as `customfields`.

pkg/customfields/customfields.go
```diff
@@ -0,0 +1,130 @@
+// Copyright The gittuf Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package customfields defines the shared format for signed,
```
patzielinski: @adityasaky this relates to my earlier note on whether we should unify the custom fields contained in `Principal` with the format validation going on here. We don't have to do this here but it'd be good to get a direction on whether we want to do that or intentionally keep them separate.

docs/gaps/7/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+gittuf records every reference update as a signed entry in the Reference State
+Log (RSL). Applications that build on gittuf often have context about a change
+that the standard entry does not capture, such as the actor who pushed it or
+the server that processed it. This GAP introduces custom fields,
+application-defined key-value pairs carried inside an RSL entry's commit
+message. The same fields, with the same format and validation rules, may also
+be carried in root metadata from schema version v0.2 onwards.
+Custom fields are opaque to gittuf and are never consulted during
+verification.
```
JustinCappos: If we are setting these up to be used by Cedar, then I think this isn't true

docs/gaps/7/README.md
```diff
@@ -0,0 +1,242 @@
...
+remains cheap to store, walk, and pack. The same year recorded as
+attestations would cost several times as many objects, and those objects are
+heavier: distinct blobs and trees delta far less cleanly than near-identical
+RSL commits do. Object count is what connectivity checks, reachability walks,
+and bitmap builds pay for, so keeping enrichment inside the entry keeps those
+operations cheap.
+
+## Backwards Compatibility
+
+This GAP does not impact the backwards compatibility of gittuf. Entries
```
JustinCappos, line 235: Do existing clients need to know to ignore custom metadata?

docs/gaps/7/README.md
pjbgf: Based on discussion around this, I'm changing this to briefly say:
- User-readable input is preferable so that no additional tooling is required to read contents.
- If content to be stored requires characters that break validation (i.e. line breaks), the application could choose to base-64 it.
- These fields are not meant to carry larges amounts of data. If an application requires such use case, they are better off storing such data else where and adding here just a pointer to it. For example, an URL to an external server, or an internal Git ref that points to a blobs.

My take is that this is opaque to gittuf, so we don't need to be very prescriptive. Would that align with your thoughts @patzielinski @adityasaky?

docs/gaps/7/README.md
pjbgf: No, it is transparent to them. The `gittuf` cli parses only the built-in headers for the verification process.

internal/tuf/v02/root.go
pjbgf: Yes, to ensure at compile time that `RootMetadata` continues to support the `customfields.Store` interface.

Some people prefer this being placed in tests, others on the file that defines the struct being checked. I don't have strong opinions on this, but I generally go for the latter.

docs/gaps/7/README.md
pjbgf: At this point in time we have no use-cases for Cedar validation on custom fields. There was some initial consideration on using this to block policy/RSL from being transplanted across repositories, but that is no longer the case. For custom fields that are owned by the forge, it can simply reject pushes (on pre-receive) with entries that try to use them, completely out-of-band of gittuf verification.

My take was that Cedar verification would take an out-of-band input, on top of Git state, that covers things like the user acting on the operation and teams/groups they are part of. Does that align with your view?

docs/gaps/7/README.md
pjbgf: I'm a bit torn by this. At first I thought so, but considering that this is to be opaque, and to not have an impact on verification, this may induce an end-user to make assumptions of this data which they probably shouldn't.

Would you rather we change that, or just use less "final" wording? Something like "At this point in time,".

docs/gaps/7/README.md
pjbgf: Added a table to this effect, please let me know if that's OK of if you want other projects to be used.

pjbgf:
Thanks for the initial review, I've made a few changes PTAL. Here's a quick summary:
- Fixes based on overall feedback.
- Remove Root metadata custom fields and add them on the Policy commits instead, aligning with the RSL approach. Making it cheaper to read/write.
- GAP-7: Clarify preference on human-readable values, base64 as the escape hatch, pointers instead of large payloads. RSL entry and root metadata examples. Write-path namespace enforcement note and object-cost table.
- GAP-7: Added a new "Not providing native support" alternative.

docs/gaps/7/README.md
```diff
@@ -0,0 +1,350 @@
...
+with `.`, `_`, or `-` permitted in between. Namespacing a key to a domain the
+application controls avoids collisions between applications writing to the
+same RSL.
+
+The whole key, including the `custom.` prefix, MUST be fewer than 250
+characters and MUST be unique within a canonical entry. Values MAY contain
+ASCII letters, digits, and the characters `-+./,()_@:=%#?&~` as well as
+internal spaces. This covers common value shapes such as emails, URLs,
+RFC 3339 timestamps, and base64. Values MUST NOT have leading or trailing
+spaces and MUST be fewer than 500 characters. An entry MUST NOT carry more
```
adityasaky, line 67: It's covered because the allowed characters are enumerated and so are internal spaces but I'd just add a very explicit note that newlines are not allowed.

adityasaky:
```
Where content
needs characters the value alphabet does not permit, such as line breaks, the
application SHOULD base64 the value rather than expect the grammar to widen.
```

Okay, I was getting a bit ahead of myself.

docs/gaps/7/README.md
```diff
@@ -0,0 +1,350 @@
...
+Git ref that points to a blob.
+
+### Supported Types
+
+All three RSL entry types carry custom fields: reference entries, annotation
+entries, and propagation entries. The commit that records a policy state
+carries them as well, on `refs/gittuf/policy-staging` and, once the state is
+applied, on `refs/gittuf/policy`. Applying a state fast-forwards the policy
+reference to the staged commit, so both references name the same commit object
+and therefore the same fields. Policy support exists for consistency: an
```
adityasaky: This gets a bit complicated because the commits outside of those that are RSL entries don't have a defined structure today. We've tried to surface information in commits for different policy changes but they've not been written to be consistently parse-able.

adityasaky: I don't want to mix types and be inconsistent, but should those be commit trailers that are reasonably well understood?

adityasaky: I believe that'd make the `key` a subset of what's allowed by the K8s custom fields spec.

adityasaky: If the policy commit is only reachable once it's already recorded in the RSL, could the commit specific custom fields just live in the RSL entry for the policy ref?

docs/gaps/7/README.md
adityasaky: I think "at this point in time" is fine. I personally think we should consider displaying the fields in `gittuf rsl log` but still have guardrails in the implementation to prevent usage in verification. If we ever want to plumb that through for verification context, we probably want a follow-up GAP.

docs/gaps/7/README.md
```diff
@@ -0,0 +1,350 @@
...
+signature over the entry's commit message covers them, so a recorded value
+cannot be altered without detection, but this does not mean the value was true
+when it was written. A forge that records `custom.gitforge.com/pusher` is
+asserting that claim, and gittuf attests only that the claim has not changed
+since the entry was signed. Data that needs to be independently verifiable
+belongs in an attestation (see [GAP-3](/docs/gaps/3/README.md)), not a custom
+field.
+
+Consumers MUST treat custom fields as untrusted input and make no assumptions
+about their contents unless they generated the data themselves. The intended
```
adityasaky:
```suggestion
about their contents unless they generated the data themselves, validated by the signature on the object. The intended
```

docs/gaps/7/README.md
adityasaky: Option A: put it in the commit message in refs/gittuf/policy

Option B: put it in the RSL entry recording an update to refs/gittuf/policy

Option C: put it in the policy metadata itself


Between the three, I'm coming around to thinking C > B > A. There's _some_ precedent for this in TUF if not exactly the same thing (https://theupdateframework.github.io/specification/v1.0.33/#custom). I think a top-level `custom` field could work well here. It also means that the data becomes self descriptive to each piece of metadata rather than surfaced at the commit layer. This becomes even more important as a commit might have modified multiple files with distinct properties, complicating the parsing of both option A and option B in surfacing this nuance.

Between A and B I prefer B because RSL entry commit messages have always had structure and it's well understood.

pjbgf: I do like the idea of custom fields becoming a consistent feature across gittuf primitives: pretty much like Kubernetes does with `Annotations`, enabling applications to extend/enrich workflows without requiring in-tree changes.

That being said, we probably could do `C` as its own PR, and restrict the scope of this PR/GAP back to RSL only. Would that work for you?

internal/tuf/v02/root.go
patzielinski: Ah, I see.

docs/gaps/7/README.md
pjbgf: The `B` can be a good alternative in the meantime. As you only need to process the raw RSL commit, as opposed to finding and parsing the policy for the custom fields. The only issue is optimising `policy change -> rsl entry`. IIRC, we only map the reverse via "target".

pjbgf:
As previously discussed:
- The scope was reduced back to RSL: GAP, PR title and description should reflect that.
- Applied all requested changes.
- Updated `gittuf rsl log` to display the custom fields when present.

PTAL

pkg/customfields/customfields.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	return nil
+}
+
+// ValidKey reports whether key is a well-formed field key: the Prefix, a
+// lowercase DNS subdomain, a '/', and a name of at most 63 characters that
+// starts and ends with a lowercase letter or digit and may carry '.', '_', and
+// '-' in between. The whole key, Prefix included, must be fewer than
+// MaxKeyLength characters.
+func ValidKey(key string) bool {
```
adityasaky, line 89: nit: do we need to export this?

pkg/customfields/customfields.go
```diff
@@ -0,0 +1,166 @@
...
+func validNameSegment(name string) bool {
+	return len(name) > 0 && len(name) <= 63 && nameSegment.MatchString(name)
+}
+
+// ValidValue reports whether value is a well-formed field value: fewer than
+// MaxValueLength characters, no leading or trailing spaces, and every rune an
+// ASCII letter, digit, or one of the characters in ValuePunctuation. This is a
+// check of a value's shape, so the empty string passes. Emptiness means the
+// field is absent, which writers handle by dropping the field.
+func ValidValue(value string) bool {
```
adityasaky, line 121: nit: do we need to export this?

pkg/rsl/rsl.go
```diff
@@ -1386,6 +1501,35 @@ func setNumber(dst *uint64, value string) error {
...
+		*fields = CustomFields{}
+	}
+	if _, exists := (*fields)[key]; exists {
+		return
+	}
+	if len(*fields) >= customfields.MaxCount {
+		return
+	}
+	(*fields)[key] = value
+}
```
adityasaky, line 1531: Should this use `Validate` that handles the max count, as well as key and value validation?

pkg/customfields/customfields.go
```diff
@@ -0,0 +1,166 @@
...
+	keys := make([]string, 0, len(effective))
+	for key := range effective {
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+	for _, key := range keys {
+		lines = append(lines, fmt.Sprintf("%s: %s", key, effective[key]))
+	}
+	return lines, nil
+}
```
adityasaky, line 166: if this is appending only for the purposes of writing an RSL entry, I'd move this into the RSL package so the logic for both writing and parsing RSL entries live in the same place. It also makes it easier to ensure validation is consistent across both flows. I think this specific function will not help us when / if we add custom fields to things like gittuf metadata?

adityasaky (commented):
LGTM! My comments are quite trivial: I'd personally prefer we keep the logic to create and parse RSL entries with custom fields in a single place and also consolidate the validation logic into a single workflow used on both ends.

pkg/customfields/customfields.go
pjbgf: This has been dropped, and replaced with `ValidateField`. This is important to support verifying each field as a consumer in a cheap/efficient manner.

pkg/customfields/customfields.go
pjbgf: Same as below: This has been dropped, and replaced with `ValidateField`. This is important to support verifying each field as a consumer in a cheap/efficient manner.

pkg/rsl/rsl.go
pjbgf: Great shout, fixed on the latest version.

pkg/customfields/customfields.go
pjbgf: This is now moved to `appendCustomFieldLines` in `pkg/rsl`. This was a hangover from the PR scope expansion.

pkg/rsl/rsl.go
```diff
@@ -1181,6 +1233,15 @@ func parseReferenceEntryText(id githash.Hash, text string) (*ReferenceEntry, err
...
 			state = done
+
+		default:
+			if strings.HasPrefix(key, customfields.Prefix) {
+				if state < expectNumber {
+					return nil, ErrInvalidRSLEntry
+				}
+				state = done
+				setCustomField(&entry.CustomFields, key, value)
+			}
```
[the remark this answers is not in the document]
pjbgf: Custom fields appearing ahead of `Number` is a malformed state. Added tests to pin the behaviour.

pkg/rsl/rsl.go
```diff
@@ -1181,6 +1233,15 @@ func parseReferenceEntryText(id githash.Hash, text string) (*ReferenceEntry, err
...
 		}
```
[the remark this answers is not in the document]
pjbgf: The GAP covers it. Canonical writer never writes such fields. Reader is more accommodating, but still enforces the correct grammar.

patzielinski (approved):
Thanks @pjbgf!

PR #1523, opened 2026-08-11
docs: expand GAP-2 with server-managed root of trust

pjbgf:
## Description

Adds Configuration `F` to GAP-2 (gittuf on the Forge), in which the forge owns the root of trust and is responsible for generating and signing RSL entries. It performs verification at the `pre-receive` phase and rejects pushes that fail. Users push ordinary Git references with no client-side gittuf tooling and no bootstrap ceremony, and can still use the gittuf CLI to verify policy against the forge's published keys.

This is the lowest entry bar for a forge becoming gittuf-aware: every hosted repository is verifiable without asking users to run `gittuf`. From there, authority can be handed back incrementally toward the more client-driven configurations (e.g. co-owning the root of trust, accepting user-generated RSL entries and attestations), up to whichever split of duties a repository's owners want.

## AI Usage

- [x] I **did not** use generative AI at all in making the content of this pull
  request.
- [ ] I **did** use generative AI in some form in making the content of this

## Contributor Checklist

  request.
  applicable.

### DIFF
--- a/docs/gaps/2/README.md
+++ b/docs/gaps/2/README.md
@@ -6,8 +6,8 @@
 * **Title:** gittuf on the Forge
 * **Implemented:** No
 * **Withdrawn/Rejected:** No
-* **Sponsors:** Aditya Sirish A Yelgundhalli (adityasaky)
-* **Last Modified:** January 20, 2025
+* **Sponsors:** Aditya Sirish A Yelgundhalli (adityasaky), Paulo Gomes (pjbgf)
+* **Last Modified:** August 11, 2026
 
 ## Abstract
 
@@ -54,17 +54,77 @@ in gittuf:
 and signed by clients when they push their changes. Thus, every push can be
 authenticated using the signature on the RSL entry, and the synchronization
 point is not responsible for ordering pushes in the repository meaning it cannot
-reorder or drop pushes.
+reorder or drop pushes. When the forge creates RSL entries (Configurations C to
+F), it may enrich them with custom fields (e.g., a stable repository identifier or
+an attestation of how it authenticated the pushing user). Such fields are advisory
+at this point in time.
 * Verification: in the standard gittuf model, every gittuf client performs
 verification when it receives changes from the synchronization point. Typically,
 this means that a change that fails verification must be fixed after the fact.
+Verification is not limited to clients. The forge can verify entries as well, and
+in some configurations it does so before persisting them. Where the forge
+verifies at the pre-receive phase, it can enforce that only valid and policy
+abiding entries are ever persisted, rejecting the rest instead of relying on a
+client to detect and repair them later.
 * Git Reference Updates: in the standard gittuf model, a gittuf client pushes
 directly to the references on the synchronization point the user wishes to
 update along with corresponding RSL entries. The RSL entries are submitted to
 the synchronization point's RSL directly after the client fetches the latest
 state of the RSL to ensure the new entries are added at the very end of the RSL.
 The gittuf client makes this push atomically, meaning either all references are
-updated or none are updated,
+updated or none are updated.
+* Root-of-Trust Ownership: in the standard gittuf model the root of trust
+(`refs/gittuf/policy`) is authored and signed by repository owners on the client,
+and the synchronization point only stores it. A forge may instead own the root of
+trust: it authors and signs root metadata and policy, and clients treat
+`refs/gittuf/policy` as read-only. To avoid locking a repository to a single
+forge, a forge that owns the root of trust must either co-own it with the
+repository owner (the owner retains an authorized signing key), or provide a
+documented takeover path so the owner can reclaim sole authority over the root
+and migrate to another forge. This changes who must be trusted for authorization
+decisions.
+
+### Reference namespace and push semantics
+
+Independent of configuration, a gittuf-aware forge should maintain one invariant
+over the reference namespace: the set of references it hides from advertisement
+and the set it rejects on push are the same set. A reference is either a user or
+verification surface (advertised and, where applicable, writable) or it is
+internal or client-local (hidden and unwritable). Nothing is hidden-but-writable
+or advertised-but-reserved. In particular, `refs/local/gittuf/**` is client-local
+and must be both rejected on push and unadvertised on fetch. A push that updates
+more than one gittuf reference (`refs/gittuf/policy`, `refs/gittuf/policy-staging`,
+`refs/gittuf/reference-state-log`, `refs/gittuf/attestations`) should be applied
+atomically, since a policy change and its RSL entry form one logical trust
+transaction.
+
+Which of these references are client-writable and advertised depends on the
+configuration. In configurations where clients own the root of trust,
+`refs/gittuf/policy-staging` is a shared, advertised reference that maintainers
+push to in order to collaboratively author and sign policy before it is promoted
+to `refs/gittuf/policy`. Where the forge owns the root of trust, that same
+reference is internal to the forge (see Configuration F).
+
+This invariant also bounds what the forge signs. When the forge creates RSL
+entries, it signs an entry for a reference only if that reference is
+client-visible, with the RSL reference itself as the single carve-out since it
+cannot record itself. Signing for a hidden reference would defeat the invariant
+and could leak internal or client-local reference spaces to anyone verifying the
+RSL.
+
+A mixed push updates both gittuf references and user references in one push. As an
+initial approach, mixed pushes need not be atomic across the two classes. The
+gittuf reference updates are kept regardless of whether the user references pass
+verification. User references that fail are rejected, and the gittuf updates in
+the same push are retained. This keeps the trust state advancing on its own terms
+and lets a client retry the content without re-authoring metadata. The non-atomic
+phase must preserve one invariant. A retained gittuf update must be independently
+valid on its own, so that a rejected user reference never leaves dangling or
+unverifiable metadata behind. Full atomicity, where a mixed push either lands
+entirely or not at all, is the intended end state.
+
+TODO: Specify the exact commit protocol for fully atomic mixed pushes, and whether
+these namespace rules are normative for all configurations.
 
 ### Configuration A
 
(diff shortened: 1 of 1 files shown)


docs/gaps/2/README.md
```diff
@@ -54,17 +54,72 @@ in gittuf:
...
+
+Independent of configuration, a gittuf-aware forge should maintain one invariant
+over the reference namespace: the set of references it hides from advertisement
+and the set it rejects on push are the same set. A reference is either a user or
+verification surface (advertised and, where applicable, writable) or it is
+internal or client-local (hidden and unwritable). Nothing is hidden-but-writable
+or advertised-but-reserved. In particular, `refs/local/gittuf/**` is client-local
+and must be both rejected on push and unadvertised on fetch, and
+`refs/gittuf/policy-staging` is server-managed and server-only, so it is never
+advertised. A push that updates more than one server-managed gittuf reference
```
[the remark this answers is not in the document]
pjbgf: I don't think this is accurate in this context, although it could be when considering Configuration F. I will improve the wording to align with intent.

docs/gaps/2/README.md
```diff
@@ -54,17 +54,77 @@ in gittuf:
 and signed by clients when they push their changes. Thus, every push can be
 authenticated using the signature on the RSL entry, and the synchronization
 point is not responsible for ordering pushes in the repository meaning it cannot
-reorder or drop pushes.
+reorder or drop pushes. When the forge creates RSL entries (Configurations C to
+F), it may enrich them with custom fields (e.g., a stable repository identifier or
```
JustinCappos, line 58:
```suggestion
F below), it may enrich them with custom fields (e.g., a stable repository identifier or
```

docs/gaps/2/README.md
```diff
@@ -54,17 +54,77 @@ in gittuf:
...
+repository owner (the owner retains an authorized signing key), or provide a
+documented takeover path so the owner can reclaim sole authority over the root
+and migrate to another forge. This changes who must be trusted for authorization
+decisions.
+
+### Reference namespace and push semantics
+
+Independent of configuration, a gittuf-aware forge should maintain one invariant
+over the reference namespace: the set of references it hides from advertisement
+and the set it rejects on push are the same set. A reference is either a user or
```
JustinCappos, line 91: I don't understand what you mean by 'references it hides from advertisement'

JustinCappos (commented):
I wonder if this is getting to be a little much with all of these configurations.  If there is a strong belief (especially with some operational experience) for a single configuration, I think we might laser focus this GAP on it and move all else to the periphery.

docs/gaps/2/README.md
PratGit1606: Feel free to correct me, but from my understanding, the references it hides from advertisement are those that the forge does not include in what it shows clients during a git fetch. So, when a client would run git fetch, the forge would only send a list of available refs, essentially "advertising" them. Refs that are internal to the forge are excluded from that list, meaning clients can't see or fetch them.

docs/gaps/2/README.md
```diff
@@ -229,14 +289,134 @@ is in the RSL next time?
...
 ## Backwards Compatibility
 
+Configuration F is additive: it introduces a new configuration, a root-of-trust
+ownership axis, and namespace guidance without changing the meaning of A to E. The
+RSL byte format, policy metadata format, and reference names are unchanged, so a
+client verifying a Configuration-F repository uses the standard gittuf verify
+path against the forge-published root with no client-side format change. Custom
+RSL fields are forward-compatible, as existing clients tolerate unknown fields.
+
+Because the format is unchanged, a forge can start with Configuration F and later
```
PratGit1606, line 366: The seamless transition is definitely a strong priority, but I had a question related to this that I think Robin also brought up in today's meeting. Let's say a repo transitions from config F to config C. How does a verifier determine which RSL entries were forge signed versus user signed? Since the format is identical, I assume the RSL entry would also become ambiguous across the configuration. In that situation, would it be worth differentiating between gittuf signatures that were created by the forge and ones that were created on the user's device?

docs/gaps/2/README.md
```diff
@@ -229,14 +289,134 @@ is in the RSL next time?
...
+
+**Cons:**
+* The forge is trusted for authorization and not only for ordering. It can author
+any policy and sign any RSL entry. This is the strongest trust assumption in
+GAP-2's spectrum. It is bounded, though not eliminated, by out-of-band publication
+of the forge's keys and tiered metadata expiry (defined by TUF but not yet
+implemented in gittuf).
+* Owning the root of trust risks locking a repository to the forge unless
+co-ownership or a takeover path is provided.
+* First contact is trust-on-first-use unless an out-of-band anchor (an
```
PratGit1606, line 343: Having these three anchoring mechanisms listed is helpful as it gives more options, but neither are shown as having more priority than the rest. So If a repo migrates from one forge to another and each uses a different mechanism, how would a client or auditor verify consistently across both? Should Config F prescribe a preferred mechanism while still leaving the others as valid alternatives, or are all three intended as completely independent and equally valid choices?

docs/gaps/2/README.md
```diff
@@ -229,14 +289,134 @@ is in the RSL next time?
...
+owners want.
+
+This transition spans a wide range of intermediate options. At one end the server
+owns everything as in Configuration F. Moving along the range, the root of trust
+can be handed back to users entirely, removing it from the server, while the
+server is still granted a narrower role in policy for the operations the forge
+wants to enforce (e.g., policy may require the server to sign or attest a given
+operation such as an RSL entry). In these cases the server retains no root
+authority but remains an authorized signer for its assigned role, and it can
+reject any push whose policy change would remove the server from that role. The
```
PratGit1606, line 380: Having a policy in place to have the forge reject any push that removes it from its role is definitely helpful to avoid accidental removals but how would a legitimate operator intentionally reassign or remove a particular server from that role? Would there be a defined override path for this, or is it still one of the cases where the takeover path TODO needs to address it?

docs/gaps/2/README.md
```diff
@@ -229,14 +289,134 @@ is in the RSL next time?
...
+with tooling) is layered on top.
+* When the root is shared fleet-wide, root-key loss or compromise is a
+fleet-scale event. Per-repository roots reduce that blast radius, but at
+server-held custody the forge holds every root key regardless of scope, so
+containment is limited unless root custody is offline.
+
+TODO: Where does the repository-identity binding live, and can gittuf make a
+custom RSL field verification-relevant rather than merely tolerated?
+
+TODO: Define the exact protocol for co-ownership or owner takeover of a
```
PratGit1606, line 354: I was thinking through possible approaches for the takeover path TODO and ran into a bit of a dilemma. I explored a few ideas for this:

The first idea I considered was co-ownership. The owner holds the root signing key alongside the forge, and they can change the root metadata for this. The problem I thought about for this was that it reintroduces a client-side signing requirement, which just contradicts the use case for having Configuration F in the first place for zero client-side ceremony.

The second idea I am thinking about is to have a key transparency log similar to how TLS has certificate transparency. The forge publishes the root key to an append-only log at initialization, creating a verifiable anchor for migration. This feels a bit circular because the root key is server-held, so the owner never has it. Any mechanism requiring the owner to hold a key is circular, and any mechanism requiring forge cooperation defeats the purpose of having a takeover path.

Would appreciate hearing your thoughts on how to break out of this loop if you guys have thought about it as well!

docs/gaps/2/README.md
pjbgf: > the references it hides from advertisement are those that the forge does not include in what it shows clients during a git fetch

That is correct. Forges may choose to use some ref spaces internally to either store metadata co-located with a Git repository, or just to ensure that objects are not garbage collected. An example of the latter being `refs/keep-around/*` and `refs/tmp/*` from [GitLab](https://gitlab.com/gitlab-org/gitaly/-/merge_requests/4727).

pjbgf:
> I wonder if this is getting to be a little much with all of these configurations. If there is a strong belief (especially with some operational experience) for a single configuration, I think we might laser focus this GAP on it and move all else to the periphery.

@JustinCappos My gut instinct is saying we could drop the configurations and instead focus on (co) ownership of the specific roles within `gittuf`, and we take one point at a time. For example, what does it mean for RoT to be owned by an user, the server and both. Same for policies and the RSL. That way it becomes role-specific. A forge could start by owning the RoT completely, but then move to supporting user provided RoT. That means they only need to consider that variable, not the different configurations/permutations.

@adityasaky any thoughts on changing the angle of the doc to that effect?

docs/gaps/2/README.md
pjbgf: The signature fingerprint is part of the RSL commit, so that is one way to determine what key was used for what. For server-generated entries, it may well decide to add custom metadata (#1512). 

I'm not sure what co-ownership of RSL signing would look like though. That could be tricky to achieve when considering FF and sequential entries.

docs/gaps/2/README.md
pjbgf: I'd say they are all valid choices. This is something to be explored on its own I think, as ideally the `gittuf` CLI would enable/support this. It could even be some sort of `known_host` but per forge/repo instead. But that is likely a GAP in itself.

docs/gaps/2/README.md
pjbgf: This configuration assumes that the forge is sole responsible for creating/signing RSL entries. However, there could be a world where a forge supports user-provided RSL entries only when it is removed from the policy. But this increases the complexity as decisions such as what happens when pushes take place without a respective RSL entry, and the server can create one. Potentially one to be called out as per of [this](https://github.com/gittuf/gittuf/pull/1523#issuecomment-5282049920).

pjbgf:
> Feel free to correct me, but from my understanding, the references it hides from advertisement are those that the forge does not include in what it shows clients during a git fetch. So, when a client would run git fetch, the forge would only send a list of available refs, essentially "advertising" them. Refs that are internal to the forge are excluded from that list, meaning clients can't see or fetch them.

@PratGit1606 that's correct. And the recommendation here is largely to ensure that the forge doesn't leak the existence of refs that were not really meant to exist (from a user perspective).

docs/gaps/2/README.md
```diff
@@ -6,8 +6,8 @@
 * **Title:** gittuf on the Forge
```
pjbgf, line 6: As part of this GAP, we should consider supporting Symbolic Refs RSL entries. 

The key motivation being that a forge should be able to declare as a RSL entry when `HEAD` changes. Without that a repo admin could change what users fetch from the default branch by temporarily pointing `HEAD` (default branch) to another branch. Ideally, the forge should be able convey the impact on the default branch of the repository from a RSL perspective. 
To some extent, it would even make sense for that to be its first entry.

This is less of a problem on `gittuf` on the client-side, as `HEAD` only affects the user itself.

adityasaky:
Discussed in the community meeting on Friday. We'll repurpose this GAP to setup a living deployment considerations document. Instead of multiple concrete configurations, the considerations will be structured as a series of decisions that forge integrators must make.

PR #1532, opened 2026-08-15
test :- add unit tests for policy lifecycle screen and forms

Aaravanand00:
## Description

This PR introduces comprehensive unit tests for the Policy Lifecycle screen and forms in `internal/cmd/tui/screen_policy_lifecycle_test.go`. It tests initializations, form submissions across lifecycle actions (Initialize Policy, Increment Version, Sign Policy, Stage Changes, Apply Changes, Pull Policy, Push Policy), warning banners, focus cycling, read-only mode protections, and view rendering.

## AI Usage

  request.

Used AI to assist in drafting unit test cases for policy lifecycle forms, read-only command protections and backendExec interface.

## Contributor Checklist

  request.
  applicable.

### DIFF
--- a/internal/cmd/tui/screen_policy_lifecycle_test.go
+++ b/internal/cmd/tui/screen_policy_lifecycle_test.go
@@ -5,6 +5,7 @@ package tui
 
 import (
 	"context"
+	"errors"
 	"strings"
 	"testing"
 
@@ -114,9 +115,12 @@ func TestPolicyLifecycleScreenFormSubmissions(t *testing.T) {
 			m.screen = screenPolicyLifecycleForm
 
 			s.initInputs(act, &m)
-			s.Update(tea.KeyMsg{Type: tea.KeyEnter}, &m)
+			_, cmd := s.Update(tea.KeyMsg{Type: tea.KeyEnter}, &m)
 			if m.screen != screenPolicyLifecycle {
 				t.Errorf("expected screen to return to screenPolicyLifecycle, got %v", m.screen)
+			}
+			if cmd == nil {
+				t.Error("expected non-nil cmd returned for lifecycle command submission")
 			}
 		})
 	}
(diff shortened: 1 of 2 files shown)


internal/cmd/tui/screen_policy_lifecycle_test.go
```diff
@@ -0,0 +1,200 @@
...
+				p:          &persistent.Options{SigningKey: "dummy-key"},
+			}
+
+			m := initialModel(context.Background(), o)
+			s := &m.policyLifecycleScreen
+			m.screen = screenPolicyLifecycleForm
+
+			s.initInputs(act, &m)
+			s.Update(tea.KeyMsg{Type: tea.KeyEnter}, &m)
+			if m.screen != screenPolicyLifecycle {
```
PratGit1606, line 119: The screen is set to `screenPolicyLifecycle` unconditionally before the command dispatches, and the test discards the return value so the operation never actually runs. All seven subtests pass identically regardless of whether the underlying git operation succeeds or fails. I think what would benefit would be having an assertion on the returned cmd to give this test real value. Also, ESC from the form doesn't navigate back despite the View telling the user "Esc to go back", which I saw none of the tests catch.

internal/cmd/tui/screen_policy_lifecycle_test.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func TestBackendExecInterface(t *testing.T) {
+	be := backendExec{run: func() error { return nil }}
+	if err := be.Run(); err != nil {
+		t.Errorf("expected nil error from backendExec.Run(), got %v", err)
+	}
+
+	// Cover setters
+	be.SetStdin(nil)
```
PratGit1606, line 193: `SetStdin`, `SetStdout`, `SetStderr` have empty bodies, so calling them with nil and asserting nothing would just be coverage padding. Rn, the only behavior backendExec actually has is Run(), adding a case that injects a failing function and asserts the error propagates would make it more thorough!

internal/cmd/tui/screen_policy_lifecycle_test.go
Aaravanand00: I've updated screen_policy_lifecycle.go to handle esc key navigation from screenPolicyLifecycleForm back to the lifecycle menu. Also updated TestPolicyLifecycleScreenFormSubmissions to assert on the returned tea.Cmd and added a test case for esc navigation from the form screen.

internal/cmd/tui/screen_policy_lifecycle_test.go
Aaravanand00: Added a test case in TestBackendExecInterface that injects a failing execution function and verifies that the error correctly propagates through backendExec.Run().

patzielinski:
Closing as these PRs will be combined as appropriate.

PR #1551, opened 2026-08-24
gitinterface: resolve worktrees for detached git dirs and linked worktrees

### LINKED ISSUE #1006: Fix worktree path resolution for repositories with detached git directories

vee1e:
## Description

gittuf figured out where a repository's checked-out files live by chopping ".git" off the end of its internal Git directory path. That only works for the most basic layout, and three common setups break because of it:

- Repositories created with `--separate-git-dir`, which store their history somewhere else entirely. gittuf ends up looking in a wrong or nonexistent directory.
- Linked worktrees created with `git worktree add`. These were misread as bare repositories (repos with no checked-out files at all), so file restoration was silently skipped. They could not even be loaded, because their configuration lives in the shared directory, not the worktree's own.
- Bare repositories named like `project.git`, misclassified for the same name-based reason.

This change introduces one function that answers "where is the worktree" for every layout, by asking Git itself rather than guessing from names:

1. For linked worktrees, read the pointer file Git writes inside the Git directory.
2. Otherwise honor an explicit worktree setting in the repository config, if one exists.
3. Otherwise use the location discovered while opening the repository, which is the only reliable source for separate-git-dir setups.
4. Otherwise fall back to the standard layout.

Resolved paths are sanity-checked, so a stale or malformed pointer fails safely instead of sending later commands into wrong directories. The bare-repository check now asks Git directly instead of checking whether the path ends in ".git", which was wrong in both directions. Errors surface as a sentinel error callers can detect, rather than Git failing inside some unrelated folder.

One tradeoff worth flagging: checking bareness now runs a quick Git command each time instead of a string comparison. I left it uncached on purpose, to keep the repository object free of shared mutable state.

Fixes #1006

## Testing

- New tests cover every layout above, plus SHA-256 variants, running from a subdirectory, stale pointer records, relative config values as submodules write them, and status inside a linked worktree.
- Full test suite passes with the race detector enabled.
- gofmt and go vet are clean.
- Each resolution branch was verified against real Git behavior on macOS.

## AI Usage

  request.

I used AI to research the issue and to draft the fix and its tests. All code was manually reviewed, run locally, and revised based on two independent adversarial review passes before I submitted it.

## Contributor Checklist

### DIFF
--- a/pkg/gitinterface/repository.go
+++ b/pkg/gitinterface/repository.go
@@ -84,7 +84,10 @@ func (r *Repository) ensureNoCompatObjectFormat() error {
 			if !filepath.IsAbs(commonDirPath) {
 				commonDirPath = filepath.Join(r.gitDirPath, commonDirPath)
 			}
-			configPath = filepath.Join(commonDirPath, "config")
+			commonConfigPath := filepath.Join(commonDirPath, "config")
+			if fileInfo, err := os.Stat(commonConfigPath); err == nil && fileInfo.Mode().IsRegular() {
+				configPath = commonConfigPath
+			}
 		}
 	}
 
@@ -199,13 +202,12 @@ func (r *Repository) GetGitDir() string {
 
 // IsBare returns true if the repository is a bare repository, i.e. it has no
 // worktree. Bareness is determined by Git itself; if Git cannot be consulted,
-// we fall back to inspecting the GIT_DIR name.
+// we fall back to checking for the sentinel files Git writes on bare
+// repositories.
 func (r *Repository) IsBare() bool {
 	stdOut, err := r.executor("rev-parse", "--is-bare-repository").executeString()
 	if err != nil {
-		// TODO: this may not work when the repo is cloned with GIT_DIR set
-		// elsewhere. We don't support this at the moment, so it's probably okay?
-		return !strings.HasSuffix(r.gitDirPath, ".git")
+		return isBareGitDir(r.gitDirPath)
 	}
 	return stdOut == "true"
 }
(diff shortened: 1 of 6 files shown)


cbrkrtek:
Hey! 
I noticed сopilot's review keeps failing with a quota error, and this has been open a month without a human review, so I'm flagging, because it looks solid to me.I went through GetWorktree() and worktree_test.go in detail. The bidirectional gitDirBelongsTo check does a good job closing off the stale/hostile-pointer redirect class of bug, and the 15 test scenarios cover pretty much every adversarial case I'd have asked about myself (stale core.worktree, a gitdir file pointing at another repo, .git-suffixed bare repos, etc.).One thing worth flagging: this branched off main about a month ago, and main has moved quite a bit since then (Lua sandbox, hooks, etc.), so it'll probably need a rebase once someone's ready to merge it.Happy to help test this against real worktree / --separate-git-dir setups, or help out with the rebase if that'd be useful.

PR #1570, opened 2026-09-05
Fix/1282 first entry violation recovery

### LINKED ISSUE #1282: Policy recovery fails when first RSL ref entry is invalid and skipped
When the first RSL entry for a ref is invalid and skipped, `VerifyRelativeForRef` fails because it expects a prior valid reference entry.

## Current behavior

- returns `rsl.ErrRSLEntryNotFound`
- because it calls `GetLatestReferenceUpdaterEntry` with `BeforeEntryID` on the invalid first entry and finds no earlier ref state

## Expected behavior

- handle the first-entry invalid case explicitly
- either accept the later valid state as recoverable, or return a clear recovery-specific error

### I realised that :
- The branch becomes unrecoverable under normal recovery logic.
- This is especially important for supply-chain workflows where the first commit may be rejected and then skipped.

ravindra-RKB:
## Description

This PR fixes the first-entry violation recovery in policy verification:
- **`verify.go`**: Updates the `isFirstEntryViolation` check's `else` block to return `verificationErr` rather than leaking the storage-layer `rsl.ErrRSLEntryNotFound`. Fixes the `verifyEntry` and `getCommits` logic to signal when recovering from a first-entry violation rather than anchoring on the skipped entry, ensuring unauthorized content from the bad first entry is properly detected.
- **`verify_test.go`**: Duplicates the four new first-entry violation recovery subtests into the `TestVerifyRelativeForRef` suite (using `createTestStateWithPolicy` and key-based principals), mirroring the structure in `TestVerifyRelativeForRefUsingPersons`.

Fixes #1282.

## AI Usage

- [x] I **did not** use generative AI at all in making the content of this pull request.
- [ ] I **did** use generative AI in some form in making the content of this pull request. I have described my use of AI below.

AI was used to draft the fixes for the code review feedback, identify the correct parameters for `verifyEntry`, and duplicate the `TestVerifyRelativeForRef` test cases into the general suite. All code has been manually reviewed, tested (`go test`), and verified locally.

## Contributor Checklist


internal/policy/verify.go
```diff
@@ -631,29 +631,41 @@ func (v *PolicyVerifier) VerifyRelativeForRef(ctx context.Context, firstEntry, l
...
-		slog.Debug("Verifying identified last valid entry has not been revoked...")
-		if lastGoodEntry.(*rsl.ReferenceEntry).SkippedBy(lastGoodEntryAnnotations) {
-			// this type assertion is fine because we use the rsl.IsReferenceEntry opt
-			return ErrLastGoodEntryIsSkipped
+			if errors.Is(err, rsl.ErrRSLEntryNotFound) {
+				firstEntryForRef, _, errFirst := rsl.GetFirstReferenceUpdaterEntryForRef(v.repo, invalidEntry.GetRefName())
+				if errFirst == nil && firstEntryForRef.GetID().Equal(invalidEntry.GetID().Bytes()) {
+					isFirstEntryViolation = true
+				} else {
+					return err
```
PratGit1606, line 641: This should return `verificationErr` rather than `err`. `err` here is rsl.ErrRSLEntryNotFound, which is a storage-layer sentinel leaking out of the policy layer. The "no fix found" case at line 745 already returns `verificationErr`, so this should match that convention. verificationErr is guaranteed non-nil here since it's set alongside invalidEntry, which is the precondition for reaching this block so having `err` just turns this into the same as returning the else

internal/policy/verify.go
```diff
@@ -680,13 +692,27 @@ func (v *PolicyVerifier) VerifyRelativeForRef(ctx context.Context, firstEntry, l
 				continue
 
 			case *rsl.ReferenceEntry:
-				newCommitTreeID, err := v.repo.GetCommitTreeID(newEntry.GetTargetID())
-				if err != nil {
-					return err
+				isFix := false
+				if isFirstEntryViolation {
+					slog.Debug("Checking if potential fix entry meets policy as there is no last good state to compare tree against...")
+					if err := verifyEntry(ctx, v.repo, currentPolicy, currentAttestations, newEntry); err == nil {
```
PratGit1606, line 698: `verifyEntry` internally anchors its commit range on the skipped invalid entry, so every commit that entry introduced gets excluded from file rule checking. What that means is that if the fix commit is a child of the bad first commit, and so `verifyEntry` only sees the delta between the bad commit and the fix, not the bad commit itself. 

So unauthorized content at a protected path that the first entry introduced would just pass through undetected. The fix here would be to signal that path when `isFirstEntryViolation` is true rather than anchoring on the skipped entry. Lmk if that didn't make sense

internal/policy/verify.go
```diff
@@ -704,7 +730,7 @@ func (v *PolicyVerifier) VerifyRelativeForRef(ctx context.Context, firstEntry, l
 					break lookForFixes
 				}
 
-				// newEntry is not tree-same / commit-same, so it is automatically
+				// newEntry is not tree-same / commit-same / verifyable, so it is automatically
```
PratGit1606, line 733: typo: should be "verifiable"

internal/policy/verify_test.go
```diff
@@ -757,6 +757,166 @@ func TestVerifyRelativeForRefUsingPersons(t *testing.T) {
...
+		// Skip the invalid entry — but add NO fix entry
+		annotation := rsl.NewAnnotationEntry([]githash.Hash{invalidEntryID}, true, "first entry was invalid")
+		annotationID := common.CreateTestRSLAnnotationEntryCommit(t, repo, annotation, gpgKeyBytes)
+		annotation.ID = annotationID
+
+		// Verification must fail: invalid entry was skipped but no fix exists.
+		verifier := NewPolicyVerifier(repo)
+		err = verifier.VerifyRelativeForRef(testCtx, firstEntry, invalidEntry, refName)
+		assert.ErrorIs(t, err, ErrVerificationFailed)
+	})
```
PratGit1606, line 919: These four subtests are only added to `TestVerifyRelativeForRefUsingPersons` but there's a parallel TestVerifyRelativeForRef that mirrors all existing subtests 1:1, and that's the convention this file follows throughout. These four cases need to be duplicated there as well using createTestStateWithPolicy and key based principals.

PratGit1606 (changes requested):

PratGit1606:
Thanks for putting this together! I added feedback for a bunch of changes, Just a heads up, the AI disclosure and PR description don't match the format we normally follow for gittuf PRs, please go through the contributor checklist format and rewrite it accordingly. Also please squash the commits down to 1

ravindra-RKB:
@PratGit1606 ptal all looks good or anything needed from my side...

shreyaabaranwal:
Thanks for tackling this , the core fix is right, and importantly it avoids the tree-sameness bypass: for the first-entry-violation case you correctly fall back to a full `verifyEntry` check on each candidate instead of comparing trees against a (nonexistent) last-good state, which is what actually closes the "push a bad first entry, skip it, then trust an arbitrary unauthorized fix" hole.

One blocking issue: this doesn't currently compile. `verifyEntry`'s signature gained the `isFirstEntryViolation bool` param, but two call sites in `verify_test.go` weren't updated (lines 4116 and 4132, in the "both global and policy rule declared..." subtest of `TestVerifyEntry`) , `go build ./...` / `go test ./internal/policy/...` fails with `not enough arguments in call to verifyEntry`. CI hasn't flagged this yet because the workflow runs on this PR are all stuck at `action_required` pending approval. Once I added the missing `false` arg to those two calls locally, the full suite passes (including all the new recovery subtests), so this looks like a one-line-per-callsite fix away from green.

Nit: could you push that fix so CI can actually run and confirm?

ravindra-RKB:
@patzielinski ptal...

patzielinski:
@ravindra-RKB This is something that needs a few pairs of eyes on, so we'll get a review on this hopefully soon. In the meantime don't worry about rebasing this PR, we can do that later.

PR #1580, opened 2026-09-11
test: enable GPG key tests on Windows

### LINKED ISSUE #1579: Enable GPG key tests on Windows (currently unconditionally skipped)
### What happened?

`SetupTestGPGHomeDir` (`internal/signerverifier/gpg/gpg.go`) unconditionally skips on Windows:

```go
func SetupTestGPGHomeDir(t *testing.T, privateKeyBytes ...[]byte) {
	if runtime.GOOS == "windows" {
		t.Skip("TODO: test gpg keys for metadata on Windows")
	}
	...
```

This disables `TestGPG` in this package as well as the GPG-key-signing paths exercised via this helper in `experimental/gittuf/verify_test.go`, `experimental/gittuf/root_test.go`, and `experimental/gittuf/keys_test.go` on every Windows CI run.

I dug into why this was skipped and reproduced it locally on Windows:
- With Git for Windows' bundled MSYS `gpg` (the one found first on PATH inside a Git Bash / MSYS shell), `GNUPGHOME` set to a native Windows path (`C:\Users\...`, which is what `os.MkdirTemp` returns on Windows) is misinterpreted as a *relative* path, since MSYS gpg only treats POSIX-style paths (starting with `/`) as absolute. This causes `gpg: no writable keyring found: Not found`.
- However, `go test` in CI runs under `shell: powershell` (the default for `windows-latest` runners), which resolves `gpg` to whatever is first on the native Windows `PATH`, not MSYS's `usr/bin`. With a native GnuPG (Gpg4win) install, the same Windows-style `GNUPGHOME` path works correctly, and the full `TestGPG` round-trip (import, sign, verify) passes.

So the actual blocker isn't the code in `gpg.go`. It's that GnuPG isn't installed on the `windows-latest` GitHub Actions runner image (confirmed via the runner-images software manifest), so `gpg` isn't found in `PATH` for the `Test` step in `.github/workflows/ci.yml` at all.

**Expected:** GPG key signing/verification tests should run on Windows CI like they do on Linux/macOS, instead of being silently skipped.

Proposed fix:
1. Add a step to `.github/workflows/ci.yml` to install GnuPG on the `windows-latest` runner (e.g. via `choco install gnupg`) before the `Test` step.
2. Remove the `runtime.GOOS == "windows"` skip in `SetupTestGPGHomeDir`.

I'll open a PR shortly with this change verified against a real Windows Actions run.

### Relevant log output

```
gpg: no writable keyring found: Not found
gpg: no writable keyring found: Not found
gpg: key DCFE043CDD2DB96E: failed to re-lookup public key: No public key
```

### Code of Conduct

bhuvan-somisetty:
## Description

`SetupTestGPGHomeDir` (`internal/signerverifier/gpg/gpg.go`) unconditionally skipped on Windows, which disabled `TestGPG` in this package as well as the GPG-signing paths exercised via this helper in `experimental/gittuf/verify_test.go`, `experimental/gittuf/root_test.go`, and `experimental/gittuf/keys_test.go` on every Windows CI run.

## Root cause

GnuPG isn't installed on the `windows-latest` GitHub Actions runner image, so `gpg` wasn't found on `PATH` for the `Test` step at all.

Installing it via `choco install gnupg` (my first attempt) turned out to be unreliable: the actual CI run got stuck on package installation and timed out after 45 minutes.

Git for Windows (always present on the runner) already ships its own MSYS build of `gpg` under `Git\usr\bin`. Using that avoids the network install entirely, but surfaces a second, more subtle issue: that MSYS build only recognizes POSIX-style absolute paths (e.g. `/c/Users/...`). Given a native Windows path (what `os.MkdirTemp` returns on Windows, e.g. `C:\Users\...`), it doesn't treat it as absolute and silently resolves it relative to gpg's working directory instead, so `GNUPGHOME` has to be rewritten to POSIX form for that build specifically.

Rewriting it unconditionally on Windows isn't correct either: a native (non-MSYS) GnuPG install, such as Gpg4win, expects the native Windows path as-is and fails to find the same directory if it's rewritten. I hit this directly testing locally, since I have Gpg4win installed and it resolves ahead of Git's bundled build on my machine's `PATH`.

## Fix

1. Add Git's `usr\bin` to `PATH` on the Windows CI runner, appended via `GITHUB_ENV` rather than prepended via `GITHUB_PATH`, so it only fills in `gpg` (otherwise missing) without shadowing tools like `ssh-keygen` that already resolve to the Windows OpenSSH build earlier in `PATH`.
2. In `SetupTestGPGHomeDir`, detect via `exec.LookPath("gpg")` whether the resolved binary is an MSYS build (Git's `usr\bin` or `mingw64\bin`) and only rewrite `GNUPGHOME` to POSIX form in that case, leaving native installs untouched.
3. Remove the `runtime.GOOS == "windows"` skip.

## Testing

- `go build ./...`: passes
- `go vet ./...`: passes
- `go test ./internal/signerverifier/gpg/...`: passes on Windows under both configurations I could reproduce locally:
  - Native GnuPG (Gpg4win) resolved first on `PATH` (my machine's default)
  - MSYS-only `PATH` (Git's `usr\bin` only, no native GnuPG), matching what the CI runner will actually have
- Added `TestMsysPath` and `TestIsMSYSGPGPath` covering the path-rewriting and build-detection logic directly, including the case that caught the bug above (a native, non-MSYS `gpg` on `PATH`).
- Confirmed this doesn't mask a pre-existing, unrelated Windows test failure: `TestVerifyRef` et al. in `experimental/gittuf` fail locally on unmodified `main` too (an `ssh-keygen -Y sign` environment issue, unrelated to GPG), so this PR doesn't touch or regress that.
- The first push of this PR already confirmed real Windows CI hits the `choco` hang described above; this push replaces that approach.

Fixes #1579

## AI Usage

- [x] I did use generative AI in some form in making the content of this pull request. I have described my use of AI below.

Used Claude Code to investigate the root cause end to end (reproducing the `choco install gnupg` CI hang, then the MSYS-vs-native `gpg` path-handling difference), and to draft the code and CI changes. I reviewed and tested the change locally on Windows under both PATH configurations before pushing.

## Contributor Checklist

- [x] I have manually reviewed all content submitted to gittuf in this pull request.
- [x] All commits in this pull request include a DCO Signoff.
- [x] By submitting this pull request, I agree to follow the gittuf Code of Conduct.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,14 @@ jobs:
       with:
         go-version: ${{ matrix.go-version }}
         cache: true
+    - name: Use Git for Windows' bundled gpg (Windows)
+      if: runner.os == 'Windows'
+      shell: powershell
+      # Appended to the end of PATH (via GITHUB_ENV) rather than the front
+      # (via GITHUB_PATH), so this only fills in gpg, which isn't on PATH
+      # otherwise, without shadowing tools like ssh-keygen that already
+      # resolve to the Windows OpenSSH build earlier in PATH.
+      run: echo "PATH=$env:PATH;C:\Program Files\Git\usr\bin" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
     - name: Test
       run: go test -race -timeout 40m ./...
   test-freebsd:
--- a/internal/signerverifier/gpg/gpg.go
+++ b/internal/signerverifier/gpg/gpg.go
@@ -155,13 +155,57 @@ func getPublicKeyForKeyID(keyID, program string) (*signerverifier.SSLibKey, erro
 	return LoadGPGKeyFromBytes(output)
 }
 
+// msysPath converts a native Windows path (e.g. `C:\Users\foo`) into the
+// POSIX-style form MSYS tools such as Git for Windows' bundled gpg expect
+// for absolute paths (e.g. `/c/Users/foo`). See the MSYS2 documentation on
+// path conversion: https://www.msys2.org/docs/filesystem-paths/
+//
+// This intentionally doesn't use filepath.ToSlash: that only rewrites the
+// separator native to the OS the code is compiled for, so on a non-Windows
+// host (as when this package's tests run on Linux/macOS CI runners) it
+// leaves Windows-style backslashes untouched instead of converting them.
+func msysPath(path string) string {
+	path = strings.ReplaceAll(path, `\`, "/")
+	if len(path) >= 2 && path[1] == ':' {
+		path = "/" + strings.ToLower(path[:1]) + path[2:]
+	}
+	return path
+}
+
+// gpgIsMSYSBuild reports whether the `gpg` binary that will actually be
+// invoked is an MSYS build, such as the one bundled with Git for Windows.
+// This matters because MSYS gpg only recognizes POSIX-style absolute paths
+// (e.g. `/c/Users/...`): given a native Windows path, it doesn't treat it as
+// absolute and silently resolves it relative to gpg's working directory
+// instead. A non-MSYS (e.g. Gpg4win) install, on the other hand, expects the
+// native Windows path as-is, so callers must not assume one or the other
+// merely from being on Windows.
+func gpgIsMSYSBuild() bool {
+	path, err := exec.LookPath("gpg")
+	if err != nil {
+		return false
+	}
+
+	return isMSYSGPGPath(path)
+}
+
+// isMSYSGPGPath reports whether path (as resolved by exec.LookPath("gpg"))
+// points at an MSYS build of gpg, such as the one bundled with Git for
+// Windows under `Git\usr\bin` or `Git\mingw64\bin`. See
+// https://www.msys2.org/docs/environments/ for background on the usr/bin
+// vs. mingw64/bin MSYS2 environments Git for Windows is built from.
+//
+// As in msysPath, this doesn't use filepath.ToSlash, since that would only
+// normalize separators when compiled for Windows and this needs to work
+// when the package's tests run on any OS.
+func isMSYSGPGPath(path string) bool {
+	path = strings.ToLower(strings.ReplaceAll(path, `\`, "/"))
+	return strings.Contains(path, "/git/usr/bin/gpg") || strings.Contains(path, "/git/mingw64/bin/gpg")
+}
+
 // SetupTestGPGHomeDir is a test helper used only to prepare a temporary GPG
 // home dir with the specified keys added in.
 func SetupTestGPGHomeDir(t *testing.T, privateKeyBytes ...[]byte) {
-	if runtime.GOOS == "windows" {
-		t.Skip("TODO: test gpg keys for metadata on Windows")
-	}
-
 	// We use os.MkdirTemp because t.TempDir can result in a path
 	// that's too long for socket files, used for gpg-agent.
 	tmpGpgHomeDir, err := os.MkdirTemp("", "gittuf-gpg-")
@@ -170,7 +214,11 @@ func SetupTestGPGHomeDir(t *testing.T, privateKeyBytes ...[]byte) {
 		os.RemoveAll(tmpGpgHomeDir) //nolint:errcheck
 	})
 
-	t.Setenv("GNUPGHOME", tmpGpgHomeDir)
+	gnupgHome := tmpGpgHomeDir
+	if runtime.GOOS == "windows" && gpgIsMSYSBuild() {
+		gnupgHome = msysPath(tmpGpgHomeDir)
+	}
+	t.Setenv("GNUPGHOME", gnupgHome)
 
 	gpgAgentConfPath := filepath.Join(tmpGpgHomeDir, "gpg-agent.conf")
 	if err := os.WriteFile(gpgAgentConfPath, artifacts.GPGAgentConf, 0o600); err != nil {
--- a/internal/signerverifier/gpg/gpg_test.go
+++ b/internal/signerverifier/gpg/gpg_test.go
@@ -90,6 +90,86 @@ func TestLoadGPGKeyFromBytes(t *testing.T) {
 	})
 }
 
+func TestMSYSPath(t *testing.T) {
+	tests := []struct {
+		name string
+		path string
+		want string
+	}{
+		{
+			name: "Windows path with backslashes",
+			path: `C:\Users\foo\AppData\Local\Temp\gittuf-gpg-123`,
+			want: "/c/Users/foo/AppData/Local/Temp/gittuf-gpg-123",
+		},
+		{
+			name: "Windows path with forward slashes",
+			path: "C:/Users/foo/AppData/Local/Temp/gittuf-gpg-123",
+			want: "/c/Users/foo/AppData/Local/Temp/gittuf-gpg-123",
+		},
+		{
+			name: "uppercase drive letter is lowercased",
+			path: `D:\gittuf-gpg-123`,
+			want: "/d/gittuf-gpg-123",
+		},
+		{
+			name: "already POSIX-style path is unchanged",
+			path: "/tmp/gittuf-gpg-123",
+			want: "/tmp/gittuf-gpg-123",
+		},
+		{
+			name: "relative path without a drive letter is unchanged",
+			path: "gittuf-gpg-123",
+			want: "gittuf-gpg-123",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			assert.Equal(t, test.want, msysPath(test.path))
+		})
+	}
+}
+
+func TestIsMSYSGPGPath(t *testing.T) {
+	tests := []struct {
+		name string
+		path string
+		want bool
+	}{
+		{
+			name: "Git for Windows usr/bin build",
+			path: `C:\Program Files\Git\usr\bin\gpg.exe`,
+			want: true,
+		},
+		{
+			name: "Git for Windows mingw64 build",
+			path: `C:\Program Files\Git\mingw64\bin\gpg.exe`,
+			want: true,
+		},
+		{
+			name: "native Gpg4win install",
+			path: `C:\Program Files\GnuPG\bin\gpg.exe`,
+			want: false,
+		},
+		{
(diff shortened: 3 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -36,6 +36,10 @@ jobs:
       with:
         go-version: ${{ matrix.go-version }}
         cache: true
+    - name: Install GnuPG (Windows)
+      if: runner.os == 'Windows'
+      shell: powershell
+      run: choco install gnupg --no-progress -y
```
patzielinski: Can this be done via `winget` instead?

patzielinski (commented):
Thanks for opening this, @bhuvan-somisetty.

As of writing the CI run is stuck on installing the package. See my inline comment on this.

Otherwise this does look to enable these tests in CI, but users on Windows machines running the test suite locally will likely fail unless they follow the same steps. We can handle this via a documentation update later.

.github/workflows/ci.yml
bhuvan-somisetty: Switched it up, using Git for Windows own bundled gpg now so there's no install step at all, should fix the CI hang too.

bhuvan-somisetty:
Pushed a couple of fixes based on the review feedback and CI failures:

- Removed the choco install for GnuPG (it was hanging on package install in CI) and switched to using Git for Windows' own bundled gpg instead, no install step needed.
- Fixed msysPath so it converts backslashes correctly on any OS running the test (the previous version only worked when the test itself ran on Windows).
- Only rewrite GNUPGHOME to POSIX form when gpg actually resolves to Git's MSYS build, so a native GnuPG install (e.g. Gpg4win) still works if present.
- Appended Git's usr/bin to PATH instead of prepending it, so it doesn't shadow the Windows OpenSSH ssh-keygen that other tests depend on.

Since this touches .github/workflows/ci.yml, the workflow runs are sitting in "action_required" and need a maintainer to approve them to run.

bhuvan-somisetty:
Pushed a small fix, all checks pass locally now. Runs are just waiting on approval to actually run, could someone approve them when free

.github/workflows/ci.yml
```diff
@@ -36,6 +36,14 @@ jobs:
...
         go-version: ${{ matrix.go-version }}
         cache: true
+    - name: Use Git for Windows' bundled gpg (Windows)
+      if: runner.os == 'Windows'
+      shell: powershell
+      # Appended to the end of PATH (via GITHUB_ENV) rather than the front
+      # (via GITHUB_PATH), so this only fills in gpg, which isn't on PATH
+      # otherwise, without shadowing tools like ssh-keygen that already
+      # resolve to the Windows OpenSSH build earlier in PATH.
+      run: echo "PATH=$env:PATH;C:\Program Files\Git\usr\bin" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
```
patzielinski: Can you substantiate why this is being done via `GITHUB_ENV` and not just via updating the `PATH`?

internal/signerverifier/gpg/gpg.go
```diff
@@ -155,13 +155,45 @@ func getPublicKeyForKeyID(keyID, program string) (*signerverifier.SSLibKey, erro
...
+// for absolute paths (e.g. `/c/Users/foo`).
+func msysPath(path string) string {
+	path = strings.ReplaceAll(path, `\`, "/")
+	if len(path) >= 2 && path[1] == ':' {
+		path = "/" + strings.ToLower(path[:1]) + path[2:]
+	}
+	return path
+}
+
+// gpgIsMSYSBuild reports whether the `gpg` binary that will actually be
```
patzielinski, line 169: Can you add a link to documentation in this comment that describes this behavior? It'd be good to have for both justifying this and for future reference.

internal/signerverifier/gpg/gpg_test.go
```diff
@@ -90,6 +90,86 @@ func TestLoadGPGKeyFromBytes(t *testing.T) {
 	})
 }
 
+func TestMsysPath(t *testing.T) {
```
patzielinski, line 93:
```suggestion
func TestMSYSPath(t *testing.T) {
```

internal/signerverifier/gpg/gpg.go
```diff
@@ -155,13 +155,45 @@ func getPublicKeyForKeyID(keyID, program string) (*signerverifier.SSLibKey, erro
...
+	if err != nil {
+		return false
+	}
+
+	return isMSYSGPGPath(path)
+}
+
+// isMSYSGPGPath reports whether path (as resolved by exec.LookPath("gpg"))
+// points at an MSYS build of gpg, such as the one bundled with Git for
+// Windows under `Git\usr\bin` or `Git\mingw64\bin`.
```
patzielinski, line 188: Same here with a link to some docs.

internal/signerverifier/gpg/gpg_test.go
patzielinski, line 1: Please also use the correct capitalization in the file of OSes and similar (e.g. `Windows`, `macOS`, `Homebrew`, etc.)

internal/signerverifier/gpg/gpg.go
```diff
@@ -155,13 +155,45 @@ func getPublicKeyForKeyID(keyID, program string) (*signerverifier.SSLibKey, erro
 	return LoadGPGKeyFromBytes(output)
 }
 
+// msysPath converts a native Windows path (e.g. `C:\Users\foo`) into the
+// POSIX-style form MSYS tools such as Git for Windows' bundled gpg expect
+// for absolute paths (e.g. `/c/Users/foo`).
+func msysPath(path string) string {
```
patzielinski, line 161: Can this use https://pkg.go.dev/path/filepath#ToSlash?

internal/signerverifier/gpg/gpg.go
```diff
@@ -155,13 +155,45 @@ func getPublicKeyForKeyID(keyID, program string) (*signerverifier.SSLibKey, erro
...
+	}
+
+	return isMSYSGPGPath(path)
+}
+
+// isMSYSGPGPath reports whether path (as resolved by exec.LookPath("gpg"))
+// points at an MSYS build of gpg, such as the one bundled with Git for
+// Windows under `Git\usr\bin` or `Git\mingw64\bin`.
+func isMSYSGPGPath(path string) bool {
+	path = strings.ToLower(strings.ReplaceAll(path, `\`, "/"))
```
patzielinski, line 190: Same here with https://pkg.go.dev/path/filepath#ToSlash.

internal/signerverifier/gpg/gpg_test.go
```diff
@@ -90,6 +90,86 @@ func TestLoadGPGKeyFromBytes(t *testing.T) {
...
+	tests := []struct {
+		name string
+		in   string
```
patzielinski, line 96:
```suggestion
		path   string
```

internal/signerverifier/gpg/gpg_test.go
```diff
@@ -90,6 +90,86 @@ func TestLoadGPGKeyFromBytes(t *testing.T) {
...
+		t.Run(test.name, func(t *testing.T) {
+			assert.Equal(t, test.want, msysPath(test.in))
+		})
+	}
+}
+
+func TestIsMSYSGPGPath(t *testing.T) {
+	tests := []struct {
+		name string
+		in   string
```
patzielinski, line 136:
```suggestion
		path   string
```

patzielinski (commented):
A few items on docs and style nits. Thanks!

internal/signerverifier/gpg/gpg.go
bhuvan-somisetty: Can't quite, ToSlash only swaps the separator for whatever OS it's compiled on, so it stays a no-op on Linux and macOS and the cross platform tests break. Went with a plain backslash replace instead and left a comment explaining why.

internal/signerverifier/gpg/gpg.go
bhuvan-somisetty: Same reasoning here, ToSlash wouldn't convert the backslashes when running on Linux or macOS, so kept it explicit and added a note.

internal/signerverifier/gpg/gpg.go
bhuvan-somisetty: Added, thanks

internal/signerverifier/gpg/gpg.go
bhuvan-somisetty: Added, thanks

internal/signerverifier/gpg/gpg_test.go
bhuvan-somisetty: Renamed

internal/signerverifier/gpg/gpg_test.go
bhuvan-somisetty: Fixed the casing throughout

internal/signerverifier/gpg/gpg_test.go
bhuvan-somisetty: Done

internal/signerverifier/gpg/gpg_test.go
bhuvan-somisetty: Done

internal/signerverifier/gpg/gpg.go
patzielinski: Got it, thanks!

patzielinski (commented):
Looks good. Please squash commits and we'll get this merged, thanks!

bhuvan-somisetty:
Squashed, thanks for the review!

patzielinski (dismissed):

bhuvan-somisetty:
Had to force push again, an unrelated commit with tui changes ended up on this branch by mistake. Back to just the gpg windows fix now, sorry for the noise, could you take another look when you get a chance

patzielinski:
Yes, this was because I rebased the branch to include the latest commits from `main`, as GitHub requires this before merging. Please don't edit the branch once approved.

bhuvan-somisetty:
Ah got it, my bad, restored it back. Won't touch the branch again, thanks for sorting it out

patzielinski:
Thanks!

patzielinski (approved):

PR #1581, opened 2026-09-11
rsl: report unknown entry types with upgrade guidance

pjbgf:
## Description

An RSL entry whose header the client does not recognize used to fail with the generic invalid entry error, leaving the user unable to tell a corrupt entry from a repository written by a newer gittuf. Introduce ErrUnknownRSLEntryType, which wraps ErrInvalidRSLEntry so existing callers keep matching, and include the entry ID, a bounded prefix of the header and a request to upgrade in the message.

The CLI now recognizes this error and the existing unknown metadata schema version errors as upgrade required, and prints the running client version after them so a bug report or upgrade decision has the version at hand.

## AI Usage

  request.

Creating tests and formulating commit message.

## Contributor Checklist

  request.
  applicable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 This file tracks the changes introduced by gittuf versions.
 
+## Unreleased
+
+### Added
+
+- A clear upgrade message when the RSL contains an entry type or the policy
+  metadata uses a schema version this client does not implement
+
 ## v0.16.0
 
 This release adds support for SHA-256 Git repositories and reworks gittuf's
--- a/pkg/rsl/rsl.go
+++ b/pkg/rsl/rsl.go
@@ -54,6 +54,7 @@ var (
 	ErrInvalidGetLatestReferenceUpdaterEntryOptions = errors.New("invalid options presented for getting latest reference updater entry (are both before or until conditions set or is the before number less than the until number?)")
 	ErrCannotUseEntryNumberFilter                   = errors.New("current RSL entries are not numbered, cannot use number range options")
 	ErrInvalidUntilEntryNumberCondition             = errors.New("cannot meet until entry number condition")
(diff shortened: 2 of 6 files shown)


pkg/rsl/rsl.go
```diff
@@ -1126,7 +1139,7 @@ func parseRSLEntryText(id githash.Hash, text string) (Entry, error) {
 		}
 		return entry, nil
 	default:
-		return nil, ErrInvalidRSLEntry
+		return nil, newUnknownEntryTypeError(id, text)
```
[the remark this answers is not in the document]
pjbgf: Fair point, fixed on 01a8f7f.

pkg/rsl/rsl.go
```diff
@@ -1152,40 +1103,51 @@ func ParseEntryText(id githash.Hash, text string) (Entry, error) {
 	return parseRSLEntryText(id, text)
 }
 
+const unknownEntryTypeGuidance = "This repository may use a gittuf feature newer than this client. Upgrade gittuf to the latest release and retry. If this client is already the latest release, the entry may be corrupt or malicious and should be reported to the repository owners."
```
patzielinski, line 1106: Given the error handler in `cmd` matches on errors other than `ErrUnknownRSLEntryType`, can we move this out so this message prints out on the other two errors? It's generic enough it can apply to unknown root/targets metadata as well.

CHANGELOG.md
patzielinski, line 1: We can likely omit this change. We normally go through the list of new commits when releasing and build the changelog that way.

patzielinski (commented):
This reminds me of a discussion we had before on making error messages more user-friendly, as is done here with the upgrade explanation. We should look into if/how we can do this for other errors, but that's for another issue/PR.

CHANGELOG.md
patzielinski: On second thought, this is probably fine. We'll need to figure out what changes are large enough to need an update here in each PR, but we can deal with that as we go?

PR #1583, opened 2026-09-11
rsl: add bulk reference entries

pjbgf:
## Description

Record several reference updates under one signature, so a push that
updates multiple refs costs one RSL entry and one signing operation
instead of one per ref. Annotation entries gain optional `ref`
qualifiers, letting a skip target a single update inside a bulk entry.

Old clients fail closed on the new entry type: every release through
v0.16.0 rejects an RSL containing a bulk entry rather than misreading
it, so verifying a repository that has one requires upgrading to the
latest gittuf. `make test-compat` pins that behaviour against v0.8.1,
v0.14.1 and v0.16.0, along with the annotation `ref` qualifier, which
those releases parse and ignore.

Depends on #1581 to improve user experience for unknown types.

## AI Usage

  request.

Prototyping, wiring up and tests.

## Contributor Checklist

  request.
  applicable.

### DIFF
--- a/docs/design-document.md
+++ b/docs/design-document.md
@@ -362,6 +362,7 @@ RSL Bulk Reference Entry
 <ref 2>: <target ID 2>
 
 number: <number>
+custom.<namespace>/<name>: <value>
 ```
 
 The following rules are enforced by the parser, and therefore by every
@@ -378,6 +379,10 @@ verifier.
 * The `number` field is mandatory for this entry type, unlike for the other
   entry types where it is optional. It always follows exactly one blank line
   after the last update line.
+* Custom fields, described below, follow the `number` line. Unlike the other
+  entry types, which ignore keys they do not recognize, any other key after
+  the `number` line is rejected, so a client fails closed on a field it does
+  not implement.
 * The entry commit uses the empty tree, like every other RSL entry, and the
   signature on that commit covers all the updates the entry lists.
 
--- a/docs/design-document.md
+++ b/docs/design-document.md
@@ -330,6 +330,7 @@ entries. Annotations have the following schema.
 RSL Annotation Entry
 
 entryID: <RSL entry ID 1>
+ref: <ref name>
 entryID: <RSL entry ID 2>
 ...
 skip: <true/false>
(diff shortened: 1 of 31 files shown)


docs/gaps/7/README.md
patzielinski: Document needs an update to last modified date.

docs/design-document.md
patzielinski: Same here with last modified date.

docs/gaps/7/README.md
patzielinski, line 1: Looks like the document needs an update to last modified date.

docs/design-document.md
patzielinski, line 1: Same here with last modified date.

patzielinski, line 1: Also as discussed in the maintainer meet we may want to codify this as a GAP to at least get our thoughts down on things and discuss the security implications of these decisions.

docs/design-document.md
```diff
@@ -340,6 +341,67 @@ custom.<namespace>/<name>: <value>
...
+verifier.
+
+* The body is one line per reference update, in the order the updates were
+  recorded. The reference name is the key and the target ID is the value.
+  Git reference names cannot contain a colon, so splitting each line on its
+  first colon is unambiguous.
+* Reference names must be fully qualified and must be unique within one entry.
+  A duplicate reference is rejected.
+* No reference may be in the `refs/gittuf/` namespace. Policy,
+  policy-staging and attestations updates keep their own single entries.
```
patzielinski, line 378: Is there any case where having a bulk entry for regular refs could be problematic? I think this would be project-specific if so?

experimental/gittuf/rsl.go
```diff
@@ -129,6 +107,166 @@ func (r *Repository) RecordRSLEntryForReference(ctx context.Context, refName str
...
+	} else {
+		slog.Debug("Not checking if latest entry for each reference has same target")
+	}
+
+	if len(resolved) == 0 {
+		return nil
+	}
+
+	// TODO: once policy verification is in place, the signing key used by
+	// signCommit must be verified for each refName in the delegation tree.
```
patzielinski, line 201: I think this is a leftover comment from forever ago that's no longer applicable?

docs/design-document.md
```diff
@@ -340,6 +341,67 @@ custom.<namespace>/<name>: <value>
 ------END MESSAGE------
 ```
 
+The `ref` lines after an `entryID` are optional qualifiers. An entry ID with no
+qualifier is referred to as a whole, while an entry ID with qualifiers is
+referred to only for the listed reference updates. Qualifiers are only valid
+against RSL bulk reference entries, and they allow an annotation to skip one
+reference update inside a bulk entry while leaving the other updates in that
+entry in force.
+
```
patzielinski, line 350: Some cases with annotation entries that we should think about:

- Bulk entry updates ref A and B, annotation entry revokes update to ref A, and then another annotation entry revokes the whole bulk entry.
- Bulk entry updates ref A and B, annotation entry revokes the whole bulk ref. Further revokes of the bulk entry are presumed to be no-ops?
- Some entry is revoked, can the fix entry be a bulk entry?

internal/display/rsl.go
```diff
@@ -75,6 +75,21 @@ func RSLLog(repo gitstore.Storer, writer io.WriteCloser, opts ...Option) error {
 				// unexpectedly closed, such as by killing the pager
 				return nil
 			}
+		case *rsl.BulkReferenceEntry:
+			// The whole entry renders when any update matches the filter,
+			// because the entry is one signed act.
```
patzielinski, line 80: This comment might need a bit of clarification.

docs/design-document.md
```diff
@@ -340,6 +341,67 @@ custom.<namespace>/<name>: <value>
 ------END MESSAGE------
 ```
 
+The `ref` lines after an `entryID` are optional qualifiers. An entry ID with no
+qualifier is referred to as a whole, while an entry ID with qualifiers is
+referred to only for the listed reference updates. Qualifiers are only valid
+against RSL bulk reference entries, and they allow an annotation to skip one
+reference update inside a bulk entry while leaving the other updates in that
+entry in force.
```
adityasaky, line 349: If the annotation needs to skip multiple refs in the same bulk entry, should the entry ID be repeated with different ref values? Or can there be multiple ref lines after a single entryID line?

adityasaky: Worth clarifying: How should an annotation that skips a bulk entry by specifying a ref be handled if that ref is a mistake and is not part of the bulk entry?

docs/design-document.md
adityasaky, line 378: in the case a push touches both `refs/gittuf/*` and `refs/heads/...`, should there be a bulk entry for the `refs/heads/...` and one apiece for `refs/gittuf/*` refs? I think this begs a question similar to Robin's -- which should take precedence in ordering? As they're part of a single push, maybe `refs/gittuf/*` should and it's less ambiguous after all 🤔

pjbgf: Great question. I'm not sure we need to be prescriptive about this.

From push semantics / server-side handling, it may make sense to handle `refs/gittuf/*` first as its own bulk entry so it always take precedence. If that invalidates the rest of the pushed commands they could be partially rejected. This would result in deterministic outcomes when processing multi-command pushes.

But there may be edge cases around on-boarding an existing gittuf enabled repo.

pjbgf: I couldn't really come up with anything around this. Is there any specific aspect you are concerned about?

docs/design-document.md
pjbgf: Good shout, let me think about this a bit further and try to add tests to cover these cases.

docs/design-document.md
patzielinski: Not in particular, just thinking out loud here. This is probably fine.

PR #1588, opened 2026-09-12
test: add end-to-end integration tests for the gittuf transport

### LINKED ISSUE #1231: Integration Testing for the gittuf Transport
The [gittuf transport](https://github.com/gittuf/gittuf/tree/main/internal/git-remote-gittuf) runs gittuf operations when the user synchronizes the repository with the remote. We should add integration tests to the transport so that we can catch issues or regressions.

This should be a new action that simulates pushing and pulling to a repository (ideally a server contained within the action run) with both the `SSH` and `curl` handlers, and various states of the repository, e.g. the remote repository has gittuf metadata, but the local does not (and vice versa), divergent gittuf policies, etc.

yashrajshuklaaa:
Closes #1231

This adds end to end integration tests for git-remote-gittuf, covering both the curl (HTTP) and SSH handlers. Tests use real git push and git clone against a real HTTP server (git-http-backend wrapped in a small Go binary) and a real local sshd, not calling transport internals directly, as discussed in the issue.

## What's tested

- HTTP: push (main + all gittuf refs together via the wildcard refspec), clone syncing gittuf refs, standalone RSL ref push after the wildcard push
- SSH: same three scenarios, using a real sshd with AcceptEnv GIT_PROTOCOL set (needed this, otherwise the client's GIT_PROTOCOL=version=2 never reaches the server and it silently falls back to v0/v1)

## Structure

Two commits on purpose, per @patzielinski's suggestion in the issue:
1. The tester itself, no special casing for anything
2. A second commit that marks the known issues below as non-fatal, so the rest of the suite can still pass in CI

Plus a third commit for the GitHub Actions workflow.

## Known issues found while building this (not fixed here)

- HTTP clone does not sync refs/gittuf/* locally. Confirmed the remote itself is fine via git fsck, so this looks like a race in main.go's loop that waits for the packfile write to finish before calling SetReference. Reproduced this consistently across 3/3 runs once I switched the test server from go run to a prebuilt binary, which removed some startup delay that was probably masking it before.
- SSH clone hangs and the remote hangs up unexpectedly. Traced this to ssh.go's stateless-connect handler. wroteWants never gets reset to false in a way that stops a second bufio.Scanner from being created over the same partially read helperStdOut, which looks like it duplicates bytes off the wire.
- Standalone RSL ref push fails the first time if the ref doesn't already exist on the remote. Already discussed in #1231, ReconcileLocalRSLWithRemote tries to fetch a remote ref that doesn't exist yet.

Happy to open separate issues for the first two if that's preferred, or keep tracking them here. Didn't try fixing any of these myself since I wanted a second opinion first, especially on the SSH one.

### DIFF
--- a/.github/workflows/transport-integration-tests.yml
+++ b/.github/workflows/transport-integration-tests.yml
@@ -1,0 +1,31 @@
+name: Transport integration tests
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'internal/git-remote-gittuf/**'
+      - 'test/integration/transport/**'
+permissions: read-all
+jobs:
+  transport-integration:
+    name: Transport integration tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
+        with:
+          persist-credentials: false
+      - name: Install Go
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e
+        with:
+          go-version: '1.26'
+      - name: Install openssh-server
+        run: sudo apt-get update && sudo apt-get install -y openssh-server
+      - name: Build gittuf
+        run: make just-install
+      - name: Run transport integration tests
+        run: bash test/integration/transport/run.sh
--- a/test/integration/transport/httpbackend/main.go
+++ b/test/integration/transport/httpbackend/main.go
@@ -1,0 +1,42 @@
+// Command httpbackend serves one or more bare git repositories over HTTP
+// using git-http-backend, for use in transport integration tests.
+//
+// Usage: httpbackend <project-root> <listen-addr>
+package main
+
+import (
+	"log"
+	"net/http"
+	"net/http/cgi"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+func main() {
+	if len(os.Args) != 3 {
+		log.Fatalf("usage: %s <project-root> <listen-addr>", os.Args[0])
+	}
+
+	root := os.Args[1]
+	addr := os.Args[2]
+
+	out, err := exec.Command("git", "--exec-path").Output()
+	if err != nil {
+		log.Fatalf("unable to resolve git exec-path: %v", err)
+	}
+	backend := strings.TrimSpace(string(out)) + "/git-http-backend"
+
+	handler := &cgi.Handler{
+		Path: backend,
+		Root: "/",
+		Dir:  root,
+		Env: []string{
+			"GIT_PROJECT_ROOT=" + root,
+			"GIT_HTTP_EXPORT_ALL=1",
+		},
+	}
+
+	log.Printf("serving %s on %s", root, addr)
+	log.Fatal(http.ListenAndServe(addr, handler))
+}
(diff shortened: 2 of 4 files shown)


.github/workflows/transport-integration-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
+        with:
+          persist-credentials: false
+      - name: Install Go
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e
+        with:
+          go-version: '1.26'
```
patzielinski, line 25:
```suggestion
          go-version: '1.27'
```

test/integration/transport/httpbackend/main.go
```diff
@@ -0,0 +1,42 @@
+// Command httpbackend serves one or more bare git repositories over HTTP
```
patzielinski, line 1: Needs SPDX header.

test/integration/transport/run.sh
```diff
@@ -0,0 +1,155 @@
...
+set -uo pipefail
+
+REPO_ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
+WORKDIR="$(mktemp -d)"
+FAILURES=0
+
+# Isolate git config from the developer's environment so a global
+# user.signingkey / commit.gpgsign can't interfere with the test repos.
+export GIT_CONFIG_GLOBAL="$WORKDIR/gitconfig-empty"
+export GIT_CONFIG_SYSTEM=/dev/null
```
patzielinski, line 12: Why the difference in where the config env vars are pointed?

test/integration/transport/httpbackend/main.go
```diff
@@ -0,0 +1,42 @@
...
+// using git-http-backend, for use in transport integration tests.
+//
+// Usage: httpbackend <project-root> <listen-addr>
+package main
+
+import (
+	"log"
+	"net/http"
+	"net/http/cgi"
```
patzielinski, line 10: The linter warning here looks to be a false positive?

patzielinski (commented):
Thanks for opening this, @yashrajshuklaaa. Please also take a look at the things the linter has flagged; the log injection bits we can ignore but please look at the other two.

Thanks!

.github/workflows/transport-integration-tests.yml
yashrajshuklaaa: Fixed.

test/integration/transport/httpbackend/main.go
yashrajshuklaaa: added : )

test/integration/transport/run.sh
yashrajshuklaaa: No real reason, nothing writes to it. Simplified both to /dev/null and dropped the now-dead touch.

test/integration/transport/httpbackend/main.go
yashrajshuklaaa: Confirmed real, not a false positive. It's the httpoxy CVE (2016-5386), only affects Go <1.6.3. Suppressed with #nosec since we pin Go 1.27 and it's local-only test infra.

PR #1589, opened 2026-09-12
rsl: preserve annotation targets when reconciling diverged local RSL

### LINKED ISSUE #1587: RSL skip annotations are silently detached from their target during ReconcileLocalRSLWithRemote
### What happened?

I was reading through `ReconcileLocalRSLWithRemote` (`experimental/gittuf/rsl.go`) to understand how it handles a diverged RSL, and noticed that skip annotations don't survive the reconciliation in a fairly common case.

Say you're working locally and haven't pushed yet: you record an RSL entry for some ref, then realize it's bad and skip it with an annotation (`gittuf rsl annotate --skip ...`), all local-only. Meanwhile someone else pushes RSL entries to the remote for a *different* ref. Now your local RSL and the remote RSL have diverged (neither is an ancestor of the other). Next time you sync, `ReconcileLocalRSLWithRemote` kicks in: it resets your local RSL ref to match the remote's tip, then replays your local-only entries on top of it one by one (`experimental/gittuf/rsl.go`, roughly lines 403-412):

```go
switch entry := localOnlyEntries[i].(type) {
case *rsl.ReferenceEntry:
    if err := rsl.NewReferenceEntry(entry.RefName, entry.TargetID, rsl.WithCustomFields(entry.CustomFields)).Commit(r.r, sign); err != nil {
        return fmt.Errorf("unable to reapply reference entry '%s': %w", entry.ID.String(), err)
    }
case *rsl.AnnotationEntry:
    if err := rsl.NewAnnotationEntry(entry.RSLEntryIDs, entry.Skip, entry.Message, rsl.WithCustomFields(entry.CustomFields)).Commit(r.r, sign); err != nil {
        return fmt.Errorf("unable to reapply annotation entry '%s': %w", entry.ID.String(), err)
    }
}
```

The replayed `ReferenceEntry` gets a new parent and a new RSL number, so it naturally gets a brand new commit hash — that part's expected and already covered by existing tests. The problem is the `AnnotationEntry` case: it's recreated with the same `entry.RSLEntryIDs` it originally had, which points at the *old* hash of the reference entry it used to annotate, not the new hash that entry just got reassigned. That old hash no longer exists anywhere in the RSL. So the annotation just stops matching anything — silently. No error, no warning. Anything checking `SkippedBy()` on the reapplied entry now reports it as not skipped.

I didn't want to just assume this from reading the code, so I wrote a small test based on the existing "remote and local have diverged" case in `experimental/gittuf/rsl_test.go`, added a skip annotation on the local-only entry, ran the reconciliation, and checked `SkippedBy()` before and after. Before: true. After: false (log output below). I didn't commit that test anywhere, just used it locally to confirm the bug against current `main`.

Why this matters: skip annotations are the only mechanism gittuf gives you to invalidate a bad or compromised RSL entry after the fact. If a reconciliation can silently undo that with no error, an entry someone explicitly marked invalid can come back as "valid" just because it went through a diverge-and-replay cycle. There's also no test coverage for annotations in the reconciliation path today — `TestReconcileLocalRSLWithRemote` only exercises plain reference entries.

The fix seems reasonably contained: while replaying `localOnlyEntries`, keep a map from each entry's old ID to its newly assigned ID as you go, and when you hit an `AnnotationEntry`, rewrite its `RSLEntryIDs` through that map before recommitting, instead of reusing the stale IDs as-is. Would want a regression test alongside it — both skip and non-skip annotations, and cases where the annotation immediately follows its target vs. is separated from it by other local-only entries.

Affected code: `experimental/gittuf/rsl.go` (`ReconcileLocalRSLWithRemote`), and anything downstream that assumes skip state survives reconciliation, e.g. verification in `internal/policy/verify.go`.

Version/commit: main @ f4538da3dc59e8dad3713ee88cf4796f4a56199b

### Relevant log output

```Shell
original entry ID: 80f370a0cb9ad42172eed83cdec162484e211be5, reapplied entry ID: 1dc4054829f5120dcd63ece547e6e199061ef974
is reapplied entry still marked skipped: false
```

bhuvan-somisetty:
## Description

Fixes #1587.

### Problem

`ReconcileLocalRSLWithRemote` handles the case where a local RSL has diverged from its remote (both sides touched different, non-overlapping refs). It resets the local RSL to the remote's tip and then replays local-only entries on top of it. Replaying a `ReferenceEntry` naturally gives it a new commit ID (its parent and RSL number changed), and that's already covered by existing tests.

The bug is in how `AnnotationEntry` was replayed: it was recommitted with its original `RSLEntryIDs`, which still pointed at the *old* ID of whatever it annotated. That old ID no longer exists in the reconciled RSL, so the annotation silently stopped matching anything. Concretely, a skip annotation on a local-only entry stopped applying to it once the RSL was reconciled with a diverged remote, and the entry would come back as unskipped, with no error or warning.

### Root cause

In `experimental/gittuf/rsl.go`, the replay loop in `ReconcileLocalRSLWithRemote` recreated each local-only entry from scratch (so inferred fields like the RSL number are recomputed correctly), but for annotations it reused `entry.RSLEntryIDs` verbatim instead of pointing at whatever new ID the annotated entry received during replay.

### Fix

While replaying local-only entries, track a map from each entry's original ID to the new ID it's assigned once committed. When replaying an `AnnotationEntry`, rewrite its `RSLEntryIDs` through that map before recommitting, falling back to the original ID for anything that wasn't itself replayed (i.e. entries from before the common ancestor, which didn't move). This is a small, localized change to the replay loop; nothing about the reconciliation algorithm itself changes.

### Testing

Added two regression tests to `TestReconcileLocalRSLWithRemote` in `experimental/gittuf/rsl_test.go`:
- a skip annotation directly on a local-only entry that gets replayed, confirming `SkippedBy` still reports the reapplied entry as skipped afterwards
- a non-skip annotation separated from its target by another local-only entry, confirming the reapplied annotation still resolves to the reapplied entry via `GetReferenceUpdaterEntriesInRangeForRef`

I confirmed both new tests fail against the code before this fix (annotation ends up detached) and pass with it.

Ran `go build ./...`, `go vet ./experimental/...`, `golangci-lint run ./experimental/gittuf/...` (clean), the full `pkg/rsl` suite, and `experimental/gittuf`'s `TestReconcileLocalRSLWithRemote` suite, all passing. The rest of `internal/policy` has pre-existing `ssh-keygen -Y sign` failures in this environment unrelated to this change (reproducible on `main` without this patch too).

### Impact

This only affects the RSL-divergence-reconciliation path (`ReconcileLocalRSLWithRemote`, used from `Sync`/`RecordRSLEntryForReference` etc. when local and remote RSLs have diverged on different refs). No change to behavior when there's no divergence, and no change to how `ReferenceEntry` replay works.

## AI Usage

  request.

I used AI to investigate the codebase. I manually reviewed and verified the change myself, including confirming the new tests fail before the fix and pass after it. The commit message and this description are my own summary of that work.

## Contributor Checklist

  request.
  applicable.

### DIFF
--- a/experimental/gittuf/rsl.go
+++ b/experimental/gittuf/rsl.go
@@ -391,8 +391,18 @@ func (r *Repository) ReconcileLocalRSLWithRemote(ctx context.Context, remoteName
 
 	// Apply local only entries on top of the new local RSL
 	// localOnlyEntries is in reverse order
+	//
+	// Reapplying an entry changes its parent and/or its RSL number, so it is
+	// committed with a new ID. Annotation entries refer to the entries they
+	// annotate by ID, so as we replay entries here, we track how each
+	// original entry's ID maps to the new ID it is assigned, and rewrite an
+	// annotation's referenced IDs using that mapping. Otherwise, a skip
+	// annotation on a reapplied entry would silently stop applying to it, as
+	// it would still refer to the original entry's now unreachable ID.
+	oldToNewEntryID := map[string]githash.Hash{}
 	for i := len(localOnlyEntries) - 1; i >= 0; i-- {
-		slog.Debug(fmt.Sprintf("Reapplying entry '%s'...", localOnlyEntries[i].GetID().String()))
+		originalEntryID := localOnlyEntries[i].GetID()
+		slog.Debug(fmt.Sprintf("Reapplying entry '%s'...", originalEntryID.String()))
 
 		// We create a new object so as to apply anything the
 		// entry may contain that is inferred at commit time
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
@patzielinski PTAL whenever u get chance

PR #1591, opened 2026-09-13
policy: Include controller principals when evaluating its global rules

### LINKED ISSUE #1590: Global rules propagated from a controller repository can never be satisfied by principals declared only in the controller
### What happened?

Global rules propagated from a controller repository into a network repository can never be satisfied by a principal that is declared only in the controller's own metadata. `State.preprocess()` merges a controller's global rules into the network repository's policy state, but never merges that controller's principals into the set of identities used to evaluate those rules. So any signer whose identity exists only in the controller's root or targets metadata, and is not duplicated into the network repository's own metadata, is invisible to global rule verification, even though the whole point of the controller/network model is to declare identities once, centrally, in the controller.

Root cause, in `internal/policy/policy.go`, `State.preprocess()` (lines 1275 to 1289):

```go
for controllerName := range s.ControllerMetadata {
	controllerRootMetadata, err := s.GetControllerRootMetadata(controllerName)
	if err != nil {
		return err
	}

	globalRules := controllerRootMetadata.GetGlobalRules()
	if len(globalRules) > 0 {
		if s.globalRules == nil {
			s.globalRules = map[string][]tuf.GlobalRule{}
		}

		s.globalRules[controllerName] = globalRules
	}
}
```

This loop copies `controllerRootMetadata.GetGlobalRules()` into `s.globalRules[controllerName]`, but it never calls `controllerRootMetadata.GetPrincipals()` and never adds those principals to `s.allPrincipals`. Compare with the equivalent handling for the repository's own root, targets, and delegated metadata a few lines above (`policy.go:1194` to `1196`, `1214` to `1216`, `1247` to `1249`), which all populate `s.allPrincipals` from the corresponding `GetPrincipals()` call. The controller branch is the only metadata source in `preprocess()` that contributes global rules but not principals.

Downstream, `getVerifierForGlobalRules` (`internal/policy/policy.go:451`) builds a single exhaustive verifier from `s.allPrincipals`, the only verifier used to determine which principals count as signers for every global rule:

```go
func (s *State) getVerifierForGlobalRules() *SignatureVerifier {
	...
	verifier := &SignatureVerifier{
		...
		principals: []tuf.Principal{},
		threshold:          1,
		verifyExhaustively: true,
	}

	for _, principal := range s.allPrincipals {
		verifier.principals = append(verifier.principals, principal)
	}
	...
}
```

That single verifier's signer count is then checked against every global rule from every controller, in `internal/policy/verify.go:1242` to `1284`:

```go
verifiedGlobalRulesPrincipalIDs := 0
if globalRulesVerifier != nil {
	globalRulesPrincipalIDs, err := globalRulesVerifier.Verify(ctx, gitID, authorizationAttestation)
	...
	verifiedGlobalRulesPrincipalIDs = globalRulesPrincipalIDs.Len()
}

for controllerName, globalRules := range policy.globalRules {
	...
	for _, rule := range globalRules {
		switch rule := rule.(type) {
		case tuf.GlobalRuleThreshold:
			...
			if verifiedGlobalRulesPrincipalIDs < requiredThreshold {
				return "", false, ErrVerifierConditionsUnmet
			}
		...
```

Since `verifiedGlobalRulesPrincipalIDs` is computed once from `s.allPrincipals`, and `s.allPrincipals` never contains controller only principals, a controller declared threshold rule counts only signatures from principals that also happen to be declared in the network repository's own metadata.

This gap was introduced in commit 08867e27 ("policy: Verify global rules from controller repository"), which added the global rules copy loop above but no corresponding principal merge, and it has not been touched since (`git log -p -- internal/policy/policy.go` shows no later change to this loop).

I confirmed this empirically rather than only reading the code. I wrote a test in `internal/policy` that mirrors the existing `TestVerifyEntry/verify_global_rules_applied_from_controller_repository` subtest, but with one change: the network repository's own metadata is built from `createTestStateWithOnlyRoot`, which registers only the shared SSH root key, instead of `createTestStateWithPolicy`, which happens to also independently declare the same GPG key used by the controller's global rule. That is exactly the gap in the existing test suite: the current controller global rule test never isolates a principal that is genuinely controller only, because its network side coincidentally redeclares the same GPG key locally, which masks the bug.

In the test, the controller declares a `threshold-2-main` global rule on `refs/heads/main` trusting two principals: the shared SSH root key, and a GPG key that only appears in the controller's targets metadata. The network repository adds the controller and propagates its policy, so the controller's global rule is present in the network state. A commit to `refs/heads/main` is then signed by the GPG key, with a reference authorization signed by the SSH root key, so both principals the controller's rule requires actually signed. Verification of that entry still fails, because the GPG key was never merged into the network state's principal set.

### Impact

This breaks the core value proposition of the controller/network repository feature. Identities are meant to be declared once in a controller and trusted by every network repository, but any identity declared only in the controller, and not duplicated into each network repository's own root or targets metadata, is silently unusable for meeting that controller's own global rule thresholds. Depending on how much a rule's trusted principals overlap with the network repository's own local principal set, this can make a controller declared threshold rule partially or completely unsatisfiable, always failing verification with a generic `ErrVerifierConditionsUnmet`, with no indication that the real cause is a valid signer not being recognized because it is controller only. Global rules and controller/network propagation are both actively developed features (recent history includes 08867e27, 77dd4dab, 58bde0d5, 96516300, 0b332fcc, 8dae73bb), so this is a functional correctness bug in a feature real adopters would use, not a cosmetic issue.

### Affected components

- `internal/policy/policy.go`: `State.preprocess()`, `State.getVerifierForGlobalRules()`
- `internal/policy/verify.go`: the global rule verification loop in `verifyGitObjectAndAttestations`
- The controller/network repository trust propagation feature end to end

### Suggested fix direction

- When `preprocess()` loads each controller's root metadata, also collect `controllerRootMetadata.GetPrincipals()` into a principal set scoped to that controller, rather than folding them into the network repository's own `s.allPrincipals` directly, since that could unintentionally let a controller's principals count toward rules declared locally by the network repository too, which is a separate trust question that should be handled deliberately rather than as a side effect.
- Make `getVerifierForGlobalRules` (or its caller) controller aware: build one exhaustive verifier per key in `s.globalRules` (the local case using the existing `s.allPrincipals`, and each named controller case using that controller's own principal set), and check each rule's threshold against the count from the matching verifier instead of a single shared count.

### Acceptance criteria

- A propagated controller global rule can be satisfied by a signer whose principal is declared only in the controller's metadata, not duplicated in the network repository's own metadata.
- A regression test that signs a real Git object with a controller only principal and asserts it is correctly counted toward that controller's propagated global rule threshold.
- A regression test confirming principals declared only in one controller are not usable to satisfy a different controller's, or the local repository's, global rules, to guard against the fix over broadening trust.
- No change in behavior for the existing single repository, no controller, global rules path.

Version/commit: main @ 7fa8fcd1 (`git log -p -- internal/policy/policy.go` confirms the code shown above is unchanged at this revision)

### Relevant log output

```Shell
=== RUN   TestReproControllerOnlyPrincipalNeverCountsTowardGlobalRule
    zzrepro_controller_global_rule_test.go:85: gpgKey ID: 157507bbe151e378ce8126c1dcfe043cdd2db96e
    zzrepro_controller_global_rule_test.go:90: gpgKey recognized in controller's own allPrincipals: true
    zzrepro_controller_global_rule_test.go:98: network repo recognizes principal: SHA256:ESJezAOo+BsiEpddzRXS6+wtF16FID4NCd+3gj96rFo
    zzrepro_controller_global_rule_test.go:101: gpgKey recognized in network repo's allPrincipals: false
    zzrepro_controller_global_rule_test.go:144: verifyEntry result (expected nil under correct behavior): verifying Git namespace policies failed, gittuf policy verification failed
    zzrepro_controller_global_rule_test.go:145:
        Error Trace: internal/policy/zzrepro_controller_global_rule_test.go:145
        Error:       Expected nil, but got: &fmt.wrapError{msg:"verifying Git namespace policies failed, gittuf policy verification failed", err:(*errors.errorString)}
        Test:        TestReproControllerOnlyPrincipalNeverCountsTowardGlobalRule
        Messages:    controller-declared threshold rule should be satisfiable by a principal declared only in the controller, but verification failed: verifying Git namespace policies failed, gittuf policy verification failed
--- FAIL: TestReproControllerOnlyPrincipalNeverCountsTowardGlobalRule (23.86s)
FAIL
```

This scratch test isolates exactly the gap in the existing suite: same setup as `TestVerifyEntry/verify_global_rules_applied_from_controller_repository`, but with the network side built from `createTestStateWithOnlyRoot` instead of `createTestStateWithPolicy`, so the GPG key is genuinely controller only rather than coincidentally redeclared locally. The test file was used locally to confirm the bug and was not committed.

bhuvan-somisetty:
## Description

Fixes #1590.

Global rules propagated from a controller repository into a network repository could never be satisfied by a principal declared only in the controller's own metadata. Identities are meant to be declared once, centrally, in a controller and trusted by every network repository that adds it, but any identity not also duplicated in the network repository's own root/targets metadata was invisible to global rule verification.

### Root cause

`State.preprocess()` (`internal/policy/policy.go`) copied a controller's global rules into the network repository's policy state, but never merged that controller's declared principals into any principal set used to evaluate those rules. `getVerifierForGlobalRules()` built a single exhaustive verifier from the current repository's own `allPrincipals`, and that one verifier's signer count was checked against every global rule from every controller in `verifyGitObjectAndAttestations` (`internal/policy/verify.go`). Since `allPrincipals` never contained controller-only principals, a controller-declared threshold rule could only be satisfied by signers who also happened to be declared in the network repository's own metadata, silently and permanently, always failing with a generic `ErrVerifierConditionsUnmet`.

This was introduced when controller global rule propagation was added and had not been touched since; no existing test isolated a genuinely controller-only principal (the current controller test's network side happens to independently redeclare the same GPG key locally, masking the gap).

### Solution

- `preprocess()` now also collects each controller's own principals (from its root, top-level targets, and delegated targets metadata) into a `controllerPrincipals` map keyed by controller name, mirroring how this repository's own principals are already collected.
- `getVerifierForGlobalRules()` takes a `controllerName` and builds (and caches) one exhaustive verifier per global rules source: the local case (`""`) still uses the repository's own `allPrincipals`; a named controller uses only that controller's own principals, never the local repository's or another controller's.
- `verifyGitObjectAndAttestations` looks up the matching verifier for each source in `policy.globalRules` instead of sharing a single count across every source, so each source's rules are checked only against principals it actually declares.

This keeps the local (non-controller) global rules path unchanged and does not widen trust: a controller's principals are still not usable for the network repository's own rules, and one controller's principals are not usable for another controller's rules.

### Testing

Added three regression tests in `internal/policy/verify_test.go` (`TestVerifyEntry`):
- `verify global rule from controller repository is satisfied by a principal declared only in the controller`: a controller declares a threshold-2 rule trusting an SSH key and a GPG key that only exists in the controller's own metadata; after propagation, a commit signed by both is now correctly accepted.
- `controller-only principal does not satisfy the network repository's own local global rule`: the same controller-only GPG key must not count toward a separate, local threshold rule the network repository declares on its own, guarding against an overly broad fix.
- `verify global rule from a controller repository that declares no targets metadata`: a controller that declares its global rule and principal entirely in root metadata, with no targets file at all, still has its rule correctly satisfied, exercising the branch in `preprocess()` where controller principals are collected only from root metadata.

Also added a small test helper, `createTestStateWithOnlyRootAndEmptyTargets`, needed because `preprocess()` returns before processing controller metadata when the repository's own `TargetsEnvelope` is nil.

`go build ./...` and `go vet ./internal/policy/...` pass. `go test ./internal/policy/... -run TestVerifyEntry` passes in full, including all pre-existing global-rule/controller subtests and both new ones.

### Impact

Correctness fix for the controller/network-repository trust-propagation feature. No change to the API surface or to single-repository (no controller) global rules behavior.

## AI Usage

  request.

I used AI to investigate the codebase. I reviewed and verified every part of the change myself: I traced the root cause through `preprocess()`, `getVerifierForGlobalRules()`, and the verification loop against the actual source, wrote and ran an isolated scratch reproduction against `main` to confirm the bug empirically before fixing it (not committed), and re-ran the targeted tests after the fix to confirm both the fix and that trust is not over-broadened. The commit message and this description are my own summary of that work.

## Contributor Checklist

  request.
  applicable.


patzielinski (commented):
Review pending discussion on how to proceed in https://github.com/gittuf/gittuf/issues/1590.

bhuvan-somisetty:
Made it opt-in per the issue discussion, off by default now. Should be good to take another look whenever.

patzielinski:
Heads up I meant that I'm seeking input from other maintainers on that issue on the implications of also having identity propagated. E.g. how do we resolve conflicts, etc.

bhuvan-somisetty:
Ah gotcha, makes sense. I'll leave this as is for now and wait to hear how that discussion lands before pushing anything further.

patzielinski:
I'm closing this PR given the discussion in #1590 as we'd like to think about the design of this a bit before we implement something into the codebase.

PR #1604, opened 2026-09-17
trust: honor --with-rsl-entry in update-propagation-directive

### LINKED ISSUE #1603: trust update-propagation-directive ignores --with-rsl-entry (TUI and CLI)
### What happened?

The `--with-rsl-entry` persistent flag on `gittuf trust` commands is supposed to record an RSL entry for the change in addition to updating the policy-staging ref. `add-propagation-directive` and `remove-propagation-directive` both honor this flag correctly, but `update-propagation-directive` silently drops it, in both the CLI command and the TUI.

`experimental/gittuf/root.go`'s `UpdatePropagationDirective` accepts `opts ...trustpolicyopts.Option` and uses `options.CreateRSLEntry` exactly like `AddPropagationDirective`/`RemovePropagationDirective` do when calling `updateRootMetadata`. However, the two callers never construct or pass any `trustpolicyopts.Option`:

- `internal/cmd/trust/updatepropagationdirective/updatepropagationdirective.go` (`Run`) calls:
  ```go
  return repo.UpdatePropagationDirective(cmd.Context(), signer, o.name, o.upstreamRepository, o.upstreamReference, o.upstreamPath, o.downstreamReference, o.downstreamPath, true)
  ```
  with no `opts` argument at all, even though `o.p.WithRSLEntry` (from the shared `persistent.Options`, exposed via the `--with-rsl-entry` flag) is available. Compare with `internal/cmd/trust/addpropagationdirective/addpropagationdirective.go` and `internal/cmd/trust/removepropagationdirective/removepropagationdirective.go`, which both build:
  ```go
  opts := []trustpolicyopts.Option{}
  if o.p.WithRSLEntry {
      opts = append(opts, trustpolicyopts.WithRSLEntry())
  }
  ```
  and pass `opts...` to the repo call.

- `internal/cmd/tui/trusthelpers.go`'s `repoUpdatePropagationDirective` has the same gap:
  ```go
  return repo.UpdatePropagationDirective(ctx, signer, name, upstreamRepository, upstreamReference, upstreamPath, downstreamReference, downstreamPath, true)
  ```
  while every other trust helper in that file (`repoAddPropagationDirective`, `repoRemovePropagationDirective`, `repoAddGlobalRule`, etc.) builds `opts` from `o.p.WithRSLEntry` and passes it through.

As a result, running `gittuf trust update-propagation-directive --with-rsl-entry ...` (or using the "Update" action on the propagation directive screen in the TUI with "create RSL entry" enabled) updates the policy-staging ref but never creates the RSL entry, unlike the add/remove equivalents. This is inconsistent behavior for an option users would reasonably expect to work the same way across all three propagation-directive subcommands.

I expect `update-propagation-directive` to create an RSL entry when `--with-rsl-entry` is passed, matching `add-propagation-directive` and `remove-propagation-directive`, and the TUI's update action to behave the same as its add/remove counterparts.

Version: reproduced on `main` at commit `067f3ea8`.

### Relevant log output

_No response_

bhuvan-somisetty:
## Description

`update-propagation-directive` (both the CLI command and the TUI) silently ignored the `--with-rsl-entry` persistent flag, unlike `add-propagation-directive` and `remove-propagation-directive`. `UpdatePropagationDirective` already accepts `trustpolicyopts.Option` and honors `WithRSLEntry()`, but neither caller ever built or passed the option.

This PR fixes that by building `opts` from `o.p.WithRSLEntry` and passing it through, in:
- `internal/cmd/trust/updatepropagationdirective/updatepropagationdirective.go`
- `internal/cmd/tui/trusthelpers.go` (`repoUpdatePropagationDirective`)

This matches the existing pattern used by add/remove propagation directive commands.

Fixes #1603

## AI Usage

  request.

I used AI to investigate the issue, implement the fix,. I reviewed all changes manually.

## Contributor Checklist

  request.
  applicable.

## Test plan

- [x] Verified the fix mirrors the existing, working `add-propagation-directive`/`remove-propagation-directive` pattern
- [x] Strengthened `TestUpdatePropagationDirective/success_with_RSL_entry` to assert a new RSL entry is created (fails without the fix, passes with it)
- [x] `go build ./...`, `go vet`, and `golangci-lint run` pass on changed packages


internal/cmd/trust/updatepropagationdirective/updatepropagationdirective_test.go
patzielinski, line 1: We can omit the tests for this.

patzielinski (commented):
One nit and this should be good to go, thanks!

internal/cmd/trust/updatepropagationdirective/updatepropagationdirective_test.go
bhuvan-somisetty: Good call, removed those assertions. Pushed.

patzielinski (commented):
Looks good. Please squash commits and we'll get this merged.

bhuvan-somisetty:
Squashed into a single commit, thanks for the review @patzielinski!

patzielinski (approved):

